PR #31, opened 2014-07-23
VI module: TCPIP compatibility / minor PNA functions added

arndtvb:
as promised...

To be checked in other lab environments with different OS/NI-VISA/Pyvisa versions (well, I guess we can assume that the last two are up to date).


arsenovic:
i cherry-picked the windowing functions from this in f011f5663789f79fb9c9ca51bf8469aeb8e46cb4 . the tcpip is now supported, through the use of the python-ivi module. 

backwards compatibility is maintained by testing if the address is an int or not. if its an int (old behavior) the correct gpib resource string is created.

PR #53, opened 2015-04-06
Changed the except clause in a Python3 compatible way.

Hash--:


arsenovic:
seems to have failed. perhaps it needs a newer version of setuptools in the requirements

Hash--:
Probably. This time it's print function without parenthesis, and there is
few to correct. What is the function of this ez_setup.py  ?



2015-04-06 21:54 GMT+02:00 alex arsenovic <notifications@github.com>:

arsenovic:
i removed the ez_setup.py ( i forgot what it did anyway). now there are syntax errors for python3 printing. i think you should be able to pull hte most recent master and apply your  changes.

https://github.com/scikit-rf/scikit-rf/commit/952db4976fb07d9afbaaa5526ba912a6898027b8

Hash--:
OK. I'll try. I've made a Python3 working version, but it was few month
ago, so I must dive into again and check against the recent master for
changes...

2015-04-06 22:10 GMT+02:00 alex arsenovic <notifications@github.com>:

PR #55, opened 2015-05-19
Added Group Delay

core-dd:
Hi,
I added group delay calculations and thought you might find it helpful.
I hope the if-statement around lines 630 in network.py is not to messy.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -268,6 +268,7 @@ class Network(object):
             npy.angle(x))),
         'arcl_unwrap'   : lambda x: mf.unwrap_rad(npy.angle(x)) *\
             npy.abs(x),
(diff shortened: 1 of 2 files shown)


arsenovic:
nice!

arsenovic:
i noticed this does not work for 2-port networks.  as a workaround, it is possible to take the subnetwork, ie `ntwk.s21.s_gd`. however the correct equation requires frequency information, for instance if the frequency vector is not evenly spaced. 

anyway, i think i will re-implement this as a property. also , i cant think of a practical use of  group delay on the other parameters. i am thinking it will be `ntwk.group_delay` .

do you have thoughts or objections?

core-dd:
Hej,
I'm not aware of any other specific first derivative w.r.t. frequency that carries information besides the phase. Hence, a specific property might be the best solution. Further, I've never considered non-evenly spaced frequency ranges. Good that you think in a broader sense. :)

PR #59, opened 2015-05-25
y label simplification and some documentation

core-dd:
Hi Alex,
here you'll find the discussed addition to the `y_label`-issue.


arsenovic:
looks good!

PR #60, opened 2015-05-29
BUGFIX: do not overwrite given network name

core-dd:


arsenovic:
@core-dd , email me  you address and size and ill send you a skrf shirt!

PR #62, opened 2015-07-31
Adds arbitrary reference immittance labeling to smith(...) function

gibbons:
These changes allow the labels (if plotted) on the radius==1 Smith charts (Z and Y) to be scaled to arbitrary reference immittances, e.g. 50 Ohms instead of 1 Ohm.

In order to accommodate that this now allows for varying-length labels, the labels' placements are given using cylindrical coordinates (implemented in complex math) and their alignments (left, center, right, up, or down, all relative to the placement coordinate) adjusted to ensure the labels don't overlap the chart.

Lastly, to allow for zooming without messing up the plotted Smith chart, two points in white (hence invisible) are plotted outside the chart to ensure proper data limits.


arsenovic:
sweet

PR #63, opened 2015-08-25
add format support to write_touchstone

aerialist:


arsenovic:
this is great!  long been on the todo list. the travis is failing just for the python 3.3 which is due to some other problems i havent had time to dig into yet. 

thanks!

aerialist:
Hi Alex,

This PR doesn't seem to be merged. Could you?

arsenovic:
sorrty this is taking so long, i just need to go back and figure out where the change was made, because i think i made the same fix at some point.

aerialist:
I saw you implemented it in networkset_2_spreadsheet function in io.general.py
And I tried to follow its argument names.

PR #64, opened 2015-08-26
accept filename in unicode

aerialist:
Hi Alex,

Sorry I started to mess around the GitHub...

I just wanted to make a quick PR for this.
https://github.com/scikit-rf/scikit-rf/commit/1d6150598b3bde16c5bd1a9820907343ccb87af8

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1688,7 +1688,8 @@ class Network(object):
         # touchstone file.
 
     def write_touchstone(self, filename=None, dir = None,
-                         write_z0=False,skrf_comment=True):
+                         write_z0=False,skrf_comment=True,
+                         form='ri'):
         '''
         Write a contents of the :class:`Network` to a touchstone file.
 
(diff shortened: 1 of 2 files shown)


arsenovic:
@aerialist  
  after reading around this python3 seems to me the largest failure and waste of time imaginable. 

regardless, i committed the `basestring` fix you suggested to master. i also started another branch called `p3k`, that has been `futurized` using the python-future script, and it passes the test-suit for python 3.3.  additionally, i copied the module `p3k.py` from `numpy.compat` into skrf, which should be useful in the conversion. 

if you are going to add things in the mean while you probably wan to use the p3k branch. do you use `conda`?

currently,  i am still migrating my other modules to python3, and i will try and get the python2 tests passing on the new p3k branch. then merge into master.

aerialist:
Excellent. I appreciate your effort, Alex!

I've been using Python(x,y) for long time but planning to use conda to try out Python3. 
What about PR#63 on format support for write_touchstone? Shall I make new PR to p3k branch?
https://github.com/scikit-rf/scikit-rf/pull/63

arsenovic:
the original code actually works in `p3k`, because in python3  `str` now represents unicode, so you shouldnt have to change anything. 

i am also considering making the jump to [Network2](http://nbviewer.ipython.org/github/scikit-rf/examples/blob/master/Network2.ipynb) or similar....  which would essentially break all backward compatability.

arsenovic:
yes, the fix in `skrf/util.py` was already made, but i  had forgotten about your other improvements. 
thanks a lot for the changes and persistence!

PR #65, opened 2015-09-28
Generalized mixed mode

core-dd:
Tried to prepare the gmm calculations today as the changes rot on the disk here for some time now. I'm using the transformation from single ended to generalized mixed mode parameters regularly but have never tried it the other way round.

I'm considering the code as highly experimental. Unsure whether you want to add them in the current stage.

Cheers,
Jens

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2629,28 +2629,26 @@ class Network(object):
     #TODO: automated test cases
     def se2gmm(self, p, z0_mm=None):
         #TODO: assumes 'proper' order (differential ports, single ended ports)
-        if z0_mm is not None:
-            self.z0_mm = z0_mm
-        elif not hasattr(self, 'z0_mm'):
+        if z0_mm is None:
             #TODO: choose z0mm automagically
             raise NotImplementedError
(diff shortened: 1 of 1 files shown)


core-dd:
the automatic calculation of z0_mm is untested at the moment :/

arsenovic:
i dont have any way to test this, but it looks sweet! 

 you could also add some info on  this in the docs and/or examples. otherwise i dont think people will notice it.

arsenovic:
when i have things that are untested, i try to raise a warning,  or at least mention it in the docs.

core-dd:
a warning is mentioned in the docstrings for the se2gmm and gmm2se method, I try to find some more time to spend on the documentation this week

arsenovic:
excellent, i think a few people have asked for this feature. you could email the scikit-rf list. getting more people to try it is good for testing!

arsenovic:
@core-dd ,  
  what do you think about making these return a new network instead of just a s-matrix?


would that make sense?

core-dd:
@arsenovic ,
that makes sense. Physically the network stays the same but structural seen excications etc change. Hence, nothing against that move.

PR #68, opened 2016-01-10
Smith Chart Options

kungpfui:
Some graphical options I like to see on smith chart plots.

chart_type="zy" or "yz"
----------------------- 
Like "z" or "y" but the circles of the second character is drawn as well but nearby transparent. In that way the chart looks not overstuffed.
chart_type_zy

draw_vswr
---------
Sometimes the VSWR circles would be nice to be seen. So an additional keyword-parameter *draw_vswr* can be set. Can be True = (my) default VSWR list, your own list of VSWR values or None. If *draw_labels* is set these circles are labeled as well.
draw_vswr

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -141,7 +141,7 @@ from six.moves import xrange
 import os
 import warnings
 try:
-    import cPickle as pickle
+    import cPickle as pickle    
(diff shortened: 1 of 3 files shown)


arsenovic:
cool!

PR #69, opened 2016-02-03
Add a2s, a2z and Network a.setter functions

Ttl:
ABCD and Z parameters are very closely related and conversion
between them is identical (Table V in [1]). This commit defines
some missing ABCD conversion functions using this relation.

[1] Frickey, D.A., "Conversions between S, Z, Y, H, ABCD, and T parameters which
are valid for complex source and load impedances," in Microwave Theory and
Techniques, IEEE Transactions on , vol.42, no.2, pp.205-211, Feb 1994

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1196,9 +1196,10 @@ class Network(object):
         .. [#] http://en.wikipedia.org/wiki/impedance_parameters
         '''
         return s2a(self.s, self.z0)
+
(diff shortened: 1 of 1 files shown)


arsenovic:
great. it would be sweet to add tests for this if you have any.

btw the travis tests are failing due to some problem with ipython, not scikit-rf

PR #70, opened 2016-02-12
Add test for conversion between parameters

Ttl:
Here's one test case that checks that converting from S-parameters to ABCD/Z/Y/T and back to S-parameters returns the original network. It uses the test networks that are defined in test_network file and tests are done with different z0s. I meant to still add one test case that checks that a simple network is converted correctly, but I don't think I have time for it right now.

There's also another small commit that adds a missing default 50 ohm z0 to s2a function.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -4058,7 +4058,7 @@ def z2a(z):
         ]).transpose()
     return abcd
 
-def s2a(s,z0):
+def s2a(s,z0=50):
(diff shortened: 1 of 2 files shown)


arsenovic:
excellent. 
  
also i looked at your blog. its very well done and the work is impressive.  it would be sweet to have something hands on like this  in the scikit-rf examples.

PR #72, opened 2016-03-05
Added the "future" package in the requirement list. 

jhillairet:
The "future" package is not installed by default in Anaconda 4.0.


arsenovic:
hi, 
   i added this in manually.  although  i think anaconda may have included this  now, because i just tested the conda builds with a fresh install for 2.7 and 3.5 and they worked. our coverage is low though...

arsenovic:
also  the codebase should work with py3.5 now, and i made a conda build

PR #74, opened 2016-07-29
Add LMR16 calibration

Ttl:
This commit adds LMR16 calibration procedure. It is a self calibration procedure for leaky VNA that corrects for crosstalk with 16-term error model. It can solve either through or reflect standard when the other one is given.

I have tried this with real measurements and it is working as it should. Tests also pass.

There are some things that are not done yet:
* I don't know proper names of the isolation terms in 16-term error model and couldn't find them anywhere.
* Coefficient list has 16 terms when 15 is enough. Now some of them are linearly dependent, but I couldn't manage to remove one without making T-matrix transformation too complex. Accuracy isn't affected so I didn't find it important enough to fix.

Implementation is based on this paper:

K. Silvonen, "LMR 16-a self-calibration procedure for a leaky network analyzer,"
in IEEE Transactions on Microwave Theory and Techniques, vol. 45, no. 7, pp.
1041-1049, Jul 1997

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -161,7 +161,7 @@ class Calibration(object):
     '''
     family = ''
     def __init__(self, measured, ideals, sloppy_input=False,
-        is_reciprocal=True,name=None,*args, **kwargs):
+        is_reciprocal=True,name=None, self_calibration=False,*args, **kwargs):
         '''
         Calibration initializer.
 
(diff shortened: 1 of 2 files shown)


Ttl:
I also added a regular non self-calibrating version of 16-term calibration.

arsenovic:
this looks great, thanks for adding! it looks like you need to remove a non-ascii character in your last commit  for the travis tests to pass.

Ttl:
By the way, during writing these functions I noticed that the "ntwk.s = npy.array([[s11, s12],[s21,s22]]).transpose().reshape(-1,2,2)" pattern used in many places in the code swaps the s12 and s21 terms of the resulting matrices. Many of the matrices it is used in are symmetrical so the result is same, but there are few places that look like it could be a mistake.

For example the above snippet is from calibration.py:3640 and to me it looks like there might a bug there.

arsenovic:
and thanks for the bug catch.  i fixed that one, but didnt see any other.  if you find others let me know or squash em.

PR #77, opened 2016-09-22
Extract port names from touchstone files according to ansys format

mcclurem:
This tries to extract port names from touchstone file comments according to the format used by HFSS and SIWave. This also enables indexing for single port networks by overloading Network.__getitem__ to accept either 2 or 3 dimensional indexes


arsenovic:
looks really good. nice doc updates too.

PR #82, opened 2016-12-15
Updated documentation for skrf.network.Network.se2gmm()

jrmagers:
Updated documentation for se2gmm()


core-dd:
Thank you for the improvement, back then as I've implemented se2gmm I haven't though to much about how to handle port numbering properly or even semi-automated.

PR #84, opened 2016-12-22
Separating out the plotting

dvincentwest:
The purpose of these changes is to separate out the matplotlib based plotting so that data visualization is separated from the rest allowing the use of skrf in other environments, like gui environments, or using bokeh instead of matplotlib


arsenovic:
looks good. 
i added setup_pylab to the test_network Setup, so that the old plotting tests pass, and fixed a minor bug in media.write_csv, which i had fixed previously. 

we will also have to update the docs to reflect this new feature.

PR #86, opened 2017-01-03
Add NIST Multiline TRL calibration algorithm

Ttl:
Implements the NIST Multiline TRL calibration algorithm. Compared to the current MultilineTRL calibration this one doesn't have bandwidth limitation when using short and long lines. Different lines are also weighted for the best accuracy at each frequency point which results in a more accurate calibration.

Algorithm is the one published in [0], but implementation is mostly straight forward ported Matlab code from [1].

Tested with real measurements and it seems to work. Tests also pass.

    [0] D. C. DeGroot, J. A. Jargon and R. B. Marks, "Multiline TRL revealed," 60th ARFTG Conference Digest, Fall 2002., Washington, DC, USA, 2002, pp. 131-155.

    [1] K. Yau "On the metrology of nanoscale Silicon transistors above 100 GHz" Ph.D. dissertation, Dept. Elec. Eng. and Comp. Eng., University of Toronto, Toronto, Canada, 2011.


arsenovic:
sweet! i will look forward to using this. if you like you can add   an entry into the doc list at the top of the module, so that it shows up in the docs (once we get them to build again!)

arsenovic:
you might consider doing  the s2t,t2s shaping  in the NISTTRL?  otherwise  it would make more sense to do this cleanly for all conversions, for consistency. at the least, you may want to adda comment in network.py so that if we know what depends on this new feature. up to you.

thanks again!

arsenovic:
@dvincentwest   you like to check this out.

PR #90, opened 2017-01-08
Add normalized standard deviation output to NIST M-TRL

Ttl:
Adds normalized standard deviation output as described in the multiline TRL paper. Basically it can be used to estimate the error of the calibration at different frequencies.

Examples:

nstd_1line
X-axis is frequency. Calibration with a single 1 mm long line. High error at 60 GHz due to phase being 180 degrees there.

nstd_3line
Otherwise the same calibration, but now done with 0.1, 0.5 and 1.0 mm long lines.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -2483,6 +2483,7 @@ class NISTMultilineTRL(EightTerm):
         Smat2 = npy.ones(shape=(fpoints, 2, 2), dtype=npy.complex)
 
         e = npy.zeros(shape=(fpoints, 7), dtype=npy.complex)
+        nstd = npy.zeros(shape=(fpoints), dtype=npy.float)
 
         def t2s_single(t):
             return t2s(t[npy.newaxis,:,:])[0]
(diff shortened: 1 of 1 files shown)


dvincentwest:
Thanks a lot for contributing this. Would you mind including an example set of Cal standards measurements, raw data and expected result for the calibration to use as a reference?

Ttl:
I have some real data, but I can't share it. You can find example of usage in the tests at "skrf/calibration/tests/test_calibration.py" file.

The required parameters are:

measured : List of measured standards. Same order as the ordinary TRL (through, reflects, lines).

Grefls : List of reflection standards reflection coefficients. 1 for open and -1 for short. If multiple reflects are used order must be the same as in the measured list.

l : Lengths of the lines in meters. Also includes length of the through. Again the order must be same as in the measured list.

Optional parameters:

er_est : Estimate of the effective permittivity of the line. Both real and imaginary part can be given. Imaginary part should be given at 1 GHz. Real part should be enough for most cases.

n_reflects : Number of reflect standards.

refl_offset : List of lengths of the reflect standards. Used if reflects are offset reflects.

ref_plane : Reference plane movement after the calibration. It uses the solved propagation constant of the line to add a specified length of line to both ports.

p1_len_est, p2_len_est : Estimates of the port lengths. I haven't had a need to change these from 0. Used for solving reflection tracking.

gamma_root_choice : Solving the propagation constant requires choosing a sign of the square root. This option specifies how it should be done. 'real' should be the best option if the lines are lossy. Maybe it even should be the default option.

Besides the error coefficients the method also solves the propagation constant and effective permittivity of the line. It's a good idea to check that the solved propagation constant looks reasonable after the calibration.

arsenovic:
i have not looked at the code for this in detail, but to determine epsilon
for non TEM media (waveguide), i think  you should have to provide a
function so that the algorithm knows  how eps  is related to gamma?

i think if you implement a test_eps, it will fail for waveguide.

Ttl:
It solves the effective permittivity which can be determined straight from the propagation constant by a calculation: "er_eff[m] = -(gamma[m]/(2*pi*freqs[m]/c))**2". If you wanted to solve the permittivity of the filling material you would need some algorithm for it.

arsenovic:
yes i agree. i made a small note about this in the doctstring for er.

PR #94, opened 2017-01-13
Analyzer Drivers Implemented, Data Grabber should be working

dvincentwest:
@Ttl @arsenovic 
environment variable "SKRF_PLOT_ENV" now specifies which plotting environment to initialize when skrf is first loaded.

- if set to 'pylab' skrf.setup_pylab() will be called (the previous default behavior)
- if set to 'pylab-skrf-style', both skrf.setup_pylab() and skrf.stylely() will be called to initialize matplotlib with skrf style plots.

If the environment variable is not set, skrf will use 'pylab' as the default, preserving the behavior before th plotting architecture changes

qtapps now sets SKRF_PLOT_ENV to 'none'

The Basic Data Grabber should be working now with one example of an instrument driver in place, along with a fleshed out base_analyzer.Analyzer class to use when making new drivers, and and Jupyter notebook that provides the important tests to make sure that the driver works with skrf_qtapps API as it 
currently exists


arsenovic:
you may consider using two env vars, one for each thing. but i dont mind either way.

also it may make sense to eventually create a skrf.conf, instead of use env vars, because i think that may be easier for windows users. we could make methods in skrf to change these as well, so they would not need to do it manually. like 

`rf.set_plot_env('pylab')` # change for this session, 
`rf.save_plot_env('pylab')` # write to skrf.conf

you could also implement a context manager for this, if it ever made sense. as long as we keep default behavior working ,  its ok.

Issue #96, opened 2017-01-15
Comment - Instrument driver architecture


helgrind:
Hello,

My two cents RE instrument drivers:

I think it would be wise if you only used individual drivers to define the SCPI commands and override any necessary methods from the parent class. The commands could be kept in a dictionary with a set of required commands for each instrument type (which can be added to and used in overridden methods). This has two main benefits, being that new drivers are simple to implement by users, and also that any update to the driver functions for speed etc are pushed out to the whole set without thinking about ways to do that through inheritance.

An example pseudocode/python structure would be:

```
PNA.py:
...
self.commands = {
  'reset': 'SYST:PRES',
  'setFreqStart': 'SENS:FREQ:STAR {e}',
  'setFreqStop': 'SENS:FREQ:STOP {e}'
}

def setFreq(self):
  super(PNA, self) #optional
  interface.write(vna.command('somethingExtra')

```
And use:
```
na.py:
(conditionally load the relevant driver)
interface.write(vna.command('setFreqStart'), [minFreq]);
interface.write(vna.command('setFreqStop'), [maxFreq]);
...
while(!interface.write(vna.command('isReady'))
  time.sleep(1);
trace = interface.write(vna.command('getTrace')
```

The conditional logic for inserting values and building SCPI strings would be put in the instrument class as they shouldn't vary heavily between models. Note the second argument is a list so this could contain strings and numerics. If the particular model needs some extra commands or needs to skip some, then override the relevant method from the instrument class.

Thoughts?

dvincentwest:
I very much appreciate the thoughts.  I've spent most of the day thinking about this and I think its an important thing to try to get right as early as possible.

I am inclined to give it a try focusing on Agilent 2-port PNAs (where should the 'instrument class' lines be drawn?).  Perhaps I can ask for your further thoughts when I have a working model system.

helgrind:
Sounds good, I'm happy for that.

Sure. I have a PNA-X (N5247A) and an 8753 I can try things out on at some point too.

arsenovic:
i dont follow your psuedo code exactly, but i think i get the idea. 

there are a lot of subtle issues i have run into with instruments, like timing, need to cache large get's, non-uniformity of commands between vendors, etc. but the current state of drivers is mostly ad-hoc and i think a refactoring would increase the uniformity, thereby removing a lot of wasted code.  

[...]

PR #97, opened 2017-01-18
Updates to the Analyzer Driver

dvincentwest:
Keysight PNA class created, with a templating scheme for SCPI commands.
sweep method of the driver accomodates specifiying, one or more channels, where all channels will be swept by default.  A new sweep will take into account the averaging state of the analyzer and if no visa timeout is specified then will query the analyzer for a sweep time and autoset the timeout to accomodate long and short sweeping channels without having to worry about it

two subdrivers --> E8363C, N5247A added, simply changing NPORTS and visa addresses, and they both work.

skrf.mptstyle file updated to use prop_cycle instead of color_cycle in accord with new matplotlib conventions


Ttl:
Shouldn't the analyzer code be under a different directory? They can be used also without the Qt stuff.

There is already some existing analyzer code in "skrf/vi/" directory.

dvincentwest:
Ideally you are correct that this should be a merged effort. The problem is that the vi code had some significant incompatibilities with what I was trying to do. Also it relies on IVI and also an older version of pyvisa. I wanted to be able to really hammer on the code while I am developing the Qt apps and also have an infrastructure where custom drivers are automatically detected and loaded into the widgets.  Any thoughts or suggestions?

Ttl:
I don't know much about the vi code, but doesn't the vna_pyvisa.py implement the PNA class with same library that you are using?

It doesn't sound like a good idea to have two different implementations for controlling a VNA. Would the other analyzers also need to be written twice if someone wants to use it both with and without qtapps?

Could they somehow be merged in vi directory?

dvincentwest:
Those are all fair points.  It seems that writing a general purpose vi driver is quite challenging to do well, and most of us have a lot of other things we are trying to accomplish.  Even simple things start to get hairy pretty fast (a simple get_snp call fails if you don't have a measurement selected on the proper channel).  My vision for the qtapps was (at least at the beginning) that vi code would be narrowly focused on pulling data from the analyzer, rather than controlling the analyzer.  As such, the required functions to implement are pretty small, and could feasibly be based upon a more general driver implemented in the main skrf library to make the requirements even smaller.  The hope was that driver development for the qtapps would be a low a barrier as possible to actually using it for data acquisition and analysis.

That being said, it may be time for a fresh vi module implementation.  Even in the vi code itself @arsenovic commented that the code is pretty spotty.  Furthermore, I would say I think the Keysight_PNA driver that I've written works equally well in the notebook as it does in the qtapps, and worked for my purposes on both a 2-port PNA and a 4-port PNA-X which was encouraging.  In fact, if you have a H.Agi.Sight PNA(X) I would be grateful if you would mind running the driver test notebook that I created:
skrf_qtwidgets/analyzers/driver development/Keysight PNA.ipynb
and let me know if it works out of the box.  I've tried to implement some of the comments from #96 and maybe this is an opportunity to make a new vi module, but it will certainly require input and testing from others.  I think that probably the vision should be leaving the old module there for people who are using it, but if more work is going to be done, starting fresh should be considered.

the vna_pyvisa.py code I made based on the vna.py module and was working with before I broke out the qtapps.  I think it snuck into the main repository on accident.  (pretty sure that is an isolated incident).

arsenovic:
eventually there should be one `skrf.vi`, but no harm in experimenting.

your templating design  is really interesting.  it would be useful to all instruments  which use scpi commands, and might belong outside of skrf (like in pyvisa or lantz). you could also use a templating engine like jinja.  if you dont have time for this you may just raise and issue and point to your code. 

there is also an ability to upload skrf calibration coef's to the PNA, which would make sense to add to your TRL app.

helgrind:
I'll see if I can test on our 5247A tomorrow. I've got Hewlett Agisight coming in tomorrow to check a dodgy service calibration they did so I'll be playing with it anyway.

helgrind:
While we're on a qtapps subject, I assume it won't play with Windows at the moment? I already have anaconda installed on there.. Failed with cairo (obviously) and dbus :(

dvincentwest:
You have tried running data_grabber.py? I use it on both Windows and Linux Mint without problems. VISA license libraries are actually tricky on Deb based Linux.

helgrind:
image
I'm getting an unterminated error on the PNA-X. Keysight IO Monitor debug attached.

helgrind:
image
Success after I added the termination lines to the INSTR as well. I think it had an issue with those periods before the newline character.

dvincentwest:
@helgrind .  Thanks very much for testing, I've just uploaded a significant update to the driver that accomodates multiple channels on the instrument in triggering new sweeps on a channel by channel basis and autocalculating the timeout.

According to your commend, I changed the def __init__ method of the base analyzer, ~line 80 to be the following:
```
        self.resource.read_termination = "\n"  # most queries are terminated with a newline
        self.resource.write_termination = "\n"
        if "instr" in resource_string.lower():
            self.resource.control_ren(2)
```

so that all write_terminations are "\n".  I didn't have a problem with this on anything I've tested so far, but that's why you need more than one tester I suppose.  Thanks very much

helgrind:
Now I have more time, I just wanted to say great work on the plotting! Looks just like the Keysight style, even the Smith charts!

I'm going to try and have a look through the cal. interfaces soon as I am working on that side of software for nonlinear measurements at the moment. Might have some input.

dvincentwest:
Thanks for the compliments. @Ttl recently added NIST multi-line TRL to the cal package which is something I had wanted to use for a long time but was a bit outside my skill set. I will aim to add a gui interface soon and any input would be greatly appreciated.

Issue #103, opened 2017-02-17
py36 doesnt work


arsenovic:

RossWilliamson:
I cannot import skrf in python3.6:

import skfr as rf given:
`/home/rossw/anaconda/envs/py3k/lib/python3.6/site-packages/scikit_rf-0.14.3-py3.6.egg/skrf/network.py in read_touchstone(self, filename)
   1408         if self.name is None:
   1409             try:
-> 1410                 self.name = os.path.basename( os.path.splitext(filename)[0])
   1411                 # this may not work if filename is a file object
   1412             except(AttributeError):

/home/rossw/anaconda/envs/py3k/lib/python3.6/posixpath.py in splitext(p)
    118 
    119 def splitext(p):
--> 120     p = os.fspath(p)
    121     if isinstance(p, bytes):
    122         sep = b'/'

TypeError: expected str, bytes or os.PathLike object, not _io.TextIOWrapper
`

arsenovic:
this got fixed. py36 is now tested on travis.

PR #106, opened 2017-02-22
Passing the filename as a string instead of fid solves python36 bug

jhillairet:
This refers to issue #103.

Passing the file name as string instead of a fid object allows to use os.path method on it. 
However I wonder why is was like that before (and why it worked)...  

Is there a py36 tests automatically run on travis?

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -148,7 +148,6 @@ except ImportError:
     import pickle as pickle
     from pickle import UnpicklingError
 
-import warnings
 from copy import deepcopy as copy
 import re
 from numbers import Number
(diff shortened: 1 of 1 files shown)


jhillairet:
Certainly. I've just didn't understood in which cases this should break. Do you have example to suggest to help me?

arsenovic:
so it looks like the functionality i was advising you to keep doesnt even work! so your fix is fine, and we can remove the ability to pass an `fid` to Network().  i will merge you commit and fix the docstrings. sorry this took so long.

arsenovic:
ok i actually spoke too soon. it appears that 
* loading from str works 
*  loading from pickle fid works, but
*  loading from a touchstone fid fails.

 ill fix this, as its the easiest solution.

PR #109, opened 2017-03-01
Add 12-term isolation calibration

Ttl:
Finish the already almost working isolation term calibration in TwelveTerm and add tests.

Tests pass and I also verified this with real measurement data.


arsenovic:
sweeet

PR #118, opened 2017-03-20
Save/Load ZipArchive for NISTTRL Cal

dvincentwest:
provides a template for implementing a detailed save/load that preserves original data and parameters while also saving the cal coefficients and gamma for quick calibration use


dvincentwest:
circular import must be resolved first

PR #121, opened 2017-04-12
Save/Load cal archive for NISTTRL, QtApps updates and Cal Upload for Keysight PNA

dvincentwest:
Several updates for QtApps

NIST_TRL calibration now has a save archive function that saves all the calibration coefficients, original standards measurements and parameters to a zipped archive.  Can reload all of these back from the zip archive to rapidly create a new calibration object

Upload a calibration object to Keysight PNA, works with an 8 term converted to a 12 term
added write_values support to the scpi parser to achieve this


skrf/io/touchstone.py
```diff
@@ -471,3 +474,27 @@ def hfss_touchstone_2_network(filename, f_unit='ghz'):
...
+    -------
+    dict
+    """
+    networks = dict()
+    for fname in ziparchive.namelist():  # type: str
+        directory, filename = os.path.split(fname)
+        if dir == directory and fname[-4:].lower() in (".s1p", ".s2p", ".s3p", ".s4p"):
+            network = Network.zipped_touchstone(fname, ziparchive)
+            networks[network.name] = network
+    return networks
```
arsenovic: it would be best if this was in NetworkSet.from_zip, or alternatively, if the networkset.from_zip() called this function.

arsenovic:
looks like the build is failing due to a versioning issue with py27 and ipython, and not your code. 

nice additions.

dvincentwest:
Is it building properly for other peoples' pull requests?  Should I be changing something somewhere?

skrf/io/touchstone.py
dvincentwest: okay, I will look at your latter proposal since there may be uses for this function that are independent of the scope of NetworkSet.

arsenovic:
this may have happened due to  a change in ipython notebook. since we dont have version info in the requirements, it gets the latest, and this may break at any moment.

skrf/io/touchstone.py
arsenovic: yes i agree. i think we should also remove requirement that NetworkSet be homogenous (same #ports, frequency, etc). this might be moved to a 'check' that happens before statistics are computed.

Issue #123, opened 2017-04-14
Add isolation (crosstalk) to EightTerm Class of calibrations, 


arsenovic:
The class of EightTerm calibrations should allow for  isolation correction terms. 
Obviously, this would make the class more appropriately named `TenTerm`!, but  i think leaving it as EightTerm is ok, even though it's misleading. 

see #122,

dvincentwest:
How do you want to do this?  There are 4 classes of EightTerm Coefficients, and they each define their own self.run method where the coefs dictionaries are defined.

Two proposals:
1) put it in EightTerm.coefs getter
    If no isolation terms are detected, initialize at zero and return
2) put it in the EightTerm..coefs setter
    If not isolation terms are detected, initalize at zero and set the _coefs attribute with this

Or something else?

arsenovic:
so  i think we want to 

 1) change the init to look for isolation temrs as a kwarg to `EightTerm`, so that we could use it in the correction. 
 2) use it in the correction (`EightTerm.apply_cal`). this may require terminate/unterminate  to be updated as @Ttl  suggests. 
  3) use it in `EightTerm.embed`. (low priority, but needed for tests)
  3) have default zero-filling fall-back behavior as you suggest above.

arsenovic:
also putting the isolation in coefs getter, as you suggest solves the problem is redundant implementation, which  is important . after looking the `EightTerm.run` directly accesses self._coefs, so it bypasses the setter. so going with the getter makes the most sense.

dvincentwest:
are we ready to close this then?

arsenovic:
Yep

PR #124, opened 2017-04-16
Isolation calibration for EightTerm

Ttl:
Fixes bug #123. Implementation was actually easier than I initially thought and there are very few new lines of code. Basically isolation calibration can be done outside of the EightTerm calibration so no deep changes were needed.

Isolation calibration can also be used with other calibrations that inherit from EightTerm such as TRL, UnknownThru and NISTMultilineTRL. Isolation calibration tests were added also for these calibrations.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -133,7 +133,9 @@ coefs_list_8term = [
     'reverse reflection tracking',
     'forward switch term',
     'reverse switch term',
-    'k'
+    'k',
+    'forward isolation',
+    'reverse isolation'
     ]
 global coefs_list_3term
 coefs_list_3term = [
@@ -721,7 +723,6 @@ class Calibration(object):
         Err = self.coefs_12term['reverse reflection tracking']
         Etr = self.coefs_12term['reverse transmission tracking']
         Esr = self.coefs_12term['reverse source match']
-
 
         return Etf*Etr - (Err + Edr*(Elf - Esr))*(Erf  + Edf *(Elr - Esf))
 
@@ -1966,7 +1967,7 @@ class EightTerm(Calibration):
     '''
     family = 'EightTerm'
     def __init__(self, measured, ideals, switch_terms=None,
-                 *args, **kwargs):
+                isolation=None, *args, **kwargs):
         '''
         EightTerm Initializer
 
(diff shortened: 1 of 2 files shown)


Ttl:
Now that I took a look at the code, I don't think this method of isolation correction is completely correct. Currently in the tests isolation terms are added after the switch termination, but I think they should be added before the switch correction. Although error is probably quite small it would be good to fix it. In the "Multiline TRL Revealed" paper are listed the changes needed to be made to the switch correction to take isolation into account. I'll try to fix it later.

Ttl:
I think that this is a correct way to implement the isolation calibration after all. It should be equivalent to the TwelveTerm isolation calibration. It does give the same results according to the "TwelveTermToEightTermTest"-test. It is not equivalent to the forward/reverse isolation terms in the SixteenTerm unless switch terms are zero, but that doesn't matter since there are also other isolation terms in the 16-term model that are not corrected with this simple isolation calibration.

This model is also given in the "Formulations of the Basic Vector Network Analyzer Error Model including Switch-Terms" paper, so I think it is okay.

I'm not sure why in the Multiline TRL paper switch terms are also modified. I don't seem to get the correct results using it. Any ideas?

arsenovic:
thanks for working on this, i will look into mark's paper tomorrow and get back to you on my opinion.

skrf/calibration/calibration.py
```diff
@@ -1993,12 +1998,32 @@ def __init__(self, measured, ideals, switch_terms=None,
...
+            self.isolation = measured[0].copy()
+            self.isolation.s[:,:,:] = 0
+        else:
+            self.isolation = isolation.copy()
+
+            measured = [m.copy() for m in measured]
+            for i in range(len(measured)):
+                measured[i].s[:,1,0] -= self.isolation.s[:,1,0]
+                measured[i].s[:,0,1] -= self.isolation.s[:,0,1]
+
```
arsenovic: this seems to correct for isolation during the init, which i think is unexpected. seems more logical that isolation is corrected for in `apply_cal`. also, i would not expect the `measured` to change. 

maybe we could make a property  `measured_unisolated` similar to `measured_terminated`. then `measured_unterminated` would `unterminate(measured_unisolated)`

Ttl:
I think that is a better way to do it. I just added a commit to implement it.

By the way, it seems that documentation building has been broken for some time. Seems to be because of tab when spaces are expected in "readthedocs-environment.yml".

arsenovic:
great!

arsenovic:
i will  look into docs momentarily

arsenovic:
it looks like the read the docs says the build fails, but they appear to be working.

PR #129, opened 2017-04-26
fft transform fix and vna driver bug fixes

dvincentwest:
1) In the network.py module transforming to the time domain and back to the frequency domain was using ifftshift and fftshift incorrectly

2) some bug fixes to the vna drivers fixing the behavior of uploading calibration sets to Keysight PNA


arsenovic:
Vince, some of your changes were in conflict with the new gating algorithm, so could not add them. you may want to test it out, and either re-add the changes or not.

PR #135, opened 2017-06-16
Fix bug in plot_uncertainty_bounds_components

Ttl:
attribute was modified when using dB ppf, but this resulted in incorrect values on the next loop iterations.

Also warnings wasn't imported and the code doesn't work so I replaced it with NotImplentedError.

### DIFF
--- a/skrf/plotting.py
+++ b/skrf/plotting.py
@@ -1541,6 +1541,8 @@ def plot_uncertainty_bounds_component(self,attribute,m=None,n=None,\
     for m in M:
         for n in N:
 
+            plot_attribute = attribute
+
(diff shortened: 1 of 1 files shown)


arsenovic:
looks like the waveguide class is not passing some test now, but it is likely some upstream change and not related to this PR.

arsenovic:
(accidentally closed/re-opened)

Ttl:
Failing test is because of a new numpy release. Array == None now does element wise comparison. It has actually complained about that before.

I think == should be replaced with "is".

arsenovic:
ok thanks, ill check it out.

PR #136, opened 2017-06-20
Added cases for empty networks

BendettaSD:
I found myself working with 'empty' networks (from subnetworks of other networks). The two places I found where empty networks cause trouble are modified below. I added two conditionals to avoid exceptions from empty networks.

Please let me know if any changes are necessary. Thank you.


arsenovic:
great!

PR #143, opened 2017-07-30
generalisation of s2t() and t2s() to 2N-ports balanced networks

mhuser:
network.s2t() and network.t2s() are used by network.inv() to compute the invert of a network, which is useful for deembedding purpose.

These methods where limited to 2-ports network, thus limiting the deembedding application to 2-ports networks.

The proposed implementation is valid for 2N-ports networks with same number of input and output ports, also known as balanced networks. This enable network.inv() be used for deembedding of 2N-ports balanced networks without restriction.

Use connect to connect the networks.

Port numbering convention:
```
     +---------+
    -|0      N |-
    -|1     N+1|-
In  ...       ... Out
    -|N-2  2N-2|-
    -|N-1  2N-1|-
     +---------+
```

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3528,12 +3528,13 @@ def s2t(s):
     Converts scattering parameters [#]_ to scattering transfer parameters [#]_ .
 
     transfer parameters are also refered to as
-    'wave cascading matrix', this function only operates on 2-port
-    networks.
+    'wave cascading matrix', this function only operates on 2N-ports
+    networks with same number of input and output ports, also known as
+    'balanced networks'.
 
     Parameters
     -----------
-    s : :class:`numpy.ndarray` (shape fx2x2)
+    s : :class:`numpy.ndarray` (shape fx2nx2n)
         scattering parameter matrix
 
     Returns
@@ -3566,15 +3567,28 @@ def s2t(s):
     -----------
     .. [#] http://en.wikipedia.org/wiki/S-parameters
     .. [#] http://en.wikipedia.org/wiki/Scattering_transfer_parameters#Scattering_transfer_parameters
+    .. [#] Janusz A. Dobrowolski, "Scattering Parameter in RF and Microwave Circuit Analysis and Design",
+           Artech House, 2016, pp. 65-68
     """
-    # TODO: check rank(s) ==2
-
-    t = npy.array([
-        [-1 * (s[:, 0, 0] * s[:, 1, 1] - s[:, 1, 0] * s[:, 0, 1]) / s[:, 1, 0],
-         -s[:, 1, 1] / s[:, 1, 0]],
-        [s[:, 0, 0] / s[:, 1, 0],
-         1. / s[:, 1, 0]]
-    ]).transpose()
+    z, y, x = s.shape
+    # test here for even number of ports.
+    # s-parameter networks are square matrix, so x and y are equal.
+    if(x % 2 != 0):
+        raise IndexError('Network don\'t have an even number of ports')     
+    t = npy.zeros((z, y, x), dtype=complex)
+    yh = int(y/2)
+    xh = int(x/2)
+    # S_II,I^-1
+    sinv = npy.linalg.inv(s[:, yh:y, 0:xh])
+    # np.linalg.inv test for singularity (matrix not invertible)
+    # T_I,I = S_I,II - S_I,I * S_II,I^-1 * S_II,II
+    t[:, 0:yh, 0:xh] = s[:, 0:yh, xh:x]-s[:, 0:yh, 0:xh]*sinv*s[:, yh:y, xh:x]
+    # T_I,II = S_I,I * S_II,I^-1
+    t[:, 0:yh, xh:x] = s[:, 0:yh, 0:xh]*sinv
+    # T_II,I = -S_II,I^-1 * S_II,II
+    t[:, yh:y, 0:xh] = -sinv*s[:, yh:y, xh:x]
+    # T_II,II = S_II,I^-1
+    t[:, yh:y, xh:x] = sinv
     return t
 
 
@@ -3989,13 +4003,13 @@ def t2s(t):
     converts scattering transfer parameters [#]_ to scattering parameters [#]_
 
     transfer parameters are also referred to as
-    'wave cascading matrix', this function only operates on 2-port
-    networks. this function only operates on 2-port scattering
-    parameters.
+    'wave cascading matrix', this function only operates on 2N-ports
+    networks with same number of input and output ports, also known as
+    'balanced networks'.
 
     Parameters
     -----------
-    t : :class:`numpy.ndarray` (shape fx2x2)
+    t : :class:`numpy.ndarray` (shape fx2nx2n)
             scattering transfer parameters
 
     Returns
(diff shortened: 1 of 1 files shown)


mhuser:
As suggested by @arsenovic in an email exchange, we could later add a `connect_bal` method and modify `Network.__pow__` in such a way 2N-ports networks may be cascaded quick and easy.

Example:
```
DUT = connect_bal(F.inv, Meas)
DUT = F.inv ** Meas

        F.inv         Meas             DUT
     +---------+   +---------+     +---------+
    -|0      N |---|0      N |-   -|0      N |-
    -|1     N+1|---|1     N+1|-   -|1     N+1|-
    ...       ... ...       ... = ...       ...
    -|N-2  2N-2|---|N-2  2N-2|-   -|N-2  2N-2|-
    -|N-1  2N-1|---|N-1  2N-1|-   -|N-1  2N-1|-
     +---------+   +---------+     +---------+
```

arsenovic:
great PR!
i just altered the `network.inv` and `Network.__pow__` to use balanced networks when appropriate as we discussed .  see  96b34b4. 


could add a test for this functionality?

arsenovic:
also, we should add your ascii art to the docs. it is good.

mhuser:
thank you for the feedback and further implementation!

For the test:
shall we add 3x 8-ports balanced networks from .s8p files to the test directory such as  Meas = Fix ** DUT, and complete the test in the same idea as 2-port version? I can generate and check these .s8p data by simulation.

For the doc:
is it sufficient to add it to docstring in `network.py` and regenerate the doc?

arsenovic:
yes to both of your suggestions. 
 
i did not add a `connect_bal` method  because, after consideration, it did not seem needed. 

also  i think i should move the functionality i put in `Network.__pow__`  to `network.cascade`, and call it from within __pow__. this will consolidate things a bit.

arsenovic:
hey mhuser, just curious if you have added the test for this?

i added your ascii art to [here](http://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.network.cascade.html#skrf.network.cascade) but the sphinx web output doesnt  preserve it. i found an extension to sphinx called [aafigure](https://pythonhosted.org/sphinxcontrib-aafig/), which we could add  to make it nicer.

mhuser:
the tests already exists for the 1-port case, so at last it works and we did not break anything.
i have performed some test with 4-port ntwks, but the M = t2s(s2(N)) fail to give M==N if there is coupling between the inputs or the outputs. function well for parallel isolated paths but somewhat disappointing.

for the ascii art, nice finding! an alternative way would be to escape them as source code with
```
::
    and then indent as a bloc
```

mhuser:
added a test using concatenated `skrf.media.CPW` lines, see #147

arsenovic:
thanks for adding the test, and fixing the art!

is this *full rank* limitation an known feature of the algorithm or a programming error?  it is an important problem that we need to handle or the users could unknowingly get incorrect results. we should mention this limitation in the docs and raise a warning, or perhaps an Error if the s-matrix is not isolated. 

i can confirm this fails,
```
from skrf import wr10,ZERO
a = wr10.random(n_ports =4)
s = a.s[:1]
assert((abs(rf.s2t(rf.t2s(s)) -s) < ZERO).all())
```
i could also add a failing test for this, (that is expected to fail) so that if we ever fix it, it will pass. 


also i guess we expect a 'working' version to produce 
```
(a.inv**a).s = I | 0
               0 | I 
```
regardless of the number of ports , where I is identity matrix?

mhuser:
thank you for the feedback!

the _full rank_ limitation is not clearly  mentioned in the literature i have seen, but neither is there examples with coupling. i have cross compared two separated source that give the same algorithm.

nevertheless,  the 8-port deembedding bloc of ADS and it is getting better results in most cases (without unveiling their algorithm). so there is room for better results...

So agree to the idea of a failing test, and a warning for the user, hoping one day we shall overcome this!
In the meantime, we shall inform the user of the current limitations we are facing. It is still a good way to apply deembedding of well-isolated 2N ports, which may be of some use.

we should have,
```
rf.s2t((a.inv**a).s) = I
```
regardless the even number of ports, where I is the identity matrix
Currently I am expecting this results for isolated balanced networks.

I am thinking to the 12-term error model of VNA and the coefficient used to isolation calibration. left-right fixture deembeding could be achieved for 2-port network by tweaking these (in former time with less software features), but the isolation coefficient remained unaffected. it is possible to measure the fixture isolation and deduct it as an additive effect (parallel). maybe will i have to solve this by hand for 2-port and 4-port network to see what happens.

mhuser:
here are some updates:

after derivation by hand the equations of the book for the well-known 2-port case and for the 2n-port case with analogy to 2-port cascading, the results are the same.

however, i missed a statement in the book, that the connection between the networks are assumed to be reflexion-less.

After some trials in an algebra software, it appears that with the current algorithm, `t2s(s2t(s))==s` is valid only with zero return loss on all ports, zero transmission in-between inputs and zero transmission in-between outputs, which sounds quite restrictive.

In some case with big isolation and small return loss, it is wrong only by a small amount, which may be confused with numerical precision, but is still wrong.

looks like the 2-port analogy is not applicable to the general case. this is not very surprising, because there is no isolation between input ports concept in the 2-port case, having a single input. still the phase and amplitude of the thru components will get corrected.

arsenovic:
ok let me clarify some terms and my understanding. so  if we have an 2N port s-matrix divided into quadrants  A,B,C,D as such, 
```
    A|B
S = ---
    C|D
```
then i interpret  *isolated* to  mean A,B,C,D are all diagonal. this does not mean *matched*. matched would be A=D=0, and B,C are diagonal. 

for an isolated, but not matched 2N port, you should still be able to determine the inverse s-matrix, because you could pull out each relevant 2-port and use the existing algorithm.  for example, given a balanced 6-port, with ports 
```
1 --- 3
2 --- 5
3 --- 6
```
you can  pull out three 2-ports of the port-pairs (1,3), (2,5), and (3,6)  and t2s(inv(s2t())) them, and put them back together.  so  the zero-reflection constraint seems unnecessary to me.

mhuser:
thank you, it is always good to get the ideas more clear and insist with simple and efficient examples.

based on your example, which is obvious, i found a small error in my verification computations in the algebra software (wrong verifications leading to bad conclusions).

Now the so-called Maxima software give me `t2s(s2t(s))==s` for any 4port, which is better (but seems too good). However this is not true with the python implementation.

I am going to compare a numerical application with the python code step by step to find the difference.

mhuser:
ok, seems i have mad a silly mess with * instead of matrix multiplication. let me tackle this.

mhuser:
see #153 ,
you may double-check this solution with
```
from skrf import wr10,ZERO
A = wr10.random(n_ports =4)
s = A.s[:1]
sp = rf.t2s(rf.s2t(s))
assert((abs(sp -s) < ZERO).all())
```

mhuser:
i have compared the result of the `Network.inv` method after #153 against ADS `de_embed` block on a 4-port simulated differential line and the difference of the resulting s-parameters is lower than a 1e-8, which is a reasonable agreement.

arsenovic:
sweet. good to get to the bottom of it. ill test some more and get back to you.

mhuser:
I have just written a short draft of an example with the removal of connectors capacitance effect.
[2n-port_deembeding_example.zip](https://github.com/scikit-rf/scikit-rf/files/1208355/2n-port_deembeding_example.zip)
```
            Left        Meas         Right
             l1          l2           l1
    0----+-=======-2 0-=======-2 0-=======-+----2
         |                                 |
         = c1                              = c1
         |                                 |
        GND                               GND
        
             l1          l2           l1
    1----+-=======-3 1-=======-3 1-=======-+----3
         |                                 |
         = c1                              = c1
         |                                 |
        GND                               GND
```
2n-port_dembedding_example_draft

Measurement was salted with some noise to looks more realistic. Of course, examples works often better than real measurements, because the fixture model and not the true fixture effect is embedded in the simulated measurement. 

I used nudge to add some crosstalk, but it did not seemed to me the right way to do.

arsenovic:
great example! i converted it into notebook format and added it to the docs. a663c9d   . feel free to edit, and PR. 

you could  manually specify crosstalk values with a loop, but i think nudge is ok. 

also this example shows that  it would be nice if the port numbering from `concat_ports` aligned with what is expected from `cascade`. we could change this.

mhuser:
Looks great! Many thanks.

For the `concat_port`, we might use the existing `port_order` parameter to choose between the two numbering scheme and keep backward compatibility.

As a matter of fact, the possibility is already implemented, but I've got an error when I tried.
`TypeError: unsupported operand type(s) for +: 'range' and 'range'`
by looking to it more carefully, it is only that `range()` does not return a list in python 3.

I will fix this

mhuser:
done in #156 

Maybe we should extend the `flip` method, giving as well a choice between the two numbering scheme.
Currently is is limited to 2-port, but may be useful in case of symmetric left-right fixture.

arsenovic:
yes thats a great idea.

mhuser:
sounds like a plan.

I have a proposition with #157

PR #145, opened 2017-08-02
Fix  __floordiv__ operator

JesterEE:
Previous implementation with a single inputs caused a frequency mismatch.  If only a single network was passed, the unpacking of `other` into `b` and `c` would slice the network into 2 single frequency point networks and cause the operation to fail.  This checks the input to decide how to operate (either with 1 network [removing `other` from `self`], or 2 networks removing `other[0]` and `other[1]` from opposite sides of `self`]).


arsenovic:
good fix. can you correct the conflicts to the tests can be run on travis?

JesterEE:
@arsenovic I had a really old version of `scikit-rf` in my repos before I forked today.  Github did the right thing and didn't squash my repo with the new one and I unknowingly edited a really old version of `network.py`.  I need to rebase my fork, but I'll need to look at this a little later.

arsenovic:
yeah i hate git. if it takes too long, i can just add your fix, and you can re-clone.

arsenovic:
looks like it is still conflicting

JesterEE:
Give me another couple days ... I'll clean it up

JesterEE:
Automatically closed by Github when I rebased my fork.  See #154.

PR #148, opened 2017-08-04
Network constructor parse HFSS Touchstone file

jhillairet:
The Network() constructor detects if the touchstone file provided has been generated from HFSS. If so, it tries to import the gamma and z0 data if available. 

With the present PR, some of the functions like hfss_touchstone_2_ become no more really usefull, as Network do the job. 

Some convenience functions have been added into the Touchstone object, in order to detect is the file has been generated by HFSS or to extract gamma and z0 data if available (if not, use the defaut reference impedance).


arsenovic:
great job.

PR #153, opened 2017-08-07
fixed s2p and t2s for 2n-port

mhuser:


mhuser:
seems travis dislike the  `@` operator

PR #154, opened 2017-08-07
Fix for __floordiv__

JesterEE:
If a single network was passed in, the logic would create 2 single frequency networks for the de-embedding.  This would cause a frequency inconsistency error (since the de-embedding network would be the 'correct' size, and the networks to be de-embedded would be 'incorrect').  New logic determines if either a single network or a list/tuple of 2 networks is given as an input and operates accordingly.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -422,37 +422,57 @@ class Network(object):
             out = self.copy()
             out.s = out.s ** other
             return out
-        
+
         else:
             return cascade(self, other)
-            
+
     def __floordiv__(self, other):
         """
-        de-embeding another network[s], from this network
+        de-embedding 1 or 2 network[s], from this network
+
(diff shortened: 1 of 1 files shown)


JesterEE:
Needed to open a new PR since the last one (#145) was closed automatically when I rebased my old fork.  The other file changes are due to my IDE (Pycharm) which automatically removes extraneous whitespace on save.  I didn't mean to commit those changes, but it did anyway.  Rather than undoing the clean-up I figured I'd let @arsenovic decide.

JesterEE:
Travis failure is unrelated to this PR.  It's same as the current master commit failure 97fe5d40b202df17f3164b5e9199c95ccfd088e2 [Travis](https://travis-ci.org/scikit-rf/scikit-rf/builds/261080445?utm_source=github_status&utm_medium=notification).

No errors in test_network.py

PR #155, opened 2017-08-08
Fix determine_reflect

Ttl:
Use determine_line to make the root choice for the propagation constant.


arsenovic:
great! thanks for the help!

PR #156, opened 2017-08-08
Fixed concat_ports port_order='second' for python 3.5

mhuser:
Modifier Balanced example accordingly to use `concat_ports(port_order='second')` instead of `renumber`


arsenovic:
maybe we should make the default order the one which works with cascade. what do you think?

mhuser:
I would not argue with that as it lead for my usage to more compact code.

There is some sense in making the default behaviour being consistent with the `**` operator, however.

Of course, this idea may be reconsidered if the entire `concat_ports` community rise in rebellion...

PR #158, opened 2017-08-09
Testcase to excercise get_sparameter_data.

jstenar:
get_sparameter_names use the xrange function which is not available
in python3. Switching to range function.


arsenovic:
hi jorgen, thanks. and good to hear from you!

PR #160, opened 2017-08-11
Remove test warnings

jstenar:
The purpose for this pull request is to show my progress in removing the test warnings and to have a place to disucss any issues during this work, 

It is NOT ready for merge yet.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -1438,16 +1438,16 @@ class TwelveTerm(Calibration):
 
         # if they didnt tell us the number of thrus, then lets
         # hueristcally determine it
+        trans_thres_mag = 10 ** (trans_thres / 20)
 
         if n_thrus is None:
             warn('n_thrus is None, guessing which stds are transmissive')
             n_thrus=0
             for k in self.ideals:
                 mean_trans = NetworkSet([k.s21, k.s12]).mean_s_mag
-                trans_db = npy.mean(mean_trans.s_db.flatten())
-
+                trans_mag = npy.mean(mean_trans.s_mag.flatten())
                 # this number is arbitrary but reasonable
-                if trans_db > trans_thres and not npy.isneginf(trans_db).all():
+                if trans_mag > trans_thres_mag:
                     n_thrus +=1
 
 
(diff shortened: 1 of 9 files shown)


jstenar:
There are a bunch of warnings during the TRL calibration tests. As far as I can see they are caused by the from_coefs calss method on the Calibration class. This method calls the TRL class with the measured data for the standards set to the same network. Causing math warnings in the determine_reflect method (the value of the a variable becomes zero).

Wouldn't it be better to run the determine_line and determine_reflect in the run method?

Or should I just silence these warnings?

arsenovic:
jorgen,  
 all of your fixes thus far look good to me. as for moving the determine functions into TRL.run. so, to make sure i understand: 
*  the init will setup all the ideals if the user gave None values, or +-1 for reflects, 
*  call EightTerm.__init__. 

then run() will call the determine methods. this sounds good to me.

arsenovic:
nice changes, let me know when you want to merge.

skrf/calibration/calibration.py
```diff
@@ -1438,16 +1438,16 @@ def __init__(self, measured, ideals, n_thrus=None, trans_thres=-40,
...
             warn('n_thrus is None, guessing which stds are transmissive')
             n_thrus=0
             for k in self.ideals:
                 mean_trans = NetworkSet([k.s21, k.s12]).mean_s_mag
-                trans_db = npy.mean(mean_trans.s_db.flatten())
-
+                trans_mag = npy.mean(mean_trans.s_mag.flatten())
                 # this number is arbitrary but reasonable
-                if trans_db > trans_thres and not npy.isneginf(trans_db).all():
+                if trans_mag > trans_thres_mag:
```
jstenar: @arsenovic I'm a little concerned about this change in the criteria. But I only have the testsuite to test against at that works. But if you could try it on some more cases it would be great.

jstenar:
I have no more changes planned so feel free to merge.

JesterEE:
It seems odd, but following [PEP8](https://www.python.org/dev/peps/pep-0008/) guidelines:
"Comparisons to singletons like None should always be done with is or is not, never the equality operators."

https://github.com/scikit-rf/scikit-rf/pull/160/commits/9b98171f42e4cdc9ed56e728401d1c714db63de5 might want to stay as is for compliance if that's a thing we want to shoot for.  Programmatically speaking, either should work so it's up to whatever programming paradigm the scikit-rf project want's to follow.

arsenovic:
no problem JesterEE, i appreciate you reading it!

PR #164, opened 2017-08-17
Remove unnecessary restriction from s_to_time

Ttl:
This works on networks with multiple ports and restriction broke time gating example.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -175,8 +175,6 @@ from .constants import ZERO
 
 def s_to_time(s, pad=0):
     """Transforms S-parameters to time-domain"""
-    if len(s.shape) != 1 and s.shape[1:] != (1,1):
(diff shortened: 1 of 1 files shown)


arsenovic:
i think i put this restriction in there because i thought that gating a s11 and s21 at the same time was an unlikely need, and perhaps would accidentally be used improperly. however, gating the DC either band-pass or stop would be a practical application.  are there any others?

Ttl:
This was introduced by me in the last pull request and didn't exist before. The function was more complicated earlier and I left this in by accident. This commit restores the earlier behaviour.

Currently the time gating example in docs is broken because of this.

arsenovic:
ahh, ok i was confused. i added a similar restriction to time gating, for the reasons i stated above.

PR #165, opened 2017-08-18
Add extrapolate_to_dc function

Ttl:
Extrapolates the low frequency points needed for time-domain transformations, when measurement doesn't include DC.

Example:

    import skrf
    skrf.stylely()

    freq = skrf.F(0.11,110,1001)
    coax1mm = skrf.media.Coaxial(freq, z0=50, Dint=0.44e-3, Dout=1.0e-3, sigma=1e20)

    X = coax1mm.line(10, 'mm', z0=50, name='X', embed=True)
    Y = coax1mm.line(80, 'mm', z0=75, name='Y', embed=True)
    dut = X**Y**X

    dut_dc = dut.extrapolate_to_dc(dc_sparam=[[0,1],[1,0]])

    plt.figure()
    plt.title('Step')
    t, y = dut.s11.step_response(pad=2000)
    t2, y2 = dut_dc.s11.step_response(pad=2000)
    plt.plot(t, y, label='Original')
    plt.plot(t2, y2, label='Extrapolated')
    plt.legend()
    plt.xlabel('Time (ns)')

    plt.figure()
    plt.title('Impulse')
    t, y = dut.s11.impulse_response(pad=2000)
    t2, y2 = dut_dc.s11.impulse_response(pad=2000)
    plt.plot(t, y, label='Original')
    plt.plot(t2, y2, label='Extrapolated')
    plt.legend()
    plt.xlabel('Time (ns)')

    plt.show(block=True)

step

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1996,6 +1996,86 @@ class Network(object):
         # freq = Frequency.from_f(f,**kwargs)
         # self.interpolate_self(freq, **interp_kwargs)
 
+    def extrapolate_to_dc(self, points=None, dc_sparam=None, kind='linear',
+            coords='polar', **kwargs):
+        """
+        Extrapolate S-parameters down to 0 Hz and interpolate to uniform spacing.
+
+        If frequency vector needs to be interpolated aliasing will occur in
+        time-domain. For the best results first frequency point should be a
+        multiple of the frequency step so that points from DC to
+        the first measured point can be added without interpolating rest of the
+        frequency points.
+
+        Parameters
+        -----------
+        points : int or None
+            Number of frequency points to be used in interpolation.
+            If None number of points is calculated based on the frequency step size
+            and spacing between 0 Hz and first measured frequency point.
+        dc_sparam : class:`numpy.ndarray` or None
+            NxN S-parameters matrix at 0 Hz.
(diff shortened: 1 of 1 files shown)


Ttl:
None of the existing interpolation methods are too good at time-domain transformations and cause aliasing in time-domain. AWR has rational interpolation (http://kb.awr.com/display/Examples/Interpolation_Example) that interpolates using poles in S-plane and works much better with time-domain transformations. It would be good to have something similar in skrf, but it's too much work compared to how much I care at the moment.

Enforcing passivity would also be a good option when extrapolating. Actually it would be good idea to have a function to force any network passive. Maybe I'll write one when I'm bored.

arsenovic:
that is interesting,  skrf has a polar/cartesian and s/y/z option  for interpolation.  it would be cool to do an example comparing all of these to the `rational` method. i have not heard of this before.

Ttl:
I found the paper about rational interpolation (https://www.in.tu-clausthal.de/fileadmin/homes/techreports/ifi0606floater.pdf) though I have no idea if this is that same method that AWR uses. It seemed simple enough that I started to implement it. I'll see if I get it to work.

arsenovic:
i added your example to the docs/examples section , let me know if you want to add/change anything.

PR #169, opened 2017-08-25
added impulse and step response plotting methods

mhuser:
First attempt to add plotting methods for impulse and step response function from @Ttl 
This solution does not satisfy me yet.

To be added directly to the Network class `COMPONENT_FUNC_DICT` dictionnary, the method has to be declared first, which sounds a bit impractical. The time axis has to be built in plotting.y for accounting to variable number of points depending on zero padding.

trough a dirty workaround, it is now possible to write:
```
plt.figure()
plt.title('Impulse Response Lowpass')
(dut2.s11.windowed()).plot_s_time(label='impulse bandpass')
dut2.s11.plot_s_time_impulse(pad=2000, window='hamming', label='impulse lowpass')
dut2.s11.plot_s_time_step(pad=2000, window='hamming', label='step lowpass')
plt.show(block=True)
```
example_plot_step

Although it would be possible to just add and option to `Network.plot_s_time_step(z0=None)`, it would be nicer to have a more general methods in such a way e.g. `Network.plot_z_time_step()` would plot the step response with an impedance y-axis.

### DIFF
--- a/doc/source/examples/networktheory/Rational Interpolation.ipynb
+++ b/doc/source/examples/networktheory/Rational Interpolation.ipynb
@@ -40,32 +40,53 @@
     "X2 = coax1mm2.line(10, 'mm', z0=50, name='X', embed=True)\n",
     "Y2 = coax1mm2.line(80, 'mm', z0=75, name='Y', embed=True)\n",
     "dut2 = X2**Y2**X2\n",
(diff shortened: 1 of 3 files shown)


arsenovic:
it is the plotting functions of skrf that is wonky, so your addition is not to blame.

PR #172, opened 2017-09-03
added microstripline media example

mhuser:


mhuser:
Not sure this is at its right location in `networktheory`, but i did not figure a better place at the moment.

If required, I might replace the picture of the DUTs by a dimensionnal drawing.

arsenovic:
great,  i like the pic. 

i could make a folder in examples for  'media', and put the "properties of rectangular waveguide' example in that as well.

JesterEE:
@mhuser I don't have time to go over all your analysis in the notebook, but after a cursory look, it looks like you are showcasing something I have done a few times before without `skrf` as an aid.  Great work and I look forward to using it in the future!

Ttl:
@mhuser Nice example. If you want to subtract the connector effects better you can use NISTMultilineTRL calibration to extract the propagation constant of the line. Reflect measurement is not necessary as it's not used in calculating the propagation constant. In a case like this where you have calibrated the VNA you can actually guess or simulate the reflect S-parameters to get usable calibration too.

If you then fit the model to solved the Z0 and propagation constant, it should be much more accurate.

Here's some code to do the Multiline TRL: https://gist.github.com/Ttl/1baac84423225c6edfafdadd37f0994e

Explanation of the characteristic impedance calculation: http://hforsten.com/trl-measurements-with-homemade-vna-and-open-source-software.html

And results:

atten

erf_eff

z0

connector

Spikes in the connector S-parameters are artifacts caused by the phase of the lines being a multiple of 180 degrees resulting in poor accuracy. Note that connector includes 50 mm of transmission line. Plot cal.nstd to see the calibration accuracy.

mhuser:
Thank you for your feedback guys!

Do you think a generic physical transmission line media such as distributed RLGC or DefinedGammaZ0 but initiated with Z0, ep_reff, A_db, tand and length would make sense?

It is indeed another view of the same concept, but may be useful to fit a generic transmission line. As an example: in the case we have measured some line but the physical dimensions are uncertain (they always get wronged by the manufacturing process to some extent).

I have also measured the two CPWG lines and am willing to apply the same process to them at some point.

@arsenovic Sounds like a plan!

@JesterEE I am happy if you can reuse and extend some of this, you are of course welcome to add your personal touch to the example!

@Ttl Thank you for the code! I looked to your website and it is great! I am willing to investigate this Multiline TRL method next weekend, it seems a lot more straightforward to get the base parameter to fit the model, as well as the connector s-parameters.

arsenovic:
@Ttl  great suggestions,  

@mhuser    i think your suggestion is interesting.   you could just make it a thin inherited class from one of the 'basic' classes you mentioned, like Freespace is. i dont see how length would be used for the 
media init, seems to me  like providing loss in dB/m makes more sense, but up to the implementor! 

also, perhaps this  line characterization processes is become refined enough to warrant a tool?

ill move docs  around after this is finished, no rush.

mhuser:
@arsenovic of course no length, it is a mistake of me. Length is a property of the line and not the media.

This may sounds a silly question, but what is a tool?

There is still a bit guessing for the connector in the process, but I am looking to @Ttl method to get rid of that.

mhuser:
Could not resist at the attraction of trying `rf.NISTMultilineTRL` on the CPWG lines measurements as a draft, see  #174 . Seems to works quite good but I still need to works on the characteristic impedance determination.

Ideally, I should also create the CPWG media as well for those at some point.

PR #174, opened 2017-09-04
Added DefinedAEpTandZ0 medium

mhuser:
Add DefinedAEpTandZ0 transmission line medium
Transmission line medium defined by A, Ep, Tand and Z0.

This medium is defined by attenuation A, relative permittivity Ep_r,
loss angle tand and characteristic impedance Z0.

Djirdjevic/Svennson dispersion model is provided for dielectric, default
behaviour is frequency invariant.

Example with DefinedAEpTandZ0 model compared to CPWG measurements.


skrf/media/physicalLine.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+'''
+from scipy.constants import  epsilon_0, c
+from numpy import real, imag, sqrt, ones, zeros, pi, log
+from .media import Media
+
+class PhysicalLine(Media):
+    '''
+    A generic physical transmission line media.
```
arsenovic: this looks like the old docstring, and doesnt explain your new parameters

skrf/media/physicalLine.py
```diff
@@ -0,0 +1,177 @@
...
+
+        .. math::
+                Z_0 = \\sqrt{ \\frac{Z^{'}}{Y^{'}}}
+
+        Returns
+        --------
+        Z0 : numpy.ndarray
+                Characteristic Impedance in units of ohms
+        '''
+        return zeros(len(self))
```
arsenovic: i dont understand this?

skrf/media/physicalLine.py
```diff
@@ -0,0 +1,177 @@
...
+Frequency invariant parameters or Djirdjevic/Svennson dispersion is provided. 
+
+
+
+'''
+from scipy.constants import  epsilon_0, c
+from numpy import real, imag, sqrt, ones, zeros, pi, log
+from .media import Media
+
+class PhysicalLine(Media):
```
arsenovic: Perhaps we could come up with a more descriptive name?  physical line could equally well be used for  RLCG. i dont have any great ideas,  perhaps Z0AEp ?

mhuser: indeed, almost all the transmission lines are so-called physical lines, or it would be difficult to manufacture them.

What's about DefinedEpTandZ0? It would keep the spirit of DefinedGammaZ0.

skrf/media/physicalLine.py
mhuser: Oops. Will fix that this weekend.

skrf/media/physicalLine.py
mhuser: I still have to finish this. This is more difficult because the impedance dispersion model depend of the geometry, and there is no geometry in this media implementation.
Maybe we should let the user provide a single value (no dispersion) or a vector, letting him compute his own dispersion.

By the way, it is required to write media.line(l, 'm', embed=True, z0=y) to build a line of impedance y which is normalized with media z0. Is there a specific reason not to have embed=True the default behaviour? I am thinking in the direction of getting the same results when plotting the line as if measuring with a VNA.

skrf/media/physicalLine.py
arsenovic: yeah good point!  that works

skrf/media/physicalLine.py
arsenovic: ok. perhaps henrik might have a good idea, 

on the second point, it is a matter of taste.  i like to explicitly add connectors, so i frequently do, 
    wg.thru()**wg.line(...z0='2')**wg.thru()

even though you could just do 

    wg.line(z0='2', embed=1)

also if you are making a taper, or a filter, ie  cascading many lines of various impedances, embedding to a `system` or `port` impedance for each element is a waste of cpu.

mhuser: Ok, I got the idea. Indeed in most simulation softwares, the ports are added with their characteristic impedance. The vna ports have intrinsically their own port impedance too. The implemention now make sense to me.

skrf/media/physicalLine.py
mhuser: Done in   7f6cc27

skrf/media/physicalLine.py
mhuser: Docstring updated in 9bf9ca9

skrf/media/physicalLine.py
mhuser: I just kept the characteristic impedance constant for time being. The media.thru() notation is now used in the example.

PR #177, opened 2017-09-13
add R&S ZVA instrument support

beuerle:
I made changes to the SCPI parser to handle an additional `help` field in the
YAML file, to insert this help as a block comment in the function description.

I already added quite some commands to the YAML file, but it is still work in
progress. Not all meta functions from `keysight_pna.py` are currently
implemented in `rs_zva.py`.


dvincentwest:
Excellent, looks very promising and thanks for the 'help' addition to the parser

arsenovic:
great addition. unfortunately i have no way to test this.

PR #181, opened 2017-09-28
merge fixed ZVA class and additional ZVA SCPI commands

### LINKED ISSUE #178: Remove default measurement parameter assignment from initialization of VNA class
Currently some measurement settings are applied during the initialization of
the VNA class in [line 117](https://github.com/scikit-rf/scikit-rf/blob/131771a128989b0eada0a5474cb891c4596ddd86/skrf/vi/vna/abcvna.py#L117) of `abcvna.py`. The settings are so far hard-coded in the VNA class in [line 53](https://github.com/scikit-rf/scikit-rf/blob/131771a128989b0eada0a5474cb891c4596ddd86/skrf/vi/vna/abcvna.py#L53).

Those settings should be given as function parameters during initialization, and only applied if set (or removed from `__init__` completely?).

As an addition a parameter to reset the tool when initialized (default: `False`) should be included.

```python
if kwargs['reset']:
    self.reset()
```

I can work on that, if there's a general agreement on how this should be handled. Especially taking out the assignment of the measurement parameters might be an issue for some, but could be worked around somehow.

beuerle:
This pull request fixes #178 (see discussion there). The non-working functions copied from PNA are fixed or removed.

More SCPI commands are defined in the YAML file.

Example instantiation:

```python
from skrf.vi.vna import ZVA


class VNA(ZVA):
    ADDRESS = '192.168.1.83'
    VISA = '@py'
    INTERFACE = 'SOCKET'

    def __init__(self, address=ADDRESS, visa=VISA, **kwargs):
        """Initialization function"""
        super(VNA, self).__init__(address=address, visa_library=visa,
                                  interface=self.INTERFACE)

        if return_bool(kwargs.get('id', False)):
            print(self.idn)

        if return_bool(kwargs.get('reset', True)):
            self.reset()

        if kwargs.get('converter', False):
            self.set_converter_name(kwargs['converter'])

        if return_bool(kwargs.get('display', False)):
            self.set_display_update(UPDATE=1)
        elif not return_bool(kwargs.get('display', True)):
            self.set_display_update(UPDATE=0)

        if return_bool(kwargs.get('correction', False)):
            self.set_corr_state(STATE=1)
        elif not return_bool(kwargs.get('correction', True)):
            self.set_corr_state(STATE=0)


def return_bool(arg):
    """return boolean value for input"""
    if type(arg) is str:
        arg = arg.upper()
    true_set = [True, 1, 'ON', 'YES', 'Y']
    if arg in true_set:
        return True
    else:
        return False
```

some function testing:

```python
In [2]: from vna import VNA

In [3]: v = VNA()
Rohde&Schwarz,ZVA24-4Port,1145111028101513,3.60

In [16]: v.idn
Out[16]: 'Rohde&Schwarz,ZVA24-4Port,1145111028101513,3.60'

In [17]: v.use_ascii()

In [18]: v.scpi.query_format_data()
Out[18]: 'ASC,0'

In [19]: v.use_binary()

In [20]: v.scpi.query_format_data()
Out[20]: 'REAL,64'

In [21]: v.scpi.query_format_binary()
Out[21]: 'SWAP'

In [23]: v.set_frequency_sweep(220e9, 325e9, 201)

In [24]: v.get_frequency()
Out[24]: 220000000000.0-325000000000.0 Hz, 201 pts

In [32]: v.active_channel = 2
Channel 2 not in list of channels. Create channel first

In [33]: v.active_channel = 1

In [34]: v.scpi.set_channel_state(cnum=3)

In [35]: v.active_channel = 3

In [36]: v.active_channel
Out[36]: 3
```

I will continue to work on implementing meta-functions in the ZVA class (new functions, and try to make it compatible to the PNA class) and add additional SCPI functions in the YAML if needed.

### DIFF
--- a/skrf/vi/vna/abcvna.py
+++ b/skrf/vi/vna/abcvna.py
@@ -138,6 +138,9 @@ class VNA(object):
     def idn(self):
         return self.query("*IDN?")
 
+    def reset(self):
+        self.write("*RST")
+
     def wait_until_finished(self):
         self.query("*OPC?")
 
--- a/skrf/vi/vna/abcvna.py
+++ b/skrf/vi/vna/abcvna.py
@@ -188,10 +188,10 @@ class VNA(object):
 
     @property
     def idn(self):
-        return self.resource.query("*IDN?")
+        return self.query("*IDN?")
 
     def wait_until_finished(self):
-        self.resource.query("*OPC?")
+        self.query("*OPC?")
 
     def get_list_of_traces(self, **kwargs):
         """
--- a/skrf/vi/vna/abcvna.py
+++ b/skrf/vi/vna/abcvna.py
@@ -48,16 +48,8 @@ class VNA(object):
     get_switch_terms
     set_frequency_sweep
     """
-    DEFAULT_VISA_ADDRESS = "GPIB::16::INSTR"
-    NAME = "Two Port Analyzer"
-    NPORTS = 2
-    NCHANNELS = 32
-    MEASUREMENT_PARAMETERS = {
-        "port1": 1, "port2": 2, "ports": (1, 2),
-        "raw_data": True, "sweep": False,
-    }
 
-    def __init__(self, address=DEFAULT_VISA_ADDRESS, **kwargs):
+    def __init__(self, address, **kwargs):
         """
         Initialize a network analyzer object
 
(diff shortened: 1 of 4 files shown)


arsenovic:
@dvincentwest  would you like to review and merge this?

dvincentwest:
this all looks good, and is a great direction.  I have not tested yet but I am pretty sure the default parameters aspects will break some of the qtapps programs.  That's okay right now because I'm pretty sure I am the only one using the NIST TRL and the DataGrabber applications.  I will try to fix those immediately.  In the meantime, I see no reason to hold this back.

PR #182, opened 2017-10-17
Correct the K_ratio function.

edumur:
Basically replace k by its complement, sqrt(1-k**2) when k<1/sqrt(2).

### DIFF
--- a/skrf/media/cpw.py
+++ b/skrf/media/cpw.py
@@ -46,10 +46,10 @@ class CPW(Media):
             resistivity of conductor (None)
 
     '''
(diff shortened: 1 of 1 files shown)


arsenovic:
hey sorry  this took so long. 
i dont have a way to verify this is correct. 
can you post a reference?

edumur:
Sure.
Easiest reference is [there](http://qucs.sourceforge.net/tech/node86.html) Eq. (12.4).
It has been a long time but I think I checked before commit.

arsenovic:
thanks, sorry for delay!

PR #184, opened 2017-10-31
Increase support for R&S ZVA

dvincentwest:
Add the following functions to R&Z ZVA:
- get_snp_network (enables get_oneport and get_twoport from base class)
- get_active_trace_as_network

Also changes to the SCPI parser so that everything is read and written with utf8 encoding
Some minor bug fixes as well

@beuerle, @torgovanov - I would appreciate a little testing with your equipment if possible as I have a ZVA here that is running Windows XP, and a little broader testing base would be a good thing.

image


beuerle:
Thanks for the additions and fixes. I will test it in the coming days when I'm back in the measurement lab.

beuerle:
All good.

PR #187, opened 2017-11-05
Update colors() for matplotlib 2.0

JesterEE:
matplotlib was issuing a `MatplotlibDeprecationWarning` for the use of
`rcParams['axes.color_cycle']`.  This uses the Cycler object to accomplish the same task.


JesterEE:
@arsenovic I think it might be better to implement this with `itertools.cycle` or  [cycler](http://matplotlib.org/cycler) but I didn't know if it would break anything or what the testing suite covers with plotting.

PR #194, opened 2018-02-07
add IEEE designators (WM-***) to instances.py

beuerle:
Complement the Hesler designators with the IEEE designators for waveguide bands.


arsenovic:
good idea.

Issue #195, opened 2018-02-11
Time-domain transform issues


ToBeXu:
Hi, I just started trying out this package, and it seems quite useful for S-parameter analysis. However, I am having some issues with the time-domain conversion in `Network.impulse_response()` and `Network.step_response()`. 

First, `s_to_time()` does not seem to correctly set the argument `n` for the output length to `numpy.irfft()`.

```python
def s_to_time(s, pad=0):
    """Transforms S-parameters to time-domain"""
    return npy.fft.fftshift(npy.fft.irfft(s, axis=0, n=s.shape[0]+pad), axes=0)
```

Per the docstring:

>>>
    n : int, optional
        Length of the transformed axis of the output.
        For `n` output points, ``n//2+1`` input points are necessary.  If the
        input is longer than this, it is cropped.  If it is shorter than this,
        it is padded with zeros.  If `n` is not given, it is determined from
        the length of the input along the axis specified by `axis`.
>>>

Take the simple case where `pad=0`. Then since we set `n` equal to the number of frequencies in `s`, the inverse FFT will crop the S-parameter to only use half of the available data. Similarly, we can see that `pad` does not equal the number of zero-padding points, since the first `s.shape[0]-2` `pad` points go toward retaining the original S-parameter.

Second, the time vector returned by the time-domain methods does not appear to be correct.

```python
t = npy.linspace(-.5/fstep , .5/fstep, self.frequency.npoints+pad)
```
The time per sample in the time-domain is `1 / 2 / f_max`, where `f_max = fstep * (npoints - 1)` is the maximum frequency in the frequency-domain and `npoints` includes the DC point and positive frequencies. But the above `linspace()` divides the interval `1 / fstep` into `npoints` points, which not only excludes the factor of 2 from ignoring the negative frequency points but also has an off by 1 error.

PR #197, opened 2018-02-13
Fix time-domain conversions

ToBeXu:
Fixes and unit test for #195. Use the expected number of frequency and pad points, and return the correct time vector points and interval spacing. Allow an odd number of points in the output time-domain response by specifying an explicit 'n' argument.

### DIFF
--- a/skrf/tests/test_network.py
+++ b/skrf/tests/test_network.py
@@ -55,6 +55,31 @@ class NetworkTestCase(unittest.TestCase):
         t = self.ntwk1.s11.s_time
         self.assertTrue(npy.sum(npy.abs(t.imag)) == 0)
 
+    def test_time_transform(self):
+        spb = (4, 5)
+        data_rate = 5e9
+        num_taps = (100, 101)
+        for i in range(2):
+            tps = 1 / spb[i] / data_rate
+            num_points = spb[i] * num_taps[i]
+            # Frequency terms should NOT contain Nyquist frequency if number of points is odd
+            inc_nyq = True if num_points % 2 == 0 else False
+            freq = npy.linspace(0, 1 / 2 / tps, num_points // 2 + 1, endpoint=inc_nyq)
+
+            dut = self.ntwk1.copy()
+            freq_valid = freq[npy.logical_and(freq >= dut.f[0], freq <= dut.f[-1])]
+            dut.interpolate_self(rf.Frequency.from_f(freq_valid, unit='hz'))
+
+            dut_dc = dut.extrapolate_to_dc()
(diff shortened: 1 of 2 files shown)


arsenovic:
@mhuser  , could you comment on this change to your time domain additions?

skrf/tests/test_network.py
```diff
@@ -55,6 +55,31 @@ def test_timedomain(self):
...
+    def test_time_transform(self):
+        spb = (4, 5)
+        data_rate = 5e9
+        num_taps = (100, 101)
+        for i in range(2):
+            tps = 1 / spb[i] / data_rate
+            num_points = spb[i] * num_taps[i]
+            # Frequency terms should NOT contain Nyquist frequency if number of points is odd
+            inc_nyq = True if num_points % 2 == 0 else False
+            freq = npy.linspace(0, 1 / 2 / tps, num_points // 2 + 1, endpoint=inc_nyq)
```
arsenovic: thanks for making this test. 
1/2 evaluates to 0 in py2.7.  i think you need to use 1.0/2 or similar for force float division.  you can see the results in travis https://travis-ci.org/scikit-rf/scikit-rf/jobs/340844075

mhuser:
@ToBeXu findings are correct. @Ttl and myself misinterpreted the n parameter for irfft at this place.
Indeed the irfft inputs are only zero and positive complex frequency terms (complex conjugates are taken for negative terms to achieve real value results), but n is the size of the output, not the input.
The consistency of the modification can be tested with `doc/source/examples/networktheory/Rational Interpolation.ipynb` (seems I need to fix the time axis to display 's' instead of 'ns').

ToBeXu:
@arsenovic thanks for the reminder (too used to working in Python 3). I've updated line 67 as well as line 63 to correctly perform floating point division in Python 2.

arsenovic:
great thanks for the fix and tests.

arsenovic:
can anyone verify that  this error in the docs related to this change? 

http://scikit-rf.readthedocs.io/en/latest/examples/networktheory/Time%20Domain.html

mhuser:
I will look at that.
Le mardi 06 mars 2018 à 16:52 +0000, alex arsenovic a écrit :

mhuser:
Ok, I've got it. Now the number of returned points is mathematically correct but is now twice the number of the network frequencies (complex conjugate is assumed for values corresponding to negative frequency), with the special case of odd number of points.

Let's see how to fix that.

mhuser:
Done. Could you check if you get the expected results with #204 ?

PR #198, opened 2018-02-17
TDR example based on measurements

mhuser:
Some new artefacts where manufactured in an attempt to provide a measurement-based time domain reflectometry example.


mhuser:
I shall come back soon with VNA measurements and an ipython notebook.

arsenovic:
ok i will wait to merge then.

mhuser:
Hi, unfortunately not yet, but I will try to put the finishing touch this weekend!

mhuser:
This is done, but unfortunately it became almost a monster and possibly missed its initial target.

Issue #29, opened 2014-05-20
vna.py: log sweeps with PNA


AntonGut:
Similar to https://github.com/scikit-rf/scikit-rf/issues/15 for different hardware.

In PNA class frequency vector gets generated using f_start/f_stop/f_npoints, which assumes linear sweep. For log sweeps frequency vector will be wrong.

No fix from me just yet since I don't need log sweep ATM. 

An idea for implementation: use first column from CALCulate<cnum>:DATA:SNP:PORTs? reply, possibly inside get_data_snp()

DavidLutton:
For generating log spaced steps [numpy.geomspace — NumPy v1.12 Manual](https://docs.scipy.org/doc/numpy-1.12.0/reference/generated/numpy.geomspace.html)

`numpy.geomspace(start, stop, num=50, endpoint=True, dtype=None)`

PR #203, opened 2018-03-06
New get_frequency implementation

jstenar:
Issue #29 mentions that LOG frequency sweeps are not handled properly and  could be done using np.logspace. This has been fixed recently but there is still another case that is not handled by get_frequency and that is the segmented sweep. A simpler solution is to use the SENS:X? command to read a vector of the x-axis sweep data. But a special case is needed for CW and POW that has constant frequency.

The new implementation use SENS:X:VALUE to get the frequency data when the sweep type is LIN, LOG or SEGM. Otherwise we just query_f_start, POW, CW, use a constant frequency.

I'm new to the scpi.yaml stuff so I may have misunderstood something.

### DIFF
--- a/skrf/vi/vna/keysight_pna.py
+++ b/skrf/vi/vna/keysight_pna.py
@@ -366,16 +366,13 @@ class PNA(abcvna.VNA):
         """
         self.resource.clear()
         channel = kwargs.get("channel", self.active_channel)
-        use_log = "LOG" in self.scpi.query_sweep_type(channel).upper()
-        f_start = self.scpi.query_f_start(channel)
-        f_stop = self.scpi.query_f_stop(channel)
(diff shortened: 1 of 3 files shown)


arsenovic:
thanks jorgen,

dvincentwest:
Hi Jorgen, this all looks correct.  I had meant to weigh in yesterday on
one aspect which is a better name for the the function.  Rather than
"sweep_data" either "sweep_frequencies" or "sweep_x_data" seems more
informative.  sweep_data sounds more like the sparameters.

jstenar:
Good point, I would prefer sweep_x_data because the data can also be power if you are doing a power sweep. I can make a new PR with that change unless anyone else has a better suggestion for the name.

PR #204, opened 2018-03-06
Fix network.plot_s_db_time() issue

mhuser:
Fix `network.plot_s_db_time()` issue induced by #197
- `frequency.t` and `frequency.t_ns` range is doubled to be coherent with irfft results.
- `network.windowed` is centered around dc point to produce the correct impulse response.

### DIFF
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -518,7 +518,14 @@ class Frequency(object):
 
         t_period = 1/f_step
         '''
-        return linspace(-.5/self.step , .5/self.step, self.npoints)
+        n = 2 * (self.npoints - 1)
+        if n % 2 == 0:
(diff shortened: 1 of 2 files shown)


arsenovic:
thanks for fixing this. it looks like the zero time component is zeroed out, which i dont recall happening. 
see plot here
http://scikit-rf.readthedocs.io/en/latest/examples/networktheory/Time%20Domain.html#Gating-The-Reflection-of-Interest

is this another bug?

(i also dont know why the plots in the docs are so small i will look into it)

mhuser:
Strictly speaking, it is not a bug, though, I do not think it is the way we want it.
This come from the use of irfft without a dc point. A workaround  is to limit the number of points of the result in such a way it correspond to the usual ifft (preventing the complex conjugates to be taken for negative frequencies)... or just to use ifft directly in this case and take only the real part of the result.
I will provide a new fix for that.

PR #205, opened 2018-03-10
Yet another time domain fix

mhuser:
Fix `network.plot_s_db_time()` and other time plotting functions from behaviour change induced by #204  and #197.


mhuser:
Maybe we should rename `mf.s_to_time()` and `mf.s_to_time_irfft()` into `mf.ifft()` and `mf.irfft()` where `mf` is the `mathFunctions` module.

arsenovic:
these  make sense to me.

PR #209, opened 2018-03-23
bugfix timedomain,use  abs() [instead of nothing or real]

arsenovic:
the was a real() part taken here, which i removed, but i think it should actually be mag, ie abs()

### DIFF
--- a/skrf/mathFunctions.py
+++ b/skrf/mathFunctions.py
@@ -625,7 +625,7 @@ def ifft(x):
     """
     Transforms S-parameters to time-domain bandpass.
     """
-    return npy.fft.fftshift(npy.fft.ifft(x, axis=0), axes=0)
+    return npy.abs(npy.fft.fftshift(npy.fft.ifft(x, axis=0), axes=0))
 
 def irfft(x, n=None):
     """


mhuser:
Looks great!
In Keysight app note 5989-5723EN "Time Domain Analysis Using a Network Analyzer" we can read:
about time domain bandpass (p. 13)
> The IFT is calculated only on the data points measured, rather than taking the 
negative frequency response to be the conjugate of the measured data as in 
the low-pass mode. This calculation gives a complex (real and imaginary parts) 
time domain response and the magnitude (linear or log mag) of the response is 
the most commonly displayed.

and about time domain lowpass (p.12) which is `step_response` and `impulse response`
>Since the frequency domain data is taken at harmonically related frequencies down to DC, 
the IFT has only a real part (the imaginary part is zero). Therefore, the most useful 
format is Real, which displays the response in reflection coefficient, rho.

mhuser:
Seems I erroneously get stuck using real part for tdr bandpass here.

arsenovic:
yeah, we might want to keep the complex time-domain in the bandpass mode, for who knows what we will do with it?  maybe we should change the test not to enforce s_time.imag  ==0?

mhuser:
Yes, I agree fully with you.

It turns out as a side effect that the `test_timedomain` in its current form has no more use then.

Eventually we could check that it has the same number of points as `frequency.t_ns` instead, but I am unsure if it is really useful.

arsenovic:
yes i agree. 
i think that ensuring the length of time-domain or time-gate is the same would have caught a few bugs, i will modify tests

PR #212, opened 2018-04-08
Additions to Network: is_reciprocal, is_passive, is_symmetric...

TomWoo:
To fulfill the first item on the TODO list, I made the following additions to network:
+ implemented is_reciprocal, is_passive, is_symmetric, and is_lossless (network) with accompanying matrix functions (mathFunctions)
+ added tests for those network classifiers (test_network)

I also updated gitignore to exclude IntelliJ idea's project file.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -173,6 +173,8 @@ from .time import time_gate
 
 from .constants import ZERO
 
+from typing import Union, Iterable, Dict, Optional
+
 class Network(object):
     """
     A n-port electrical network [#]_.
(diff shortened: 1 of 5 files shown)


skrf/network.py
```diff
@@ -1305,30 +1305,53 @@ def group_delay(self):
...
-    ## NETWORK CLASIFIERs
+    ## NETWORK CLASSIFIERs
     def is_reciprocal(self):
         '''
         test for reciprocity
         '''
-        raise (NotImplementedError)
+        return self.is_symmetric() and self.is_passive()
 
     def is_symmetric(self):
```
arsenovic: this is incorrect.

arsenovic:
thanks for adding these, they look good.   

there is a problem with the symmetry definition:

the term symmetric  in regards to microwave networks means it's matrix fufills `s_11 == s_22`.    this is not the same as the usual  term symmetric in regards to matrices, which means `S.Transpose() == S`, and is usually defined only for real matrices. the equivalent of symmetric for complex matrices is hermitian.  so, your test for symmetry is not what people will expect. 

but the symmetry test gets more complicated; how do we extend the concept of  symmetric  to n-ports? i would say that 1-plane symmetry is only meaningful for balanced networks, ie has 2n-ports, because you have to be able to know where to draw the symmetry line (plane) and then which ports to test for equallity. thus,  there is not a unique test for first order symmetry for a 2n-port network.    for higher order symmetry, like a branchline coupler,  we have second order symmetry for a 4-port network, and we can extend the simple test;  just do `s_ii ==s_jj` for all i,j.  

so what. well i would think we need to  have arguments in `is_symmetric` to define what `order` symmetry we want to test for,  and if `2*n-ports > order` we need to have a port-order argument which can default to some convention (probably should look at how @mhuser  did balanced network cascading.)

while we are at this you might have a tolerance parameters passable to all these tests. up to you.

TomWoo:
Thanks for your feedback. I conflated network symmetry with reciprocity, and I think the best way for me to fix is_reciprocal is to use the reciprocity() metric that's already been written + tolerancing.

Before I reimplement is_symmetric, I'd like your opinion on whether the behavior described in the following docstring might be a better way:
```
    def is_symmetric(self, p0=None):
        # type: (Union[int, Iterable(int)]) -> bool
        '''
        :param p0: nearest port(s) (based on default zero-indexed numbering scheme)
        from which the line of symmetry is draw (see notes for details)
        :return: whether network is symmetric for all indicated lines of symmetry
        Tests for n-order symmetry in a 2n-port network.

        Notes: This function checks for symmetry against line(s) connecting these endpoints:
        (1) between ports p0-1 and p0
        (2) between ports p0+n-1 and p0+n (diametrically opposite side of the network)
        automatically wrapping around port index between 0 and 2n whenever necessary.
        Using the default p0 results in all 2n orders of symmetry being checked
        (s_ii == s_jj for all i, j <= 2n).

        https://en.wikipedia.org/wiki/Two-port_network#Scattering_parameters_(S-parameters)
        '''
```
I propose the above method because it allows the caller to select arbitrary lines of symmetry, as well as pass in a consecutive range of port indices with a known order (length of that range) if desired. Then, there's no need to make 2n > order a special case. I agree with all your other points.

Optional tolerance params for all tests is a good idea, so I'll add those.

TomWoo:
Apologies for the delayed response. While drawing out some examples, I realized there's no need for the level of flexibility I described... We know exactly where the lines of symmetry are given a port-order argument (to figure out where to draw the first line) and the order to check for. (Looks like I need to retake grade-school geometry!)

I made my docstrings conform to numpy-style, and I think the one for is_symmetric() is consistent with your description. I still want to write some tests for the symmetry test in case I discover any bugs (it was trickier to write than I thought), as well as refactor it for readability, so hang tight.

TomWoo:
I'm happy with my changes now. Let me know what you think.

skrf/network.py
```diff
@@ -4609,7 +4675,7 @@ def reciprocity(s):
 
 ## renormalize
 def renormalize_s(s, z_old, z_new):
-    '''
+    '''wave casca
```
arsenovic: typo?

skrf/tests/test_network.py
```diff
@@ -309,5 +309,126 @@ def test_slicer(self):
...
+        s8p_diag = [1j, -1j, -1j, 1j, 1j, -1j, -1j, 1j]
+        s8p_mat = npy.identity(8, dtype=complex)
+        for row in range(8):
+            s8p_mat[row, :] *= s8p_diag[row]
+        y = rf.Network(f=[1, 2],
+                       s=s8p_mat,
+                       z0=50)
+        self.assertTrue(y.is_symmetric())
+        self.assertTrue(y.is_symmetric(n=2))
+        self.assertFalse(y.is_symmetric(n=4))
```
arsenovic: these tests are very well thought out  and nicely implemented.

skrf/network.py
```diff
@@ -1305,30 +1306,95 @@ def group_delay(self):
...
         '''
         test for reciprocity
         '''
-        raise (NotImplementedError)
+        return npy.allclose(reciprocity(self.s), npy.zeros_like(self.s), atol=tol)
 
-    def is_symmetric(self):
-        '''
-        test for symmetry
+    def is_symmetric(self, n=1, port_order={}, tol=mf.ALMOST_ZERO):
```
arsenovic: i think any method of expressing  symmetry will be complicated due to the number of possibilities, and confusions with port indecies and numbering. if this feature becomes popular we may consider writing an example notebook in the  docs to demonstrate it.  i think you tests below are good at this as well.

skrf/network.py
TomWoo: Good catch. I'll fix that on my next commit.

PR #221, opened 2018-08-20
Update time.py (time_gate function)

ykonter:
I've started using skrf recently and find it very intuitive to use! Thanks very much for making this package available.

I'd like to recommend some changes to the time_gate function.
I get strange results using start and stop for the time_gate. I believe line 234 should be:
center = (stop + start)/2  # no minus sign

Also line 266 I believe should be fftshift(ifft(**i**fftshift(...etc in order to be compatible with an odd amount of datapoints. 
Shouldn't matter too much for the end-result though.
Also, using (i)fftshift here (line 266) implies that gate has it's zero phase component in the center of the dataset. This is not how gate is created in line 261. Nevertheless, since only the abs value is used (line 267), it should be sufficient to use only an ifft in line 266:
kernel = fft.ifft(gate)  # not sure, I did not test this (yet)

Cheers, 

Yuri

### DIFF
--- a/skrf/time.py
+++ b/skrf/time.py
@@ -231,7 +231,7 @@ def time_gate(ntwk, start=None, stop=None, center=None, span=None,
         start *= 1e-9
         stop *= 1e-9
         span = abs(stop-start)
-        center = (stop-start)/2.
+        center = (stop+start)/2.
     
     else:
         if center is None:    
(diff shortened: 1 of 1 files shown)


arsenovic:
good catch on center. the even odd npoints with ifft/fft came up a while ago. ... i forgot what the resolution was.  we currently dont have a test for this, so hard to verify whats correct.

PR #222, opened 2018-08-24
made Network.crop() work with bounds outside min/max of network

Gloamglozer:
Previously, the skrf.Network.crop() method would check if frequency bounds were within the extreme frequencies of the network without cropping. I think the behavior should be different -- e.g. let's say you're trying to crop a collection of Network objects to be within a certain frequency range -- If it so happens that you make a call to crop with your `f_start` and `f_stop` outside the extrema of a given network object, a ValueError is raised, where it might make more sense to just return the same skrf.Network without any frequencies removed. This works in my application, and is a relatively simple change. If need be, I am happy to create a corresponding unit test.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2164,7 +2164,8 @@ class Network(object):
         result.s[0,:,:] = result.s[0,:,:].real
         return result
 
-    def crop(self, f_start, f_stop):
+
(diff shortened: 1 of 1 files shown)


arsenovic:
looks good.

PR #228, opened 2018-10-23
fix IEEE waveguide designators

beuerle:
The nominal internal dimensions of the IEEE waveguide bands differ
slightly from the ones given by Hesler.

With my initial commit I simply copied the rectangular waveguide definitions from the Hesler ones. This pull request fixes this and gives the correct internal dimensions when IEEE waveguide bands are defined.

### DIFF
--- a/skrf/instances.py
+++ b/skrf/instances.py
@@ -106,6 +106,8 @@ f_wm250  = f_wr1
 f_wm200  = f_wr0p8
 f_wm164  = f_wr0p65
 f_wm130  = f_wr0p51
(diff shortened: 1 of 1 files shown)


arsenovic:
slight different specs! wtf.  ill ask hesler.  but i think going with the WM makes more sense,   so i think switching to the metric specs, and defining WR through them  is a good idea.

arsenovic:
ok, so it appears as though the WR and WM can be milled to different dimensions, so the original PR is good, thanks for adding this,

PR #240, opened 2019-01-09
Corrected few typos in the doc

jhillairet:
and a broken URL. 

However I'm not sure this is the correct way to update the notebook, since the original one had empty evaluated cells...


arsenovic:
hi julien, is there any way you could seperate the notebooks edits from this PR? i think you can create a new branch and then cherry pick them into it or something. 
my concern is that the notebooks have data in them, which will grow the repo. 

if you dont want to do this, i will just copy your changes on my branch and submit a PR.

jhillairet:
OK I'll try to do git magics.



Le ven. 8 févr. 2019 à 13:32, alex arsenovic <notifications@github.com> a
écrit :

PR #244, opened 2019-02-20
Matplotlib compatibility updates

jhillairet:
- updated the definition of the color cycler to the latest matplotlib API
- updated the setting RC properties according to the latest matplotlib API
this should avoid some warning when using `skrf.stylely()` function


arsenovic:
thanks!

PR #245, opened 2019-02-20
Interpolation of a Network from a set of network ("Multi-File N-port")

jhillairet:
First implementation of `interpolate_from_network` method of NetworkSet

This method allows to create the equivalent of ANSYS Designer "Multi-File N-Port Model", that is, a N-port network which is interpolated from a set of N-ports.

For example, this method is used below on an adjustable phase shifter in the band 1-2 GHz. The scattering parameters of the phase shifter has been measured on all its frequency band for 19 values of its tuning knob. The method then allows to interpolate the phase shifter network to other values of its tuning knob: 

phase_shifter_interpolated_network

I guess the formulation of the documentation could be improved. Should I add some assertion concerning the length of the parameter ntw_param (which should be equal to the length of the NetworkSet) ?


arsenovic:
this is great. it would be best if we could consolidate the existing interpolate functions (which has more features). those existing functions assume you want to interpolate over frequency as the independent variable, but there is no reason for this. 

anyway we can always do this in the future.

jhillairet:
Thanks.

 As static functions you mean?

I intend to make a dedicated notebook to document this interpolating
methods. I guess I should add few sNp files, is it a problem? Should I find
an other way?



Le mer. 20 févr. 2019 18:07, alex arsenovic <notifications@github.com> a
écrit :

arsenovic:
adding a few snp files is fine, as long as they are not modified and saved (which grows the repo). 

im not sure what you mean by static functions. i meant that interpolation methods should be agnostic to their independent variable (frequency, phase shift position, temperature ,etc) so we dont need redundant interpolation methods.

jhillairet:
ok. The one I proposed is agnostic (except from frequency since everything
already exist for that). A user just has to pass the parameter array which
have changed between the N-port networks. It can be temperature, phase
shift, stub position, etc...

I wonder however how to generalize to multi-parameters networks, in order
to perform design of experiments (DOE) analysis...

Le mer. 20 févr. 2019 à 19:07, alex arsenovic <notifications@github.com> a
écrit :

PR #246, opened 2019-02-21
Documentation example on network set interpolation

jhillairet:
Added a example in the documentation on the use of the network set in…terpolation. Added also figure and set of scattering parameters.


arsenovic:
looks nice, thanks

jhillairet:
Thanks. 

Is there something to do to update the documentation website to integrate this notebook?

arsenovic:
it should happen automatically,  not sure why it didnt, i just triggered it .

jhillairet:
Thanks. There are errors due to the fact it's a Python 2 engine (or old Python 3?). I can update the string handling accordingly.

arsenovic:
yeah i hate that .  i think the notebook should be python3 compatable, and the kernel should not be specified in the notebook.

PR #249, opened 2019-03-05
fix xlim error in RWG example

beuerle:
Frequencies are given in Hz, so the correct command would be

```
xlim(100e9, 1300e9)
```

However, this is not really necessary in this case.

This fixes the _empty plot_ in the documentary.


doc/source/examples/networktheory/Properties of Rectangular Waveguides.ipynb
```diff
@@ -112,7 +112,6 @@
     "xlabel('Frequency(GHz)')\n",
     "ylabel('Loss (dB/m)')\n",
     "title('Loss in Rectangular Waveguide (Au)');\n",
-    "xlim(100,1300)"
```
arsenovic: did you mean to add a 1e9 to this line?

Issue #251, opened 2019-03-20
Proposal: adding noise component to Network


cactorium:
Hello! Would there be any interest in adding noise information to `Network`? It looks like some s2p files already include some noise information along with the scattering parameters. This would be useful at least for me because I've been messing around with designing a transistor amplifier using this library. I'd be up for trying to code up some initial version of it, but I don't know exactly what you guys would like it to be designed. It seems like it'd involve a lot of mostly straight forward coding so that cascading networks works correctly, and it'd probably be nice if there were some way to calculate NF from it

arsenovic:
i think that is a great idea

TomWoo:
I'm very interested in this feature. Let me know if there's anything I can help with.

cactorium:
Awesome, and thanks! I'm not great at RF stuff so I'm definitely down for some help

Right now I'm actually confused about how to calculate the noise correlation matrix from the noise parameters. Qucs provides these formula: http://qucs.sourceforge.net/tech/node9.html , where I'm kind of stuck because I don't know where to get Tmin in c11, and Googling got me this one: https://www.cambridge.org/us/files/8313/6698/2373/HFIC_chapter_3_noise.pdf , where c11 is a lot simpler. Do you know which formula is better or more usable for this case? I'm converting to a correlation matrix because calculating network cascades is done using the correlation matrices instead of noise parameters, and I'm planning on using the ABCD matrix-based transform here: https://onlinelibrary.wiley.com/doi/pdf/10.1002/9781119073093.app3

There's also the question of how to store the noise data, since it uses an independent set of frequencies than the scattering parameters. Right now I'm thinking of just keeping them separate and interpolating during cascades or when .interpolate() is called. Does that seem like a sane design approach?

arsenovic:
@jstenar  have you experience with noise data?  do you have suggestions about a design architecture?

TomWoo:
The page from Qucs generalizes an n-port network, whereas the 'simpler' formulation is for a 2-port, which I think is all I would ever use. Storing a separate set of freq's makes sense to me, to minimize loss of numeric precision.

cactorium:
Got some code written, please review

FournierVincent:
Noise calculation of cascaded networks is of major interest. Congratulation for the work done.

It's **deembed** conterpart would be extremely useful as well. 
For information, it is implemented in the ADS **de_embed** component.

[...]

PR #252, opened 2019-03-22
Add some features for noisy networks

cactorium:
For https://github.com/scikit-rf/scikit-rf/issues/251 , definitely need some review/advice

Right now, it adds a bunch of properties to the networks, failing if the network doesn't have any associated noise parameters:
 - minimum noise figure
 - noise figure calculator (you plug in an impedance and it spits out the noise figures across the frequency band)
 - optimum impedance/admittance
 - equivalent noise resistance
 - cascading 2-port networks where at least one of them is noisy seems to work

Probably needs tests to verify proper functionality. I cascaded an inductor and capacitor with a s2p model and they seem to have shifted the Zopt correctly, and the getters for all the noise parameters return the same values as in the s2p model, so the transformations seem to be working

Current definite issues:
 - ~~there's little to no safety wrt to making sure it's grabbing good ABCD/Z0 values from the network in its calculations~~
    - ~~any advice on how to this this? Is there any recommended interpolation helper?~~
 - connecting the networks in any nontrivial way (ie. that isn't a cascade using **) doesn't work; need to work out how to transform the noise to get it to the other port so do this correctly
 - flipping the network probably doesn't work right, because of the same problem as the last thing
 - need to add noise to resistors


skrf/network.py
```diff
@@ -1141,6 +1144,74 @@ def f(self, f):
         tmpUnit = self.frequency.unit
         self.frequency = Frequency.from_f(f, unit=tmpUnit)
 
+    @property
+    def f_noise(self):
+      """
+      the frequency vector for the noise of the network, in Hz.
+      """
+      return self.freq_noise
```
arsenovic: should this be self.noise_freq?

skrf/network.py
```diff
@@ -1141,6 +1144,74 @@ def f(self, f):
...
+      return fmin + rn/gs * npy.square(npy.absolute(ys - y_opt))
+ 
+    @property
+    def rn(self):
+      """
+      the equivalent noise resistance for the network, relative to the frequency values in self.noise_freq
+      """
+      if self.noise is None:
+        raise ValueError('network does not have noise')
+      k = 1.38064852e-23
```
arsenovic: would be good to define these constants once and import them to prevent errors,

skrf/network.py
```diff
@@ -2963,6 +3063,34 @@ def connect(ntwkA, k, ntwkB, l, num=1):
         ntwkC.renumber(from_ports=from_ports,
                        to_ports=to_ports)
 
+    # if ntwkA and ntwkB are both 2port, and either one has noise, calculate ntwkC's noise
+    if num == 1 and ntwkA.nports == 2 and ntwkB.nports == 2 and (ntwkA.noise is not None or ntwkB.noise is not None):
```
arsenovic: wow what a conditional!

arsenovic:
looks really good. very readable and  nice comments.  couple points/ideas 
  - needs tests like you mentioned. 
 -  i dont know what the best way to do the frequency is.  i imagine you always want to always keep the original data (noise,freq_noise) un-changed, and then have interpolated noise data which matches network.frequency.f.
- if you make scalar noise properties like noise_mag or noise_db, we can add them to the plotting functions and you could do. ntwk.plot_noise_db(). 
- for tests maybe you could just manually assign  the noise from a statistical distribution. i dont know much about noise.

skrf/network.py
cactorium: Yes! Fixed

skrf/network.py
cactorium: Done!

skrf/network.py
cactorium: Haha, yeah! I couldn't think of an easier to way, but it is a little shorter now

cactorium:
Thanks for the compliment, and the quick feedback!
- I will work on adding tests over the weekend
- Yeah, that's definitely the biggest issue right now I think. Should the getters return the noise parameters interpolated into the network.frequency.f space? I'm guessing cascading/flipping should be done in the noise frequency space, and interpolate the network parameters into there. Does that make sense?
- Do you have any advice on how to do that? I don't know how the plotting methods work
- Oh, the values are all deterministic, it's basically all noise power estimates instead of noise in the time domain. I'm guessing the tests I'll add will be basically making sure that it preserves the original noise parameters the noise matrix is generated from, making sure interpolation works right, and making sure cascading works right

TomWoo:
- One way to verify cascading is to use Friis' formula for noise, using available gain instead of |s21|^2. The resulting noise figure should agree.
- Here's skrf's 1d interpolation function: https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.network.Network.interpolate.html
I recommend parametrizing the interp domain and type. The defaults (cartesian and linear) can produce 'un-smooth' results. Interpolating network parameters into the noise freq domain makes sense to me.
- The plotting functions seem to be created from code generation. Adding 'noise' to the PRIMARY_PROPERTIES list under the Network class and the scalar properties mentioned above should work. Alex will know more details about that.

cactorium:
Sorry for the delay! My weekend was busier than expected

- Good idea! I'll do that on top of the tests I already had planned
- Do you have any advice on where to parameterize it? Should the getters be turned into functions so they can have interpolation parameters? What's a safe default for the cascade function, which might be called from the `**`operator where it wouldn't be possible to add some interpolation parameters?
- Will do, I'll add minimum noise figure and gamma_opt to that list

TomWoo:
Ah, I see the problem. For lack of a better way, I propose defining a class setter, say, set_noise_interp_params(domain, kind). While I tend to avoid class variables, using them here allows you to not need to modify many other sections of code. I think the defaults can still be cartesian and linear. As a side note, I found that interpolating off raw datasets works best, but clearly isn't doable if the source Touchstone files have different scales or different coordinate systems.

cactorium:
Should I try to support all of the combinations of types/kinds of interpolations? They're used in a couple different places so I'm really tempted to try to write a helper function that'll perform the right type of interpolation given some type of data input. I added a class variable to set the interpolation type that's passed to `interp1d` when an interpolation's needed, so right now it only does different interpolations within the Cartesian plane

Lemme know if you want to work on that, I'm gonna try to actually get the noise test stuff working soon. I know there's a bug in the noise test I added, I just figured out how to run the tests locally and I'll fix that in the next commit

Also, there's a really big number of warnings during the tests, is that something you guys would want fixed too?

TomWoo:
Sure, I'll work on it after your branch is released. I'd like to rely on your tests to make sure interpolation is numerically accurate.

cactorium:
Hey, I'm getting a weird error using this code:
```        b = rf.Network(f=[1, 2],
                       s=[[0, 1],
                          [1, 0]],
                       z0=50).interpolate(a.frequency)
```

It produces this error:
```----------------------------------------------------------------------
  File "/home/kelvin/repos/scikit-rf/skrf/tests/test_network.py", line 442, in test_noise
    z0=50).interpolate(a.frequency)
  File "/home/kelvin/repos/scikit-rf/skrf/network.py", line 2081, in interpolate
    interp_z0_re = f_interp(f, self.z0.real, axis=0, **kwargs)
  File "/home/kelvin/.local/lib/python3.7/site-packages/scipy/interpolate/interpolate.py", line 433, in __init__
    _Interpolator1D.__init__(self, x, y, axis=axis)
  File "/home/kelvin/.local/lib/python3.7/site-packages/scipy/interpolate/polyint.py", line 60, in __init__
    self._set_yi(yi, xi=xi, axis=axis)
  File "/home/kelvin/.local/lib/python3.7/site-packages/scipy/interpolate/polyint.py", line 125, in _set_yi
    raise ValueError("x and y arrays must be equal in length along "
ValueError: x and y arrays must be equal in length along interpolation axis.
```

Is this a bug or am I misusing the interpolate() function or Network constructor?

TomWoo:
Try this instead:
`rf.N(f=[1,2],s=[[[0,1],[1,0]], [[0,1],[1,0]]],z0=50)`
I think the constructor is misleading for not broadcasting a single S-matrix over the entire range of frequencies. Perhaps there's a good reason it doesn't do that, but if that's the case, then throwing an exception would've helped

cactorium:
Ah, I gotcha, I gotta say that that error message isn't the most obvious since the issue's with S instead of z0

I added a test that actually passes, so PTAL, I hadn't heard of any advice on whether to add an interpolation helper function or letting the interpolation mode be a class variable, so noise interpolation is still always done in Cartesian mode right now

TomWoo:
Tests look good so far. Do you have more substantial examples to add? If not, I'm going to chain multiple noisy networks together as a final litmus test. I'm fine with your use of defaults for interpolation, though adding a setter for domain wouldn't hurt. I think making the interpolation helper more dynamic requires storing the domain that the s-params were originally in, and I'm still planning to work on that after this release.

cactorium:
No, I couldn't think of any more useful examples, and I'm guessing I can implement the other features I had in mind (more operations than just cascades, noisy resistors) in later PRs. If it looks good to you I think it's ready to merge

TomWoo:
I'm getting some amount of miscorrelation with ADS noise sims for some reason. Let me double-check in the morning and get back to you.

cactorium:
So any news on the ADS simulation? I'm curious to see what might be causing the discrepancy

TomWoo:
Thanks for hanging tight. One issue is that the NF is calculated using the nfmin property instead of nfmin_db. I suggest renaming them to nfactmin and nfmin, respectively (or something along those lines) so that it's easier to distinguish between noise factor and noise figure.

Even after correcting that, there's still some residual error I want to investigate. I also noticed that nfmin (min noise factor) is computed from n:

image

Is there a reason you calculate these values instead of using ones given in the touchstone file? If you have one, can you point me to a reference for this formula?

cactorium:
I use this correlation matrix because it's the easiest form to use when you're cascading networks. I figured that keeping it in this form reduces errors from converting to noise parameters and back during each cascade

I got the noise paramters->correlation matrix equations from slide 27 here: https://www.cambridge.org/us/files/8313/6698/2373/HFIC_chapter_3_noise.pdf (although the units don't seem to make sense here so I ended up just adding a 4kT to get everything scaled correctly) and the cascade transformations from here: https://onlinelibrary.wiley.com/doi/pdf/10.1002/9781119073093.app3

TomWoo:
Turns out any remaining discrepancies are due to the gamma_opt angle not being converted to radians. After correcting that, there is excellent agreement between your implementation and ADS. With that one change (use npy.deg2rad(gamma_opt_angle) around line 1600 in network.py), I say it's ready for release. I'm planning on adding tests with larger gamma and rn to expose this kind of issue after the release.

cactorium:
Good catch! My bad for not realizing that early

skrf/network.py
```diff
@@ -2963,6 +3095,45 @@ def connect(ntwkA, k, ntwkB, l, num=1):
         ntwkC.renumber(from_ports=from_ports,
                        to_ports=to_ports)
 
+    # if ntwkA and ntwkB are both 2port, and either one has noise, calculate ntwkC's noise
+    if num == 1 and ntwkA.nports == 2 and ntwkB.nports == 2 and (ntwkA.noisy or ntwkB.noisy):
```
arsenovic: when you add a new property to the class and want to test for it,  you should except an AttributeError. otherwise any  old pickled networks will fail if they are loaded with the new code. 
i dont know how many people use pickling to store their data.

skrf/tests/test_network.py
```diff
@@ -429,6 +429,53 @@ def test_is_lossless(self):
         self.assertTrue(b.is_lossless(), 'This unmatched power divider is lossless.')
         return
 
+    def test_noise(self):
```
arsenovic: test looks great!

skrf/network.py
cactorium: Fixed I think!

skrf/tests/test_network.py
cactorium: Thanks!

cactorium:
Is there anything else you guys would like added to this PR?

arsenovic:
any chance we could get a tutorial notebook in the docs on this?

cactorium:
Hmm, give me a couple days

cactorium:
Would a pared down version of this work? https://github.com/cactorium/ism-915mhz-transceiver/blob/master/sim/low_noise.ipynb Or would something simpler and more focused on showing off the features be better?

arsenovic:
probably a little simpler would be better  with 
* more explanation and discussion,  you can use the markdown cells, and even headings to break it up if that makes sense
* you can change the ntwk name to make plots legends more readable
* the stability circles  are cool. 

overall i think this addition will be used a lot in the future

cactorium:
Sorry for the delay, I'm slowly working on the notebook! I'm thinking of removing `network.gamma_opt` in favor of a `network.g_opt` getter that generates it from the noise correlation parameters. This is because `network.gamma_opt` only exists when the network was generated from a tombstone file. Since the `network.g_opt` is generated from the correlation data, it'll exist in any case where the network has noise, although it may be slightly less accurate than `network.gamma_opt`. Does this seem like a reasonable choice?

arsenovic:
ill have to defer to someone else on the noise theory. for the code, a conditional property can return None, and this sounds sensible in your case.

TomWoo:
I think it's reasonable to store correlation matrices in that case. It might help to normalize their values by Boltzmann's constant, but probably not a big deal seeing as everyone has 64-bit machines nowadays.

cactorium:
So there's now a notebook you guys can review
- I'm guessing I need to add more discussion?
- Any advice on where to rename the networks?

The gamma_opt data is now no longer stored, and it's recalculated from the noise correlation matrix instead

arsenovic:
hi,  
 the notebook looks good.  couple things 
* the examples go in doc/source/examples/ (i should delete the old examples folder) 
* the notebooks should be stripped of output  before committed ( otherwise the repo grows every commit due to plots) 
* you text you can put in markdown cells instead of comments, but up to you. 
see more info here,  https://github.com/scikit-rf/scikit-rf/wiki/Development#examples

if its easier to seperate the functionality from the docs in seperate PR, that would work to. 
thanks for the feature !

arsenovic:
great!

cactorium:
Okay, documentation's been moved into that directory, and I stripped the output out of it and moved the text into Markdown, and I've merged it back with master. There's one weird test failure:

```
======================================================================
FAIL: The networkx package should be available to run these tests
----------------------------------------------------------------------
  File "/home/kelvin/repos/scikit-rf/skrf/tests/test_circuit.py", line 681, in test_is_networkx_available
    self.failUnless('networkx' in sys.modules)
AssertionError: False is not true
```

I actually just figured out how to get noise parameters for passive networks too, which would mean this could generate noise data for a lot more networks, including attenuators and stuff like that, so I hope it's alright if we wait a few more days so I can hash that out before merging this in

For reference I'm using info from https://www.cambridge.org/us/files/8313/6698/2373/HFIC_chapter_3_noise.pdf and a couple other papers. My current approach is to use Bosma's theorem, which says that the noise matrix for a passive network is C_z =4kT Re([Z]) and C_z = 4kT Re([Y]) . The noise correlation matrix has multiple forms, and the one that's currently being used is related to the ABCD matrix. So either C_z or C_y needs to be transformed into the ABCD matrix form, which can be done using some matrix multiplication from the formulas on page 18 of the link above. I'm not the best at numpy, so I might need some help making it more performant

cactorium:
Haha, nvm I'll do that in a new PR

FournierVincent:
The actual skrf formula for NF seems to have been super-seeded. 
The difference would rise in case of source impedance far from 50, which is the general case. 
The ADS formula is an updated implementation I think. If you don't have access to the ADS documentation I will look for public references.  

image

    def nf(self, z):
...      return fmin + rn/gs * npy.square(npy.absolute(ys - y_opt))

FournierVincent:
The code below doesn't allow me to enter in the noisy world. 
The documentation is sparse. 
do you have any suggestion ? 

    N= rf.N(f=[1,2],s=[[[0,1],[1,0]], [[0,1],[1,0]]],z0=50, z_opt= complex(0.5,0.2), rn=10,nfmin_db = 2.5, f_noise=[1,2], noisy= True  )

Bye,
Vincent

cactorium:
It looks like you're trying to set the nfmin_db parameter, but that's not how you set the noise in a network. The easiest way to do it is to use tombstone files (s2p), and you can also enter it in using `noise_freq` and `noise` parameters. If you want to enter it programmatically `noise_freq` is a list of frequencies, and `noise` is a list of noise correlation matrices in the scattering parameter form (I'm pretty sure it's in this form) (this might be the wrong wording), matching the list of frequencies in `noise_freq`

Issue #254, opened 2019-04-08
Proposal: Advanced Circuit 


jhillairet:
Connecting many Networks together can be a bit burden presently. 

I've started writing a new class named for the moment `Circuit` which is constructed by passing a list of connections. I've implemented the method described in [1], which apply to any combination of N-ports. 

For example, if one want to create the following Wilkinson divider circuit [1]:
![image](https://user-images.githubusercontent.com/4642848/55738763-13ddb780-5a28-11e9-9f24-0dfd79407b17.png)

the current implementation is:

```
connections = [
           [(port1, 0), (branch1, 0), (branch2, 0)], # intersection 1
           [(port2, 0), (branch1, 1), (resistor, 0)], # intersection 2
           [(port3, 0), (branch2, 1), (resistor, 1)] # intersection 3
        ]

C = rf.Circuit(connections)
```
where the `connections` is a list of all intersections of the circuit, describing for each which Networks and associated port are connected to the intersection.

Presently, there are few constraints:
- all Network must have a name (and distinct ones).
- The ports must be 1-port Network and the string "port" included the Network name

For example, in the example above, ports have been defined like :

```
# Ports must be defined as Network
def Port(frequency, name, Z0=50):
    '''
    Return a port Network. Passing the name is mendatory
    '''
    _media = rf.media.DefinedGammaZ0(frequency, Z0=Z0)
    return _media.match(name=name)

port1 = Port(freq, name='port1')
port2 = Port(freq, name='port2')
port3 = Port(freq, name='port3')
```

The other components are standard Networks, for example:
```
freq = rf.Frequency(start=1, stop=2, npoints=101)
Z0_ports = 50

# resistor
R = 100
line_resistor = rf.media.DefinedGammaZ0(frequency=freq, Z0=R)
resistor = line_resistor.resistor(100, name='resistor')

# branches
Z0_branches = np.sqrt(2)*Z0_ports
line_branches = rf.media.DefinedGammaZ0(frequency=freq, Z0=Z0_branches)
branch1 = line_branches.line(90, unit='deg', name='branch1')
branch2 = line_branches.line(90, unit='deg', name='branch2')
```

The class also provide a way to test if the circuit is correctly connected (all ports of all Networks should be connected) and to visualize the graph of the circuit : 

![grap](https://user-images.githubusercontent.com/4642848/55738987-79ca3f00-5a28-11e9-9c08-3e3405f6cf86.png)

[...]

PR #257, opened 2019-04-21
Circuit

jhillairet:
Connecting many Networks together can be a bit burden presently. The `Circuit` object aims to simplify the connection of many multi-ports Networks.

Creating a `Circuit` requires describing first the connection between elements, that is listing all the nodes of a circuit. An example is provided in #254.

The PR also provide many tests for various kinds of circuit. It has been tested for 2-ports nodes (ex: classical cascading) and 3-ports nodes (ex: Wilkinson splitter).


jhillairet:
Ok, it passed Python 2.

Eventually there are few remaining things to do:
 - [ ] Improve the documentation
 - [ ] Improve the plotting of the circuit graph. If possible to display the internal ports number
 - [ ] Add methods to retrieve the internal port scattering parameters. This is the big advantage of this method, it allows one to access the S-parameters of the inside ports, so voltages and currents at any connections points inside the circuit. Needs specific tests.
  
Any other ideas?

jhillairet:
Last commit improve the plotting of the circuit graph, with customizable parameters.

 Also display the port and characteristic impedances of the circuit. 

plot_example2
plot_example1

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_install:
 install:
 - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION
 - source activate test-environment
-- conda install numpy matplotlib scipy pandas nose pip ipython=5.1
+- conda install numpy matplotlib scipy pandas networkx nose pip ipython=5.1
```
arsenovic: i think you need this new requirement in many places.  alternatively, you could do an import in circuit with a try/except error message to install networkx

skrf/circuit.py
```diff
@@ -0,0 +1,586 @@
...
+
+   Circuit
+
+'''
+from . network import Network
+from . media import media
+
+import numpy as np
+import networkx as nx
+import matplotlib.pyplot as plt
```
arsenovic: @dvincentwest  moved all plotting to skrf/plotting.py. thinking back i think the idea was allow for non-plotting (headless) installs  and other backends, maybe he can comment on this

arsenovic:
this is really cool.

skrf/circuit.py
jhillairet: OK. Should I move the plot function into plotting.py under another name (`plot_graph`) ? 

It's not very clear to me how the method then appear in the Circuit object, I'll look into the code.

.travis.yml
jhillairet: OK. Only the method checking if the network is all connected (no open port) and the graph plot require `networkx`. If the latter is not present, it will raise an `ImportError` error informing this is required.

skrf/circuit.py
```diff
@@ -25,8 +25,12 @@
...
 
 import numpy as np
-import networkx as nx
-import matplotlib.pyplot as plt
+import sys
+
+try:
+    import networkx as nx
+except ImportError as e:
+    pass
```
arsenovic: you may want to raise a warning?

arsenovic: nm i see you did it down below

arsenovic: nm i see you did it down below

skrf/circuit.py
```diff
@@ -137,6 +141,7 @@ def networks_dict(self, connections=None, min_nports=1):
                 ntws.append(ntw)
         return {ntw.name: ntw for ntw in ntws  if ntw.nports >= min_nports}
 
+    @property
     def networks_list(self, connections=None, min_nports=1):
```
arsenovic: properties dont take arguments,

skrf/plotting.py
```diff
@@ -757,7 +757,8 @@ def setup_matplotlib_plotting():
     networkSet.NetworkSet.plot_uncertainty_bounds_s = plot_uncertainty_bounds_s
     networkSet.NetworkSet.plot_logsigma = plot_logsigma
     networkSet.NetworkSet.signature = signature
-
+    
+    circuit.Circuit.plot_graph = plot_circuit_graph
```
arsenovic: @dvincentwest can you take a look at this plotting architecture? i think it is as you  intended, but i cant remember the logic.

skrf/circuit.py
jhillairet: Ho indeed, you're right. It's corrected now.

jhillairet:
I've also added the Wilkinson splitter as a documentation example.

jhillairet:
Added a new notebook to illustrate how to build lumped-element circuits (like illustrated below), using both the `Circuit` and the 'classic' approach

designer_capacitor_adv


designer_bandpass_filter_450_550MHz

arsenovic:
i think we can merge this when you are happy with it,

jhillairet:
I think it is OK to merge. 

Concerning the documentation, I didn't make any tutorial notebook yet, but 2 examples. By the way, in the notebook dedicated to the Wilkinson splitter example, I can't achieve the correct result using the "classic" way (with `rf.connect()`)... The resulting network should give -3dB at 1 GHz:

```
z0_port = 50
z0_lines = line_branches.z0[0]
z0_R = line_resistor.z0[0]
# require to create the three tees
T0 = line_branches.splitter(3, z0=[z0_port, z0_lines, z0_lines])
T1 = line_branches.splitter(3, z0=[z0_lines, z0_R, z0_port])
T2 = line_branches.splitter(3, z0=[z0_lines, z0_R, z0_port])

_wilkinson = rf.connect(T0, 1, branch1, 0)
_wilkinson = rf.connect(_wilkinson, 2, branch2, 0)
_wilkinson = rf.connect(_wilkinson, 1, T1, 0)
_wilkinson = rf.connect(_wilkinson, 1, T2, 0)
_wilkinson = rf.connect(_wilkinson, 1, resistor, 0)
wilkinson = rf.innerconnect(_wilkinson, 1, 3)
```
Do you see the error if any ?

jhillairet:
I've cleaned a bit the branch from other work under the hood.  

I'm working on getting Voltage/Current at circuit nodes. It works, but I need more work to make a clean and convenient API for that. I'll make a PR later for that.

So I think it's OK to merge.

arsenovic:
sweet

jhillairet:
Great, thanks.

PR #264, opened 2019-05-10
Performance Improvements of Network s/z/y parameters conversions

jhillairet:
The conversion functions `y2s`, `z2s`, `s2y`, `s2z` used a for loop on frequencies. When dealing with many network operations, most of the calculation time was devoted into these loops. 

The proposal uses `einsum` in order to create the  necessary `(nfreqs, nports, nports)` matrices, and use the `matmul` operator (`@`) which uses broadcasting to perform operations for each frequencies. 

NB: As the `@` operator has been introduced in Python 3.5 so this PR will break the compatibility with previous Python versions, unless `matmul` used instead (at the detriment of the code readability)...

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3930,24 +3930,19 @@ def s2z(s, z0=50):
     '''
     nfreqs, nports, nports = s.shape
     z0 = fix_z0_shape(z0, nfreqs, nports)
+   
+    s = s.copy()  # to prevent the original array from being altered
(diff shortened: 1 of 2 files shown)


arsenovic:
i think keeping its worth keeping backward compatablility with matmul.

jhillairet:
OK, I've replaced the @ with `np.matmul` and tests passe under Python 2. (I've kept in comments the expression with @, much easy to understand)

PR #265, opened 2019-05-11
Added active s/y/z/vswr parameters to Network and Circuit

jhillairet:
Active parameters are useful when dealing with antenna arrays. 

For a forward wave excitation array `a` at ports, active s parameters, which are defined by:

`active(s)_m = \sum_{i=1}^N ( s_{mi} a_i ) / a_m`

are a sort of reflection coefficient when other ports are excited (while s-parameters are defined for other port matched). Definition for active y/z/vswr parameters are defined from active s-parameters.

NB: I do not understand the original `network.sa()` meaning. It is not, to my knowledge, "active s-parameters" as they do not depend on any external excitation. How and when it is used ?

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -539,4 +539,128 @@ class Circuit():
         return ntw
 
 
+    def s_active(self, a):
+        '''
+        Returns active s-parameters of the circuit's network for a defined wave excitation a.
         
+        The active s-parameter at a port is the reflection coefficients 
+        when other ports are excited. It is an important quantity for active
+        phased array antennas.
+        
+        Active s-parameters are defined by [#]_:
+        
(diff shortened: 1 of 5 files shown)


arsenovic:
excellent.  i dont remember when the `sa`  property  was added, but i agree with you  that it has a different meaning.  it looks like `sa`  just  inverts the smith chart, so the outside becomes the inside, for each individual s-parameter. 

what you have done makes more sense to me.  which i interpret as the same transformation acting on a different basis.

i think maybe we should just rename the `sa` property   to `s_invert`, otherwise we will have a redundancy.

jhillairet:
I agree. I've renamed `sa` into `s_invert` and update the documentation accordingly.

PR #266, opened 2019-05-12
Fix `plot_uncertainty_bounds_component()`

boxyrobot:
Issue one: the supplied axes object `ax` is not used, current one is taken instead. When using sub-plots, this results in all plots of given type ending in the last active plot.

Issue two: `numpy.ComplexWarning` is raised when the `ax.fill_between()` is called due to complex values of both `upper_bound` and `lower_bond`.


arsenovic:
great fix.

PR #268, opened 2019-05-14
Allow terminating a 2N-port network with a N-port network

mhuser:
There is something strange renumbering done in connect(A, k, B, l) when B is 2-port and A is N-port with N>2. 
This cause 2-port failing to cascade with 4-port balanced network.
This is commented out for testing purpose, but a better solution has to be found.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2165,7 +2165,7 @@ class Network(object):
             mag = npy.abs(x)
             interp_rad = interp1d(f, rad, axis=0, fill_value='extrapolate')
             interp_mag = interp1d(f, mag, axis=0, fill_value='extrapolate')
-            dc_sparam = interp_mag(0) * npy.exp(1j * interp_rad(0)).real
+            dc_sparam = interp_mag(0) * npy.exp(1j * interp_rad(0))
         else:
             #Make numpy array if argument was list
             dc_sparam = npy.array(dc_sparam)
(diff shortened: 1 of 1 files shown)


mhuser:
So Travis is unhappy with this change, which isn't surprising, let's understand what happens:
Cascading 2N-ports is already possible like this (port numbeed from 1 to 2N for readability)
```
A ** B
  +------+  +------+
1-|1  N+1|--|1  N+1|-N+1
2-|2  N+2|--|2  N+2|-N+2
N-|N   2N|--|N   2N|-2N
  +------+  +------+
```
Terminating a 2N-port with a N-port shall be allowed and result in a N-port network:
```
A ** B
  +------+  +------+
1-|1  N+1|--|1     |
2-|2  N+2|--|2     |
N-|N   2N|--|N     |
  +------+  +------+
```
Doing this, there is an issue with the `Network.connect` function which implement a special case for when connecting (N>2)-port network to a 2-port. For this specific condition, there is a port renumbering, let's see what happens if cascading a 4-port to a 2-port, with renumbering (former behaviour):
```
rf.connect(A, 2, B, 0, num=2)
  +----+  +----+
0-|0  2|--|0   |
1-|1  3|--|1   |
  +----+  +----+
```
This will trigger an error, because cascade will try to pop a third port from the port list that has only two elements, so cascade implementation does not works well with num being different than 1.
Should we remove this behaviour, this cause other connection problems. Why is this implemented? Let's cascade again but with a single connection:
```
rf.connect(A, 2, B, 0, num=1)
  +----+    +----+
0-|0  2|----|0  1|-2
1-|1  3|-3  |    |
  +----+    +----+
```
if the renumbering was not done, it would results in a port strange order, with remaining port of the 2-port being always at the end of the list.
```
rf.connect(A, 2, B, 0, num=1)
  +----+    +----+
0-|0  2|----|0  1|-3
1-|1  3|-2  |    |
  +----+    +----+
```
To keep backward compatibility, we may keep this renumbering only when num==1, it has no sense if num==2 and num cannot have any different value if the second network is a 2-port.

mhuser:
Travis is now satisfied once more.

arsenovic:
great  job.  more complicated that  expected..

Issue #272, opened 2019-06-14
Thoughts on allowing skrf.read_all to accept a file list


iancleary:
Is there other interest in modifying the read_all function to also accept a file list. https://github.com/scikit-rf/scikit-rf/blob/994b92b03602a2cdadf647d01d48603467128173/skrf/io/general.py#L262-L267 

There are situations in which I have a list of files that span multiple directories, and have already built a list of files from s2p or ntwk files.

It would be my desire to modify the function to allow a `files` parameter while keeping the same behavior if a dir is passed in (as per the current functionality)

## From: 
`def read_all(dir='.', contains = None, f_unit = None, obj_type=None):`

...
~~~~
out={}
for filename in os.listdir(dir):
        if contains is not None and contains not in filename:
            continue
        fullname = os.path.join(dir,filename)
        keyname = os.path.splitext(filename)[0]
~~~~

--------------

## To: 

`def read_all(dir='.', contains = None, f_unit = None, obj_type=None, files=None):`

~~~~
    out={}
    
    filelist = files
    if files == None:
        filelist = os.listdir(dir)
    
    for filename in filelist:
        if contains is not None and contains not in filename:
            continue
        if files == None:
            fullname = os.path.join(dir,filename)
            keyname = os.path.splitext(filename)[0]
        else:
            fullname = filename
            keyname = os.path.splitext(os.path.basename(filename))[0]
~~~~

--------------
The following code would be unchanged and the documentation would be updated.

Thoughts?

Thank you for your time and I would update the tests and documentation accordingly.

arsenovic:
i think this would be fine. i like that its backward compatible.   go ahead and make a PR.

iancleary:
@arsenovic thanks for the merge, would you please close this

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #274, opened 2019-06-22
Feat/iac 272 read all files parameter

iancleary:
This implements https://github.com/scikit-rf/scikit-rf/issues/272 and allows users to pass in a files parameter, which is a list of files.  

This is helpful for situations when a directory parameter and a contains string is not the desired search mechanism for networks. Files can span multiple directories. My intended addition is for cases when the list of files is determined outside of the read_all function.  

This change is backwards compatible, as agreed.

- Updated read_all function, in a backwards compatible way, per issue https://github.com/scikit-rf/scikit-rf/issues/272 agreement
- add test case with a list of networks in test folder 
  - (intentionally not testing multiple directories as this is self contained in the data folder, but a list of absolute paths is now supported)
- Updated Networks.ipynb file showing the new function

Please let me know if there is any thing more you would need before merging.

Thank you,
Ian

### DIFF
--- a/doc/source/tutorials/Networks.ipynb
+++ b/doc/source/tutorials/Networks.ipynb
@@ -683,6 +683,23 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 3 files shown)


arsenovic:
looks great

PR #278, opened 2019-07-13
Added Python 3.7 to Travis

jhillairet:


jhillairet:
seems not that simple. Will see later.

PR #269, opened 2019-05-16
pre-compute scaled frequency in write_touchstone

JAnderson419:
write_touchstone called f_scaled a number of times equal to len(network.f). This presented a large bottleneck for my Dash program, which has to write data back to touchstone to serialize it and pass it between callbacks (don't want to use pickling for a webapp with untrusted data).

The following is profiling of my code importing and writing to touchstone the same five snp files, which have 20001 frequency data points each. Before the edit, write_touchstone was 90% of the 90s compute time with almost all of that being f_scaled. After the edit, compute time is more than halved and write_touchstone is only 60% of total compute time with c2str_A and c2str_B being the main performance bottlenecks.

Before Edit:
=============

> Thu May 16 11:35:37 2019    debug\skrf_improvement\POST._dash-update-component.89097ms.1558020937.prof
>          13906347 function calls (13906326 primitive calls) in **89.096 seconds**
>    Ordered by: internal time, call count
>    List reduced from 344 to 30 due to restriction <30>
**>    ncalls  tottime  percall  cumtime  percall filename:lineno(function)**
>    **100015   71.764    0.001   71.846    0.001 .\SNaP\venv\lib\site-packages\skrf\frequency.py:415(f_scaled)
>         5    5.912    1.182   79.608   15.922 .\SNaP\venv\lib\site-packages\skrf\network.py:1451(write_touchstone)**
>         1    3.502    3.502    3.502    3.502 {method 'compress' of 'zlib.Compress' objects}
>         5    1.807    0.361    4.026    0.805 .\SNaP\venv\lib\site-packages\skrf\io\touchstone.py:106(load_file)
>    400020    1.318    0.000    1.318    0.000 .\SNaP\venv\lib\site-packages\skrf\io\touchstone.py:224(<listcomp>)
>         3    0.573    0.191    0.573    0.191 C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\json\encoder.py:204(iterencode)
>   3700275    0.408    0.000    0.408    0.000 .\SNaP\venv\lib\site-packages\skrf\network.py:763(s)
>   1600080    0.397    0.000    0.397    0.000 .\SNaP\venv\lib\site-packages\numpy\lib\type_check.py:124(real)
>        21    0.380    0.018    0.380    0.018 {method 'recv_into' of '_socket.socket' objects}
>   1600080    0.363    0.000    0.363    0.000 .\SNaP\venv\lib\site-packages\numpy\lib\type_check.py:170(imag)
[...]

PR #279, opened 2019-07-20
Add methods for reading/writing networks from/to JSON strings.

JAnderson419:
To further speed up my web-based frontend for skrf, I moved to JSON serialized representations of Network objects for caching rather than encoding/decoding touchstone objects. JSON was about 2-3x faster than the touchstone route (after pull request #269) for the test data I was using (4-port, 20001 point sweep, 5 files). 

This pull request implements the functions I wrote in the skrf.io submodule, as well as provides a shorthand method for Network objects (to_json_string). Passes unit test.

Writing this test did bring up a more fundamental question about skrf in my mind, however. Why is it that __eq__ for networks only looks at the s parameter matrix? An early version of my encoder test compared network objects, but then failed when I added a comparison of frequency objects in because I wasn't handling the funit property correctly. I would propose that __eq__ should compare both f and s for a Network. While the /objects/ may still not be equal in all their properties, this would at least ensure that the measured objects were numerically equivalent. Alternatively, if you wanted more thorough equality for the object, you could check additional parameters such as name and have a separate function for numerically comparing s parameter measurements (with some tolerance for experimental variation) -> ex) two devices represented by Ntwk1 and Ntwk2 are the same because their S parameters are within 5% (user-defined) of each other over the measured range.


helgrind:
Technically, for the measured objects to be equivalent, as you say, the measurement port impedance (characteristic impedance matrix) should also be equal!

arsenovic:
i like it. 
is there reason why this is failing for 2.7?

JAnderson419:
It looks like when Network.py runs 

`from io import TextIOWrapper`

Python 2.7 is trying to import TextIOWrapper from the skrf.io module whereas python 3.x imports the function from the standard library as intended. I have not used Python 2 at all, so I am not too familiar with the intricacies of how it handles imports compared to 3.x, but searching for an answer to the solution, most people say to either rename the module to not mask the standard library or to alter sys.path so that it doesn't go for relative imports first. Neither sound particularly attractive, so I am open to other suggestions from anyone more familiar with 2.7.

JAnderson419:
Travis CI was giving me problems with xvfb. Initially it was passing on my local repo and failing to start on the scikit-rf branch. It looks like the scikit-rf build was happening on xenial (16.04) which is the new Travis CI default and uses a different syntax for xvfb. My local repo was still building on trusty (14.04), which passed with the old syntax and failed with the new syntax. I added an explicit dist statement to the .travis.yml to ensure that it builds on xenial so that the x11 testing works.

JAnderson419:
> It looks like when Network.py runs
> 
> `from io import TextIOWrapper`
> 
> Python 2.7 is trying to import TextIOWrapper from the skrf.io module whereas python 3.x imports the function from the standard library as intended. I have not used Python 2 at all, so I am not too familiar with the intricacies of how it handles imports compared to 3.x, but searching for an answer to the solution, most people say to either rename the module to not mask the standard library or to alter sys.path so that it doesn't go for relative imports first. Neither sound particularly attractive, so I am open to other suggestions from anyone more familiar with 2.7.

fyi, this was giving me problems on 3.x as well today. It may have worked locally when I was troubleshooting #280 because I was running python from a different directory and had already imported the standard library io before I created a network. Not sure why it passed in CI if this is the case however...

arsenovic:
ci is hard.  the xvfb is left over from plotting stuff which vince fixed. i think.

PR #281, opened 2019-07-24
Additional Transmission line and unit conversions functions

jhillairet:
- Added convenience math functions to deal with US/SI units (`meter_2_feet`, `feet_2_meter` and `db_per_100feet_2_db_per_100meter`)
- Added associated tests of math functions
- Added transmission lines functions to calculate vswr (`Gamma0_2_swr` and `zl_2_swr`)
- Added transmission line function to calculate the total loss in a loaded transmission line. (Used "exact" expression, not the usual approximation for low losses lines)
- Added transmission line function to propagate voltage and current on a transmission line
- Added associated tests for voltage and current propagation

PR #282, opened 2019-07-27
Create Coaxial media from attenuation and VF values

jhillairet:
It is convenient to create a coaxial media from the attenuation and velocity factor values, as they are often given in coaxial manufacturer datasheets. This classmethod of the Coaxial class allows to do that.


jhillairet:
It currently fails as it requires some function proposed in the PR #281

jhillairet:
hum... obtain a SyntaxError: Non-ASCII character '\xc2' despite the UTF8 line at the beginning of the file...

PR #284, opened 2019-08-03
Documentation on basic transmission line analysis

jhillairet:
I've added a notebook to document the transmission line analysis with scikit-rf, with a lossy loaded line example. 

- Impedance, Gamma, SWR are evaluated using both transmission line functions and network cascading
- voltages and currents are evaluated using transmission line functions. Voltages and currents are swept along transmission line length.
- another example of deducing the propagation constant of an unknown shorted transmission line.

The examples have been tested against Ansys Circuit, Qucs, SimSmith and TLW. (NB: to my knowledge, Qucs transmission line element does not implement Velocity Factor, so it works only for VF=1, unless tricking with transmission line length)


jhillairet:
Added two other notebooks on 
- the calculation of losses in a lossy transmission line
- the evolution of the SWR along a lossy transmission line
Both notebooks were inspired from ARRL Pacificon Antenna Seminars.

PR #301, opened 2019-12-03
Hybrid parameters added

whipstein:
Added hybrid parameters to the network object.

### DIFF
--- a/network.py
+++ b/network.py
@@ -1,0 +1,5204 @@
+"""
+.. module:: skrf.network
+========================================
+network (:mod:`skrf.network`)
+========================================
+
+
+Provides a n-port network class and associated functions.
+
+Much of the functionality in this module is provided as methods and
+properties of the :class:`Network` Class.
+
+
+Network Class
+===============
+
+.. autosummary::
+    :toctree: generated/
+
+    Network
+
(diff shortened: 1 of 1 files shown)


alexpacini:
I think something related to a different line termination (`\r\n` instead of `\r`) is happening, as the diff shows that the entire file has been changed.

jhillairet:
Dear @whipstein, I remove this PR which was buggy due to the line termination. By the way, the conversion to H parameters is only in one direction so far (z2h). Will you add the other direction (h2z)?

whipstein:
I created PR307. It should fix the Network class referencing of h-parameters. The h2z is already there. I even checked the master of the main repository. It was incorporated with PR302.

[Matt Braunstein - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=adt0e)	[adt0e]

On December 12, 2019 at 7:24 GMT, Julien Hillairet <notifications@github.com> wrote:

Dear [@whipstein](https://github.com/whipstein), I remove this PR which was buggy due to the line termination. By the way, the conversion to H parameters is only in one direction so far (z2h). Will you add the other direction (h2z)?

PR #307, opened 2019-12-12
H param fixes

whipstein:
Full conversion to and from h-parameters added. Updated allowed Network class attributes to call h-parameters directly with a '.h' attribute. Added conversion tests to the Network testing file.


arsenovic:
looks good thanks.

Issue #313, opened 2019-12-24
methods (s2z, z2s, s2y, y2s, renormalize, ...) should be extended to complex impedances 


FournierVincent:
The actual behavior is wrong with complex z0 impedances and should at least give an error. 
The correct formulas are easy to implement.  

example : z2s
       sqrty0 = npy.mat(npy.sqrt(npy.diagflat(1.0 / z0[fidx])))
        s[fidx] = (sqrty0 * z[fidx] * sqrty0 - I) * (sqrty0 * z[fidx] * sqrty0 + I) ** -1

should be coded as :
        gref = npy.mat(npy.sqrt(npy.diagflat(0.25 / npy.real(z0[fidx]))))  
        zref = npy.mat((npy.diagflat( z0[fidx])))
        s[fidx] = gref *(z[fidx] -npy.conj(zref))*(z[fidx]+zref)**-1 * gref ** -1

The formulae can be found in [this reference](https://onlinelibrary.wiley.com/doi/pdf/10.1002/0471200689.app4) but the best implementation can be found in the C file from the ADS simulator
[network_fun.ael.txt](https://github.com/scikit-rf/scikit-rf/files/3997638/network_fun.ael.txt)

These methods are critical for most RF circuit activities (matching, calibration, analysis, ...). Just think of matching a network as renormalising (ie s2z(z2s)) its S parameters to simultaneously matching  impedances Zi,Zo. 

By the way, these formulae z1=sqrt(z11*DZ/z22) are missing. ...

alexpacini:
Is it because of the pseudo-waves or power-wave definition (see also 10.1109/MMM.2013.2279494)? 
I think skrf has the ability to use also the power wave formulation, for example in renormalize:
```python
    def renormalize(self, z_new, powerwave=False)
```

FournierVincent:
I read the power wave reference (the guy is very clever). 

I tried the powerwave option also but it doesn't give the proper result. 

powerwave and pseudo waves definition are different. 

I need pseudo wave with a complex reference impedance. 

Typical usecase : The S parameter simulation a saw bandpass filter is run with a complex port impedance (23+j4.2, 45+j3) to get its best performances. 




Vincent Fournier, 
RF board design, RF measurements, IOT applications, embedded SW 

06 01 84 88 97 
vf.fournier@free.fr 
www.linkedin.com/in/vincentfournier 



----- Mail original -----

De: "Alex Pacini" <notifications@github.com> 
[...]

PR #320, opened 2020-01-10
Correct s-parameters calculations for complex characteristic impedance 

jhillairet:
Issue #313 has shown that s-parameters calculations is broken for in case of complex characteristic impedances. 

While I believed that pseudo-waves s-scattering parameters formulation, this is in fact not the case. 

This PR implements both power-waves and pseudo-waves definitions. The default choice is for _power-waves_ definition, as it is the most common implementation in all computer circuit softwares (like ADS of ANSYS Circuit). Hence, results obtained by default should match these tools, even now for complex characteristic impedances.

It is still possible to change the s-parameter definition to pseudo-waves definition using a keyword parameter `s_definition` in `Network` class or in `s2z/s2y/y2s/z2s` functions. 

A side-effect of the changes is that the numerical precision had changed a bit, which led me to decrease the absolute precision in order for two tests related to calibration to pass.

List of things to do before this PR could be merged:

 - [x] Python 2 compatibility (no `@` operator)
 - [x] Add more test cases for complex characteristic impedances
 - [x] Fix/improve documentation (esp. equations)
 - [x] Make a clear notebook documenting clearly the differences between power and pseudo-waves implementation
 - [x] check with `Circuit` results 

Anyway, I'm not really convinced of the need for pseudo-waves implementations (any user cases?). Moreover, I don't know any software/code from which we could test this implementation. So, the present implementation is not tested (yet?).


jhillairet:
To test the PR locally: 
```
git checkout -b jhillairet-complex_char_impedance master
git pull https://github.com/jhillairet/scikit-rf.git complex_char_impedance
```

arsenovic:
thanks for this great effort julien . i think you are right in that it is most important for us to agree with other simulators, and to keep backwards compatibility when possible.

skrf/calibration/tests/test_calibration.py
```diff
@@ -715,16 +716,17 @@ def test_z0(self):
...
     def test_shift2(self):
         feed = self.rlgc.line(50,'um')
         dut = self.wg.random(n_ports=2)
         #Thrus convert the port impedances to 50 ohm
         dut_feed = self.wg.thru()**feed**dut**feed**self.wg.thru()
         dut_meas = self.measure(dut_feed)
-        self.assertTrue(self.cal.apply_cal(dut_meas) == dut)
-
+        npy.testing.assert_allclose(self.cal.apply_cal(dut_meas).s, 
+                                    dut.s, atol=1.1e-3)
```
arsenovic: do we know what the tol was before?  1e-3  is large. , is that the limit of current calcs, or am i mis-interpreting this?

skrf/network.py
```diff
@@ -336,6 +337,9 @@ def __init__(self, file=None, name=None, comments=None, f_unit=None, **kwargs):
             its a str
         comments : str
             Comments associated with the Network
+        s_definition : str
```
arsenovic: `s_def` would be shorter. but i leave it up to you.

skrf/network.py
```diff
@@ -374,6 +378,7 @@ def __init__(self, file=None, name=None, comments=None, f_unit=None, **kwargs):
 
         self.name = name
         self.comments = comments
+        self.s_definition = s_definition
```
arsenovic: we might need this in Network.copy() as well.

arsenovic:
this existing  [example](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Measuring%20a%20Mutiport%20Device%20with%20a%202-Port%20Network%20Analyzer.html)  would be a good additional test to ensure nothing has changed about renormalization.

skrf/network.py
jhillairet: OK, I agree

skrf/network.py
jhillairet: OK

skrf/calibration/tests/test_calibration.py
jhillairet: yes that was not normal. I found that the calibration methods also had normalization functions, which were interfering... Once corrected, the precision is the same as before (<1e-6 (ZERO) when equalling two network).

jhillairet:
> this existing [example](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Measuring%20a%20Mutiport%20Device%20with%20a%202-Port%20Network%20Analyzer.html) would be a good additional test to ensure nothing has changed about renormalization.

The error seems even lower in most cases except for Gamma = 0 where it's 10 to 100x higher (?)

skrf/calibration/tests/test_calibration.py
jhillairet: I even put the absolute tolerance at 1e-11 and it passes.

jhillairet: hum, it's more complicated. There are difference between previous impementation and the new pseudo-wave implementation, which is the one that should be used for calibrations. In particular, the `NISTMultilineTRLTest2` test case shows differences. 

Previous implementation of the `s2z` is 
```
npy.einsum('ijj->ij', sqrtz0)[...] = npy.sqrt(z0)
z =  sqrtz0 @ npy.linalg.inv(Id - s) @ (Id + s) @ sqrtz0
```
while the new one for pseudo-waves is:
```
npy.einsum('ijj->ij', U)[...] = npy.sqrt(z0.real)/npy.abs(z0)
npy.einsum('ijj->ij', ZR)[...] = z0
USU = npy.linalg.inv(U) @ s @ U
z = npy.linalg.inv(Id - USU) @ (Id + USU) @ ZR
```
In theory, both equations are the same when the characteristic impedance are real. But numerically, not. Playing with the precision of the numpy array helps in increasing the precision, but is not sufficient.

arsenovic: maybe you could just test for imaginary part in z0 ,  and then do the numerically more stable thing if its less than tol.

jhillairet: In fact, I was not correct. The two tests in `NISTMultilineTRLTest2` are "failing" (lower precision than before) _because_ they use a complex characteristic impedance. 

So, either I allow a lower precision (1E-4) for this test, or I keep the old implementation as a 3rd choice of `s_def`...

arsenovic: so what is the difference in precision for the two implementations?  1e-4 vs 1e?

jhillairet: vs 1e-6 (the default precision when doing `ntw1 == ntw2`).

jhillairet:
> > Haaa, long debate. UK or US ? ;)

Maybe to avoid any problem of spelling, the word 'generalized' would be a better choice than 'traveling' or 'travelling'. Any advice?

alexpacini:
>Maybe to avoid any problem of spelling, the word 'generalized' would be a better choice than 'traveling' or 'travelling'. Any advice?

In order of my preference:
- travelling/traveling (as it refers to the "wave" rather than to the s-parameters, and is the same as in your reference)
- generalised/generalized (again, dilemma...)

I personally prefer BE, but in the end just use what you prefer and it will be alright.
If you want to avoid any worries you can check both spellings...

arsenovic:
i would recomend the use american english, as the rest of the package is in american english. i would also avoid redundant properties. i regret every time i did this in the past. (like the tline module)

skrf/calibration/tests/test_calibration.py
jhillairet: At the end, I kept using the previous implementation for calibration in `EightTerm.renormalize`, mostly for compatibility reason.

jhillairet:
Presently, the circuit building algorithm in `Circuit` does not give "expected" power-waves results (it however gives same results as the previous s-param implementation). So to avoid confusion, I've modified the documentation to precise that z0 is expected to be real (yet).

I've defined the s-param definition names and the default used definition into `constants`. The various `s_def=` are now referring to these constants.

> i would also avoid redundant properties. i regret every time i did this in the past. (like the tline module)

What do you mean exactly?

arsenovic:
hi this sounds good to me.  i mean in refence to the tline module i have things like `reflection_coeficient` and `gamma_in`.   i have done this in other places too when i like to have more than way to name things, but it is bad. i regret it.  so  with the american/british english. just pick one.

jhillairet:
> hi this sounds good to me. i mean in refence to the tline module i have things like `reflection_coeficient` and `gamma_in`. i have done this in other places too when i like to have more than way to name things, but it is bad. i regret it. so with the american/british english. just pick one.

OK. So I'll merge once you will have validated the review.

doc/source/examples/networktheory/Renormalizing S-parameters.ipynb
```diff
@@ -122,61 +111,62 @@
...
+    "match_at_50.renormalize(50j)  # same as renormalize(50j, s_def='power')\n",
+    "match_at_50.plot_s_smith(**kw)  # expect -1j"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Complex Impedances "
+    "This is because the default behaviour of scikit-rf is to use _power-waves_ scattering parameter definition (since it is the most popular one is CAD softwares). But the _power-waves_ definition is [known to fail in such a case](https://www.nist.gov/system/files/documents/2017/05/09/MicrowaveCircuitTheory-proof.pdf). This is why scikit-rf also implement the _pseudo-waves_ scattering parameters definition, but you have to specify it using the `s_def` parameter:  "
```
arsenovic: great explanation and ref link.

skrf/network.py
```diff
@@ -174,6 +175,7 @@
 from .constants import ZERO
```
arsenovic: we could probably put these on one line, but doesnt matter too much either way.

skrf/network.py
```diff
@@ -316,7 +318,7 @@ class Network(object):
     noise_interp_kind = 'linear'
 
     # CONSTRUCTOR
-    def __init__(self, file=None, name=None, comments=None, f_unit=None, **kwargs):
+    def __init__(self, file=None, name=None, comments=None, f_unit=None, s_def=S_DEF_DEFAULT, **kwargs):
```
arsenovic: this may make all old `pickled` networks   fail to load now. but that is ok,

skrf/network.py
```diff
@@ -336,6 +338,10 @@ def __init__(self, file=None, name=None, comments=None, f_unit=None, **kwargs):
             its a str
         comments : str
             Comments associated with the Network
+        s_def : str
```
arsenovic: when the  arg  can only be a  value  out of a list, usually you list all options (when possible)
`s_def :str` -> `s_def : ['power','pseudo']`

skrf/network.py
```diff
@@ -379,6 +385,11 @@ def __init__(self, file=None, name=None, comments=None, f_unit=None, **kwargs):
         self.noise = None
         self.noise_freq = None
 
+        if s_def not in S_DEFINITIONS:
+            raise ValueError('s_def parameter should be either:', s_def)
```
arsenovic: this seems to not make sense.  shouldnt   you print `S_DEFINITIONS?`

skrf/network.py
```diff
@@ -2486,23 +2519,17 @@ def renormalize(self, z_new, powerwave=False):
         z_new : complex array of shape FxN, F, N or a  scalar
             new port impedances
 
-        powerwave : bool
-            if true this calls :func:`renormalize_s_pw`, which assumes
-            a powerwave formulation. Otherwise it calls
-            :func:`renormalize_s` which implements the default pseudowave
-            formulation. If z_new or self.z0 is complex, then these
-            produce different results.
+        s_def : str
```
arsenovic: same as above

skrf/network.py
jhillairet: Yes indeed.

skrf/network.py
jhillairet: OK

doc/source/examples/networktheory/Renormalizing S-parameters.ipynb
jhillairet: Thanks. Also learned a lot, for example that few textbooks are very bad at explaining this, or even mix up in the definitions... Even commercial software helps are not very clear about what is really implemented...

jhillairet:
Let's see how it goes for complex characteristic impedance users!

PR #324, opened 2020-01-21
is_from_hfss(): fix TypeError when comments is None

sxntxn:
This fixes:
[...]/skrf/io/touchstone.py in is_from_hfss(self)
    417         '''
    418         status = False
--> 419         if 'exported from hfss' in str.lower(self.comments):
    420             status = True
    421         return status

TypeError: descriptor 'lower' requires a 'str' object but received a 'NoneType'

... when comments is None. Return False instead.

### DIFF
--- a/skrf/io/touchstone.py
+++ b/skrf/io/touchstone.py
@@ -414,12 +414,15 @@ class Touchstone:
         status : boolean
             True if the Touchstone file has been produced by HFSS
             False otherwise
-        '''    
-        status = False
+        '''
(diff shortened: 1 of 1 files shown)


jhillairet:
Indeed. Did you encounter such a situation?

sxntxn:
Yes. Some of our equipment writes Touchstone files without any comments whatsoever.

jhillairet:
OK, thank you

PR #326, opened 2020-01-29
Fix bug in write_touchstone (Issue Number 325)

mike0164:


jhillairet:
Thanks for noticing and correcting.

PR #344, opened 2020-03-31
added noise deembeding feature

FournierVincent:
- added N.deembed property, set by the inv method

	- added code for deembedding noise correlation matrix
	caution : the noise of access network is also deembedded,
	  possibly causing matrix invertion fails in case of highly non-reciprocal element (isolator)

	- moved code from read_touchstone() into function s_noise_a()
	- corrected a bug causing z0(freq)  to seem constant

	- added nfdb_gs() method , handy for generating noise_contour plots

	TODO single point noise_frequency operation
	TODO : use s_noise_a() for noise parameter setters

added preliminary test
	retrieve the original network after deembed+cascade operations

	noise contour plots

	TODO : comparison with ADS on non-trivial access network
	TODO : noisy access network


skrf/tests/test_network.py
```diff
@@ -597,6 +598,52 @@ def test_noise(self):
...
+
+
+        self.assertEqual(retrieve_thru, thru)
+        self.assertTrue(ntwk4_thru.noisy)
+        self.assertTrue(retrieve_thru.noisy)
+        self.assertTrue(abs(thru.nfmin[0] - retrieve_thru.nfmin[0]) < 1.e-6, 'nf not retrieved by noise deembed')
+        self.assertTrue(abs(thru.rn[0]    - retrieve_thru.rn[0])    < 1.e-6, 'rn not retrieved by noise deembed')
+        self.assertTrue(abs(thru.z_opt[0] - retrieve_thru.z_opt[0]) < 1.e-6, 'noise figure does not match original spec')
+
+        ntwk4_n_thru = ntwk4_n ** thru
```
TomWoo, line 623: When I de-embed on the load side using the following code to replace these lines, the tests below fail. Is right-side 'inverse connect' not supported?
```
ntwk4_n_thru = thru ** ntwk4_n
retrieve_n_thru =  ntwk4_n_thru ** ntwk4_n.inv
```

skrf/network.py
```diff
@@ -1642,7 +1682,33 @@ def copy_subset(self, key):
...
+          nf_min = npy.power(10., nfmin_db/10.)
+          # TODO maybe interpolate z0 as above
+          y_opt = 1./(self.z0[0, 0] * (1. + gamma_opt)/(1. - gamma_opt))
+          noise = 4.*K_BOLTZMANN*T0*npy.array(
+                [[rn, (nf_min-1.)/2. - rn*npy.conj(y_opt)],
+                [(nf_min-1.)/2. - rn*y_opt, npy.square(npy.absolute(y_opt)) * rn]]
+              )
+          self.noise = noise.swapaxes(0, 2).swapaxes(1, 2)
+          self.noise_freq = noise_freq
+          xx=1
```
TomWoo, line 1707: Delete me

TomWoo:
@FournierVincent I like most of the changes you made. The unfortunate fact is that the 'inverse' noise matrix depends on other blocks in the cascade, so we'll have to live with using a deembed flag to signal connect().

skrf/network.py
FournierVincent: Hi Tom,
Thank you for your comments and corrections. 
I corrected the extra line. 
The noise parameters are traditionally connected to port 1. 
and the noise calculation is performed only in one direction in the "cascade" operation. So does the .inv operation. 
ADS is a bit more clever with that respect , but there are still constraints. I would leave it as it is and put a note in the doc.

Gloamglozer:
> @FournierVincent I like most of the changes you made. The unfortunate fact is that the 'inverse' noise matrix depends on other blocks in the cascade, so we'll have to live with using a deembed flag to signal connect(). Once you address those two comments, I think it's ready to be merged.

Hi guys, I've implemented a version of this at my workplace and I'd just like to let you know that there does indeed exist an inverse of the ABCD form of the NCM which **does not** depend on other blocks in the cascade. 

I whipped up a quick proof for ya. 
[NCM_Inverse_Proof.pdf](https://github.com/scikit-rf/scikit-rf/files/4534605/NCM_Inverse_Proof.pdf)

I have a good amount of experience with both Python and working with NCM's so I would love to help you guys out.  It would be a lot cleaner if we can keep the methods consistent, @TomWoo how can we do this without passing a flag to the connect method?

TomWoo:
> > @FournierVincent I like most of the changes you made. The unfortunate fact is that the 'inverse' noise matrix depends on other blocks in the cascade, so we'll have to live with using a deembed flag to signal connect(). Once you address those two comments, I think it's ready to be merged.
> 
> Hi guys, I've implemented a version of this at my workplace and I'd just like to let you know that there does indeed exist an inverse of the ABCD form of the NCM which **does not** depend on other blocks in the cascade.
> 
> I whipped up a quick proof for ya.
> [NCM_Inverse_Proof.pdf](https://github.com/scikit-rf/scikit-rf/files/4534605/NCM_Inverse_Proof.pdf)
> 
> I have a good amount of experience with both Python and working with NCM's so I would love to help you guys out. It would be a lot cleaner if we can keep the methods consistent, @TomWoo how can we do this without passing a flag to the connect method?

@Gloamglozer Thanks for the proof. I had been searching for a multiplicative inverse, but I realize that the additive inverse could still work like this:
1. Have inv() return A.Ca.A^H.
2. Re-implement the noise portion of connect() to simply add correlation matrices, including a check on the reciprocity of the left ABCD matrix.

TomWoo:
> > I have a good amount of experience with both Python and working with NCM's so I would love to help you guys out. It would be a lot cleaner if we can keep the methods consistent, @TomWoo how can we do this without passing a flag to the connect method?
> 
> This is great, thank you. I suggest you and @FournierVincent and @TomWoo to discuss on the scikit-rf slack Channel: http://scikit-rf.slack.com/ in order to find a way to work together

@jhillairet Do you need to invite us on slack? I'm gated at a login screen.

jhillairet:
should not. Try to use [this link.](https://join.slack.com/t/scikit-rf/shared_invite/zt-d82b62wg-0bdSJjZVhHBKf6687V80Jg)

FournierVincent:
> [NCM_Inverse_Proof.pdf](https://github.com/scikit-rf/scikit-rf/files/4534605/NCM_Inverse_Proof.pdf)

The demonstration is cute and the simplification is of great interest. Bravo !

PR #346, opened 2020-04-01
TS v1 2port data order

sxntxn:
This fixes the swapping of S21 and S12 for 2port Touchstone V1 files:

- update test case
- update get_sparameter_data() to exchange values when needed


sxntxn:
Thank you!

Issue #360, opened 2020-06-30
skrf.media.cpw.CPW ---->Z0 seems to be not correct. 


RF-NanoDetec:
`skrf.media.cpw.CPW` ---->`Z0 `seems to be not correct. 
For :
```
freq = rf.Frequency(start=10, stop=13, npoints=5000, unit='GHz')
cpw_line=CPW(freq, w=200e-6, s=21e-6, ep_r=3.75, t= 1e-6, z0=50, rho=1.68e-8)
```

The `K_ratio `is correct but the characteristic impedance is somehow wrong. 
`cpw_line.K_ratio = 1.189`---- OK
`cpw_line.Z0=72.759`----to high---- should be ~50
I double-checked it in the Qucs Studio software.

jhillairet:
Indeed, I found a bug in the implementation. The `K_ratio`is not correct either, to me it should be around 0.84 (it should be between 0 and 1).

Once, corrected, it gives Z0=51.40, which agrees with the following online calculator:
https://www.microwaves101.com/calculators/864-coplanar-waveguide-calculator
https://chemandy.com/calculators/coplanar-waveguide-with-ground-calculator.htm

![image](https://user-images.githubusercontent.com/4642848/86247894-6bc48900-bbad-11ea-8e16-e35a5de8f9b5.png)

but is slightly different than in Qucs (49.4 Ohm), so I guess there are little differences in the Qucs model:
 
![image](https://user-images.githubusercontent.com/4642848/86248862-c3afbf80-bbae-11ea-83ab-37d30be4a639.png)

jhillairet:
@RF-NanoDetec, I've fixed the problem. Please try the latest github version and let me know if it solves the bug.

RF-NanoDetec:
thank you. It looks much closer to the Qucs solutions. The Qucs solutions have some frequency dependence maybe you can implement it as well. At lower impedances around 30, the difference between Scikit_rf solution and the Qucs solution increases.
Here is an example:    
Qucs:
![grafik](https://user-images.githubusercontent.com/62340429/86473701-b414b000-bd41-11ea-8d22-8789dbcec1c9.png)
and for Scikit-rf:
![grafik](https://user-images.githubusercontent.com/62340429/86473960-28e7ea00-bd42-11ea-99e4-96063d55a69e.png)
so the difference is 6 ohms.

jhillairet:
Solved with #361

PR #361, opened 2020-07-03
CPW Media fix and tests

jhillairet:
Fixed the bug reported by @RF-NanoDetec in #360. 

The impedance should now be correct, taking into account that the current coplanar waveguide media has an infinite substrate thickness.

I've added some tests for the CPW Media. so far I didn't succeed in comparing the associated Qucs project to the results, but it's certainly my fault as I do not master Qucs.


jhillairet:
but why the Travis test result status disappeared??

PR #365, opened 2020-07-11
Update general.py

xafra:
add sorted in the read_all.
file input was nonsensical.
Probably happened due to transfer of files from Windows to Kubuntu.


jhillairet:
Thank you.

PR #376, opened 2020-07-26
Circular waveguide media

scimax:
The main goal of this PR is the circular waveguide media. There are two important parts still missing compared to the rectangular waveguide:
- [x] attenuation alpha has to be implemented for TE11 mode

While I guess I can take care of the second one, the first one is not so easy to figure out in my opinion. Maybe it should be dropped for this medium, shouldn't it?

Apart from that I fixed a few minor things in the rectangular waveguide. Correct me if I'm wrong but I'm convinced that the `from_Z0` method handles only the fundamental mode at the moment. When I derived the expression already present, I had to impose that the cutoff frequency is `pi/a`. For higher order modes this would be e.g. `m *pi/a`, or even had to include the case ´n>0`. So for the moment I thought it would be better to simply mention it in the docs.

I would leave the surface roughness for a future improvement. `from_z0` may need additional discussion.


skrf/media/rectangularWaveguide.py
```diff
@@ -121,9 +121,10 @@ def __repr__(self):
 
     
     @classmethod
-    def from_Z0(cls,frequency, Z0,f, m=1,n=0,ep_r=1, mu_r=1, **kw):
+    def from_Z0(cls,frequency, Z0,f, ep_r=1, mu_r=1, **kw):
```
jhillairet, line 124: I know it was `from_Z0` in the existing code, but it's time to consider renaming it to `from_z0` to keep consistency with the convention adopted in scikit-rf.

jhillairet, line 124: why removing the mode index (m, n) choice ?

skrf/media/rectangularWaveguide.py
```diff
@@ -248,11 +249,12 @@ def f_cutoff(self):
 
             max ( \frac{m \cdot v}{2a} , \frac{n \cdot v}{2b})
 
-        where v= sqrt(ep*mu)
+        where v= 1/sqrt(ep*mu)
```
jhillairet, line 252: Indeed, good catch. It means this was not tested correctly before

skrf/media/rectangularWaveguide.py
```diff
@@ -317,13 +319,12 @@ def lambda_cutoff(self):
...
+            v/f
 
-            f_c * v
-
-         where v= sqrt(ep*mu)
+         where v= 1/sqrt(ep*mu)
         '''
         v = 1/sqrt(self.ep*self.mu)
-        return self.f_cutoff*v
+        return v/self.f_cutoff
```
jhillairet, line 327: Indeed

skrf/media/rectangularWaveguide.py
scimax: I will do it with the next commit

skrf/media/circularWaveguide.py
```diff
@@ -0,0 +1,433 @@
+'''
+.. module:: skrf.media.rectangularWaveguide
+================================================================
+rectangularWaveguide (:mod:`skrf.media.rectangularWaveguide`)
+================================================================
+
+Represents a single mode of a homogeneously filled circular
+waveguide of cross-section `r^2 pi`. The mode is determined by
+`mode-type` (`'te'` or `'tm'`) and mode indices ( `m` and `n` ).
+
```
jhillairet, line 10: Unless you plan to take them into account, corrugated circular waveguides can support HE modes as well. So maybe you should indicate it only concerns circular waveguides (not corrugated).

skrf/media/circularWaveguide.py
```diff
@@ -0,0 +1,433 @@
+'''
+.. module:: skrf.media.rectangularWaveguide
```
jhillairet, line 2: pay attention to change the name

skrf/media/circularWaveguide.py
```diff
@@ -0,0 +1,433 @@
...
+
+        # if self.rho is None:
+        #     return 0
+        # a,b,w,ep,rho,f_n = self.a, self.b, self.frequency.w, self.ep, \
+        #     self.rho, self.f_norm
+        # return 1./b * sqrt( (w*ep)/(2./rho) ) * (1+2.*b/a*(1/f_n)**2)/\
+        #     sqrt(1-(1/f_n)**2)
+
+    @property
+    def Z0(self):
```
jhillairet, line 424: maybe it should be `z0` instead of `Z0` to be consistent with skrf conventions. tbd.

skrf/media/rectangularWaveguide.py
scimax: That's what I mentioned in the main comment. I might be wrong, but when I derived the formula given here, the cutoff frequency is used and here, only $omega_c = c * \pi/a$ is considered, although the cutoff frequency for a non fundamental mode with different `m` and `n` than (1,0) would be $omega_c = c * \sqrt( (m \pi/a)^2 + (n \pi/a)^2)  $. Or am I wrong in the derivation?

scimax: Sorry for the latex-like math

jhillairet: You're right, I've updated my comment after reading more carefully your initial PR message. 

The formulas are (pay attention to the b-dependance of the n index)
image

scimax: True! The missing b-dependence was just a typo. I have to admit that I didn't directly generalized the given calculation to any mode since the formula was a little cumbersome. But since I got it now it should be fine to implement it.
One question regarding the original formula in `rectangularWaveguide`: Is there a specific reason to have the `j` in the square of the impedance? To me it just adds complexity as it is mathematically equal to a minus sign.
So originally:
```py
1./sqrt((1/(z0*1j)**2+ep/mu))
```
could be replaced with
```py
1./sqrt((ep/mu  - 1/z0**2))
```

jhillairet: Yes indeed. (Unless side effects with inverse of negative imaginary part of z0.)

However, upon reflection, initializing from z0 only works for TE10 (or TEm0, or TMm0), but not for TEmn or TMmn. In the latter case, both `a` and `b` are unknown and can't be retrieved from z0 only.

jhillairet:
Hi @scimax. Did you continue to work on that PR?

scimax:
Hi @jhillairet , I'm sorry, I got a little bit distracted and forgot it. Let me take a look over the weekend. I'm still quite busy so I'm not sure how far I can make it.

jhillairet:
> 
> 
> Hi @jhillairet , I'm sorry, I got a little bit distracted and forgot it. Let me take a look over the weekend. I'm still quite busy so I'm not sure how far I can make it.

No problem. You've made a lot of work on this, so it would be too bad to not finish it. Let me know if you need some help to finish it.

scimax:
> > Hi @jhillairet , I'm sorry, I got a little bit distracted and forgot it. Let me take a look over the weekend. I'm still quite busy so I'm not sure how far I can make it.
> 
> No problem. You've made a lot of work on this, so it would be too bad to not finish it. Let me know if you need some help to finish it.

Well, actually, yes, I have a more theoretical question^^ the formula for attenuation in a rectangular waveguide, L. 421, is taken from Orfanidi. But in the reference Orfanidi states it's for TE01 mode. Is it also valid for other modes, e.g. any TM mode? My first guess would be no for TMmn, but yes for TEmn. But I would have to go through the derivation.

I'm asking since I implemented it for the circular waveguide but in the reference I used, it's only derived for the fundamental TE11 mode. So for generalization to any mode I will have to go through the derivation with the bessel functions. I guess there has to be a general solution somewhere, isn't it?

scimax:
While going through the `from_z0` code I found a bug in the current version of the rectangular waveguide. In [line 143](https://github.com/scikit-rf/scikit-rf/blob/367a7f5d024c9e016b0ce226a0af1da4cf454316/skrf/media/rectangularWaveguide.py#L143) the waveguide size `a` is computed. But the parameters `m` and `n` are ignored. You are absolutely right that it won't work for non-zero `n` as the system of equations does not have a unique solution. But if you are able to provide `m` then `a` should be:
```py
a =m*pi/(w*mu) * 1./sqrt((1/(Z0*1j)**2+ep/mu))
```
Or is it supposed to work only for the fundamental TE10 mode? Then, the parameters n and m in the method are confusing.

I will first merge the current master into my branch and then update it with the missing `m`.

scimax:
@jhillairet I would leave the surface roughness for a future improvement. `from_z0` may need additional discussion. I would revise my initial thought of extending it to higher order modes (e.g. TEm0 as you mentioned). It may be confusing. What do you think? Or should we implement it with more points of `(f, z0)`? Then one could extend it to any TEmn or TMmn mode.

Also the attenuation is only implemented for the fundamental TE11. But this is just because of confusion, see my comment above.  
I also dropped the roughness parameter. I would like to check if the HFSS formula used in `rectangularWaveguide` is general or only specific for this geometry. But I don't have access to HFSS. Have you found a resource for that?

So I would say we can merge it as it is at least.

jhillairet:
ok, thanks for your work. There is an error remaining with Rectangular waveguide. Can you correct that before we merge.

> I also dropped the roughness parameter. I would like to check if the HFSS formula used in rectangularWaveguide is general or only specific for this geometry. But I don't have access to HFSS. Have you found a resource for that?

ok, eventually I can do that.

scimax:
> ok, thanks for your work. There is an error remaining with Rectangular waveguide. Can you correct that before we merge.

Ah, now I realized, where it comes from! I think we have to revert [this change](https://github.com/scikit-rf/scikit-rf/pull/376#discussion_r460715939), meaning `Z0` instead of `z0`, and following the naming the function should also be called `from_Z0` instead of `from_z0`. With this the port impedance `z0` and the characteristic impedance `Z0` can be distinguished. But I have to say, `Z0` wouldn't be pythonic. So maybe it's also worth to think of a different name for the characteristic impedance, e.g. `z_c` or similar.  

The error originates from the abstract parent `media` class, as it sets the port impedance `z0`. But the child class defines the property `z0` without setter. So `self.z0 = z0` in the  parent class is invalid.

skrf/media/circularWaveguide.py
```diff
@@ -0,0 +1,405 @@
...
+    #     .. math::
+
+    #             k_x = m \\frac{\pi}{a}
+
+    #     Returns
+    #     -------
+    #     kx : number
+    #             eigen-value in `a` direction
+    #     '''
+    #     return self.m*pi/self.a
```
jhillairet, line 212: if not used, please remove it from the PR

jhillairet:
> Ah, now I realized, where it comes from! I think we have to revert [this change](https://github.com/scikit-rf/scikit-rf/pull/376#discussion_r460715939), meaning `Z0` instead of `z0`, and following the naming the function should also be called `from_Z0` instead of `from_z0`. With this the port impedance `z0` and the characteristic impedance `Z0` can be distinguished. But I have to say, `Z0` wouldn't be pythonic. So maybe it's also worth to think of a different name for the characteristic impedance, e.g. `z_c` or similar.

OK. Keep `Z0` for the moment, since it's define like that in all the code base.

scimax:
I'm not sure what the original motivation for the weird formula of the rectangular waveguide's cutoff frequency was, But I've updated it with the cut-off frequency you also presented in the literature page above, @jhillairet .

scimax:
@jhillairet What do you think? Is it ready?

jhillairet:
Sorry, I was away from computers. I think it's OK.

jhillairet:
@arsenovic , do you have any comments?

skrf/media/circularWaveguide.py
```diff
@@ -113,26 +113,27 @@ def __repr__(self):
...
         
         mu = mu_0*mu_r
         ep = epsilon_0*ep_r
         w = 2*pi*f
         # if self.mode_type =="te":
         u = jnp_zeros(1, 1)[-1] 
-        r =u/(w*mu) * 1./sqrt((1/(z0*1j)**2+ep/mu))
+        r =u/(w*mu) * 1./sqrt((1/(Z0*1j)**2+ep/mu))
         
         kw.update(dict(frequency=frequency, r=r, m=1, n=1, ep_r=ep_r, mu_r=mu_r))
```
arsenovic, line 138: m =0? right ?

skrf/media/rectangularWaveguide.py
```diff
@@ -121,28 +121,27 @@ def __repr__(self):
 
     
     @classmethod
-    def from_Z0(cls,frequency, Z0,f, m=1,n=0,ep_r=1, mu_r=1, **kw):
```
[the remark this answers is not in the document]
scimax: Well, `m=1, n=0` is hard-coded on line 144. I think we discussed it a little bit, and it mainly makes sense for the fundamental mode. Otherwise you would have to deduce two parameters from one given parameter which leaves some ambiguity. So in my opinion having it as a parameter in the function is not so useful. That's also the reason why I updated the docstring documenting the function. But I'm still open for discussion^^

skrf/media/rectangularWaveguide.py
```diff
@@ -47,7 +47,7 @@ class RectangularWaveguide(Media):
     z0 : number, array-like, or None
         the port impedance for media. Only needed if  its different
         from the characterisitc impedance of the transmission
-        line. if z0 is None then will default to Z0
```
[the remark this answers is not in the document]
scimax: True, I think this is a leftover from the confusion @jhillairet and I had about `z0` vs. `Z0`, as solved in commit https://github.com/scikit-rf/scikit-rf/commit/e6d8cd22abc33919a0afa84d07bbd2c14c3712af

I will undo this in a second.

jhillairet:
Thank you @scimax for your work.

PR #377, opened 2020-07-26
fixed error: plotted different mode gain instead of common mode gain

jrmagers:
The last plot should have been common mode gain, not differential mode gain. Also, fixed incorrect turns ratio on transformer.  The results are now consistent with the results from Microwave Office.


jhillairet:
great, thank you.

jhillairet:
Documentation has been updated.

Issue #368, opened 2020-07-15
Check for equality rather than identity (SyntaxWarning on Python 3.8)


JAnderson419:
Running skrf on python 3.8 will result in many warnings similar to the following:

```
SyntaxWarning: "is" with a literal. Did you mean "=="?
 if smithR is 1:
```

From the python 3.8 release notes:

> The compiler now produces a SyntaxWarning when identity checks (is and is not) are used with certain types of literals (e.g. strings, numbers). These can often work by accident in CPython, but are not guaranteed by the language spec. The warning advises users to use equality tests (== and !=) instead. (Contributed by Serhiy Storchaka in bpo-34850.)

To be in compliance with the language spec and prevent edge cases from causing problems, we should probably move all the is/is not statements over to == and !=

jhillairet:
I've changed the `is` to `==` which generated Warnings `plotting.py` (in #375). There are probably some other in the code, but the tests did not made them appeared.

PR #378, opened 2020-07-27
Fix Python 3.8 SyntaxWarnings

sxntxn:
This fixes 

```
[...]/python3.8/site-packages/skrf/network.py:3039: SyntaxWarning: "is not" with a literal. Did you mean "!="?  
  if Pa.shape[0] is not 0:  
[...]/python3.8/site-packages/skrf/network.py:3385: SyntaxWarning: "is" with a literal. Did you mean "=="?  
  if common_freq[0].size is 0:  
```

(Line numbers are a little off, I'm using the 0.15 release branch.)

See https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior , 2nd bullet.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3044,7 +3044,7 @@ class Network(object):
             Pca[l, 4 * (l + 1) - 1 - 1] = True
             Pdb[l, 4 * (l + 1) - 2 - 1] = True
             Pcb[l, 4 * (l + 1) - 1] = True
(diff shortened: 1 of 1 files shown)


jhillairet:
Thank you. I think it's only the beginning, as there are probably many other similar `is` in the code ;)

JAnderson419:
Partially addresses #368.

PR #381, opened 2020-08-06
Voltages and currents calculation in Circuit

jhillairet:
This PR adds the voltages and currents calculations at internal and external ports in `Circuit`. It also adds a dedicated notebook example, in which the results from full-wave and commercial circuit simulator are compared to `Circuit`. 

The example is a coaxial T like the following, for which a power/phase excitation is generated at each three ports. Methods have been in `Circuit` in order to extract the currents and voltages at ports from a given excitation at these ports (power/phase). 

circuit_vi_HFSS_Coaxial_T

Moreover, it can also give the currents and voltages at all internal nodes of a `Circuit`. In order to solve the current direction ambiguity, I've used the following rule:

circuit_vi_convention_currents

Hence, the voltages and currents at circuit's internal ports is obtain from:

```
>>> power = np.array([1, 2, 3])
>>> phase = np.deg2rad([-10, -20, +60])
>>> Is = crt.currents(power, phase)
>>> Vs = crt.voltages(power, phase)
>>> print(np.abs(Is))
[[0.02141285 0.02141285 0.13813294 0.13813294 0.50843596 0.50843596]
 [0.27407176 0.2740559  0.38946015 0.38943179 0.61798199 0.61795885]
 [0.21955389 0.21953817 0.42410748 0.42406926 0.3808368  0.38081916]]
>>> print(np.angle(Is, deg=True))
[[ 156.86473046  -23.13526954  -90.9007748    89.0992252    80.25214823  -99.74785177]
 [   2.6464094  -177.35027819    6.32827992 -173.66755158   44.66449676  -135.33335815]
 [  33.71397369 -146.28192378   -8.14673035  171.8584302    -4.40183035   175.60082298]]
```

and for external ports:

```
>>> Is = crt.currents_external(power, phase)
>>> Vs = crt.voltages_external(power, phase)
>>> print(np.abs(Is))
[[0.02141285 0.13813294 0.50843596]
 [0.27406227 0.38944806 0.61796953]
 [0.2195534  0.42408194 0.38082007]]
>>> print(np.angle(Is, deg=True))
[[156.86473046 -90.9007748   80.25214823]
 [  2.64953862   6.33353633  44.66515454]
 [ 33.71970029  -8.14144956  -4.40271754]]
```


alexpacini:
I just noticed this and I wanted to say that is a really nice feature! That is terrific work!

jhillairet:
Apart from using transmission line theory on a simple transmission line, or comparison with full-wave/circuit codes, I wonder if someone has a good idea for making (unit) tests of current/voltage calculation?

jhillairet:
OK, I think it's ready for merging. I've removed Python 3.5 from Travis Tests as it (suddenly?) triggered a SyntaxError into IPython.display when importing Image... (?)

jhillairet:
The associated documentation is here : 
https://scikit-rf.readthedocs.io/en/latest/examples/circuit/Voltages%20and%20Currents%20in%20Circuits.html

any comment appreciated.

wktanghk:
Hi, this is related to a separate Discussion thread I started a week ago (and I can close that discussion thread if needed). My goal is similar to what is described in this thread. I would like to know if it is possible to apply a voltage swing on one port, and measure the power on another port. Using the function described in this thread ['voltages_external']. The input stimulus is a power souce at one port, and the observed outputs are the voltages & currents on other ports.

In my case, I would like to do the opposite which is similar to an AC simulation. The input stimulus is a voltage source with a given input amplitude (e.g. 1V, 0phase) across frequency. And measure the delivered power on another port [i.e. 0.5x Re (VI*) flowing out at the output port].

One example could be a 2-port transformer. In a traditional schematic test bench, I can apply a Vac (AC voltage source) across the primary differential port. This would be the only connection on the input side (i.e. no other termination). On the output side, it would be loaded with some arbitrary load (e.g. 50ohm). And I could then measure the power delivering to this load by monitoring the voltage across the load & the current flowing into the load.

I wonder if there are existing functions in scikit-rf that can handle this. Or if there are other functions I can get started with, I would really appreciate if someone could recommend. Thanks!

PR #396, opened 2020-10-01
Additional Mixed Mode Tutorial

JAnderson419:
Added a basic mixed mode conversion tutorial for documentation. Moved @jrmagers tutorial into a new "Mixed Mode Analysis" folder that includes both notebooks.


skrf/tests/test_network.py
```diff
@@ -639,8 +640,15 @@ def test_noise_deembed(self):
           self.assertTrue(abs(gamma_opt_rb.s[0,0,0] - gamma_opt_set) < 1.e-1, 'nf not retrieved by noise deembed')
 
 
+    def test_se2gmm2se_mag(self):
+        ntwk4 = rf.Network(os.path.join(self.test_dir, 'cst_example_4ports.s4p'))
+        ntwk4t = deepcopy(ntwk4)
```
jhillairet, line 645: Why do you make a deepcopy of ntw4?

JAnderson419: se2gmm and gmm2se modify the network object directly, so I want a deepcopy around so that there is an unmodified original copy to compare against. Alternatively, to eliminate the deepcopy, we could just create another object from the same input file.

jhillairet:
Nice addition! Is it ready to be merged?

JAnderson419:
I think it is all ready on my end. I double checked @jrmagers notebook after moving it and I think it has all his changes updated as well (I started this branch a while back and had to merge his updates in).

PR #399, opened 2020-10-08
Clean Up Memory Leaks and Remove Python Warnings

bogdantabacaru:
Close file ID after loading the touchstone file because they get leaked into memory and cause resource issues.
Transform regular-expression strings into raw strings.


jhillairet:
Good catch, thank you.

PR #403, opened 2020-11-05
Replace B and C in s2a conversion

keikawa:
B and C were swapped in the s2a conversion equations.
I referred to the following document.
https://ieeexplore.ieee.org/document/275248


jhillairet:
The result of the Travis Continous Integration is not shown (?) but the tests are failing : https://travis-ci.org/github/scikit-rf/scikit-rf/jobs/741634342

I didn't check the change, but can you check your modification with respect to the existing tests? thanks.

keikawa:
The error existed in not s2a() but a2s().
I undid the yesterday's modify, and anew, I swapped S12 and S21 in the a2s conversion formulas. 
The tests were also passed, thank you for your advice!

By the way, the networks that used in the conversions tests (ntwk1.s2p, ntwk2.s2p, ntwk3.s2p) were reciprocal.
The tests using these networks overlook errors like this time (swap of 12 and 21).

jhillairet:
> The tests were also passed, thank you for your advice!

OK, nice. 

> By the way, the networks that used in the conversions tests (ntwk1.s2p, ntwk2.s2p, ntwk3.s2p) were reciprocal. The tests using these networks overlook bugs like this time (swap of 12 and 21).

I wonder if this was not made intentionally, because Numpy operation with (FxNxN) must often be transposed to get the expected result. 

Did you try using a non-symmetric matrix as a test case? Creating a random complex matrix of shape(s)=(FxNxN) and testing if s is equal to `a2s(s2a(s))` ? Adding such an additional test case would be very nice if you can.

keikawa:
I added a test case using a random complex matrix of shape(s)=(Fx2x2).
The modified formulas passed the additional tests.
Please check it out.

jhillairet:
Great fix and thank you for the additional test.

PR #423, opened 2021-02-04
subnetwork  

jhillairet:
Creating the `rf.subnetwork()` function and `Network.subnetwork()` method for creating sub-Network for a given list of port indices. 

This comes from two discussions on Twitter (@GProtoZeroW) and Slack (@danmcmahill), asking if there an easy way to select a subset of the ports from an N-port network, like for example picking out 4 ports from an 8 port (assuming all the unused ports are matched).

As an illustration, below three 2ports subnetwork are extracted from a reference 3ports Network. These 3 subnetworks are after recombined into a 3ports Network. 

```python
import skrf as rf

# 3 port Network
tee = rf.data.tee
# 2 port Networks as if one measures the tee with a 2 ports VNA
tee12 = rf.subnetwork(tee, [0, 1])  # 2 port Network from ports 1 & 2, port 3 matched
tee23 = rf.subnetwork(tee, [1, 2])  # 2 port Network from ports 2 & 3, port 1 matched
tee13 = rf.subnetwork(tee, [0, 2])  # 2 port Network from ports 1 & 3, port 2 matched

# recreate the original 3 ports Network from the thee 2-port sub-Networks
ntw_list = [tee12, tee23, tee13]
tee2 = rf.n_twoports_2_nport(ntw_list, nports=3)
print(tee2 == tee) # --> True

## same with the Network.subnetwork method
tee12 = tee.subnetwork([0, 1])
tee23 = tee.subnetwork([1, 2])
tee13 = tee.subnetwork([0, 2])
ntw_list = [tee12, tee23, tee13]
tee2 = rf.n_twoports_2_nport(ntw_list, nports=3)
print(tee2 == tee)  # --> True
```


skrf/network.py
```diff
@@ -4129,6 +4162,55 @@ def evenodd2delta(n, z0=50, renormalize=True, doublehalf=True):
     return n_delta 
 
 
+def subnetwork(ntw, ports, offby=1):
```
arsenovic, line 4165: consider ntwk,  since this is used elsewere,  not a big deal

arsenovic: `ntwk` instead of `nwk` that is

arsenovic:
great addition!

skrf/network.py
jhillairet: sure

jhillairet:
I will modify after the Measuring a [Multiport Device with a 2-Port Network Analyzer](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Measuring%20a%20Mutiport%20Device%20with%20a%202-Port%20Network%20Analyzer.html) example, to add this kind of example in the preamble to the mismatched network case.

jhillairet:
I think it's OK

arsenovic:
do you want me to merge this?

GProtoZeroW:
So the issue that I am seeing that I had to solve in
https://github.com/GProtoZeroW/scikit-rf_apps/blob/main/s2ps_to_sNp_tool_proto.ipynb
under the section `Working with sparse/overlab snps` is that Andewes data
set is not complete in 2/9 parameters and has multiplicity 2 in 1/9
parameters. See:
https://twitter.com/azonenberg/status/1355542642159833092
```
In the example I linked earlier in the thread I measured a 3-port device
between ports 1/3 and 2/3. This gives me S11/S33/S13/S31 and
S22/S33/S23/S32. So the actual S-matrix is S11 (unknown) S13 (unknown) S22
S23 S31 S32 (S33 measured twice)
```

Where I solved this issue in the inherited class `arb_snps_2_NportJoint` by
extracting port name using the `*p{i}{j}.snp` convention that is already
used by  `skrf.network.n_twoports_2_nport` and then looping over the
network set  of snps over the Sij elements to create a list of
measurements for each Sij and has action insert a "null" (s=1+0j) Sij
vector or to perform some action on the multiplicity listing for a Sij to
reduce it to a single network vector. The final result of a "complete" Sij
set is fed into `skrf.network.n_oneports_2_nport` to generate a full
network element that can be used for further actions such as writing a
single touchstone file. This is all well documented in the code in
`arb_snps_2_NportJoint.__init__` & `arb_snps_2_NportJoint.do_joint`

GProtoZeroW:
Also, I misspelled Andrew Zonenberg's name, got to love dysgraphia like a
bad Tx in a SERDES. And also is there any convention to extend the file
naming of `skrf.network.n_twoports_2_nport` '*p{i}{j}.s2p" to four-port or
greater measures. Where.s4p measures are now the norm when doing SI
measurements of high-speed differential lanes.

jhillairet:
> So the issue that I am seeing that I had to solve in https://github.com/GProtoZeroW/scikit-rf_apps/blob/main/s2ps_to_sNp_tool_proto.ipynb under the section `Working with sparse/overlab snps` is that Andewes data set is not complete in 2/9 parameters and has multiplicity 2 in 1/9 parameters. See: https://twitter.com/azonenberg/status/1355542642159833092 In the example I linked earlier in the thread I measured a 3-port device between ports 1/3 and 2/3. This gives me S11/S33/S13/S31 and S22/S33/S23/S32. So the actual S-matrix is S11 (unknown) S13 (unknown) S22 S23 S31 S32 (S33 measured twice)

OK, I understand. So this is a different and independent issue. 

`rf.n_twoports_2_nport()` function works only if the measurement set is complete, that is, for example if a n-ports is measured with a 2-ports VNA, it requires n(n-1)/2 distinct measurements (with non used ports matched)

jhillairet:
> 
> is there any convention to extend the file naming of `skrf.network.n_twoports_2_nport` '*p{i}{j}.s2p" to four-port or greater measures. Where.s4p measures are now the norm when doing SI measurements of high-speed differential lanes.

One should eventually write a `n_fourports_2_nport()` function or a generic `n_mports_2_nport`

GProtoZeroW:
Yeah that why I had to complete with Nulls and multiplicity handling

GProtoZeroW:
Also, I am not looking for a function, looking for a naming convention. So
for a 2 port, I have ports `i` & `j` whereas for a 4 port I have ports `i`
& `j` & `k` & `l`  . So say a 2 port on port 3&4 the file would be named
"*p34.s2p"  where ideally one would keep to the scikit-rf port ordering
working around the full port set. But for a 4 port where say for some
reason (bad layout typical), I had to violate convention and I measured 1-3
2-4 should I  then name the file like "*p1324.s4p" to then do a port name
extraction to make assigned port names to aid in converting to a
conventional port ordering. Cause technically if there was a full good set
of throughs in the cal the VNA don't care what your port ordering should be
when making the measurement

Looking to extend the code I wrote for unequal measurements to beyond s2ps,
cause physical access when micro probing might mean I could get a series of
4 port measures and then have to do a 3 port measure before having to
rotate the something to get the test of the measurement, Yeah abstraction
is heard but having sone convection to have as a reference keeps the
abstraction in for just care by case issues from becoming untraceable to
base code that the abstraction is based on

jhillairet:
> 
> Also, I am not looking for a function, looking for a naming convention. So for a 2 port, I have ports `i` & `j` whereas for a 4 port I have ports `i` & `j` & `k` & `l` . So say a 2 port on port 3&4 the file would be named "*p34.s2p" where ideally one would keep to the scikit-rf port ordering working around the full port set. But for a 4 port where say for some reason (bad layout typical), I had to violate convention and I measured 1-3 2-4 should I then name the file like "*p1324.s4p" to then do a port name extraction to make assigned port names to aid in converting to a conventional port ordering. 

In the way it is currently done in `n_twoports_2_nport`, the name of the file doesn't matter. What is only important is the `name` parameter of the subnetwork `Network` objects that are passed into the list of Network `ntwk_list`:

https://github.com/scikit-rf/scikit-rf/blob/c80f81ba03d661b19e1358dd9c56d367f19e9291/skrf/network.py#L4195-L4214

(NB: note the function fills the n-port with zero first, so it does what you've made)

If the port order is reversed from the one expect, the code flips the order. So I guess a `n_fourports_2_nport` function should have the same behaviour and not impose a pre-definite port ordering and adapt to what the user pass as subnetwork.

But this is a bit out of scope of this PR. You should eventually open a dedicated issue/PR.

jhillairet:
> do you want me to merge this?

Yes, I think it's OK.

PR #427, opened 2021-02-19
Fix rf.io.Touchstone.get_comment_variables() parsing to handle units, decimals properly

JAnderson419:


JAnderson419:
closes #427

jhillairet:
Good catch, thank you!

PR #428, opened 2021-02-23
write_touchstone: write port names to file

luar123:
If port names are available in Network.port_names, write these to the touchstone file as '! Port[#] = name'

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1838,8 +1838,8 @@ class Network(object):
     def write_touchstone(self, filename=None, dir=None,
                          write_z0=False, skrf_comment=True,
                          return_string=False, to_archive=None,
-                         form='ri',format_spec_A='{}',format_spec_B='{}',
-                         format_spec_freq='{}'):  
+                         form='ri', format_spec_A='{}', format_spec_B='{}',
+                         format_spec_freq='{}'):
         """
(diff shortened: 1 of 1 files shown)


skrf/network.py
```diff
@@ -1955,6 +1955,16 @@ def get_buffer():
...
             output.write('# {} S {} R {} \n'.format(self.frequency.unit, form, str(abs(self.z0[0, 0]))))
 
+            # write ports
+            try:
+                if self.port_names and len(self.port_names) == self.number_of_ports:
+                    ports = ''
+                    for port_idx, port_name in enumerate(self.port_names):
+                        ports += '! Port[{}] = {}\n'.format(port_idx, port_name)
+                    output.write(ports)
+            except(AttributeError):
```
jhillairet, line 1965: I would suggest using the more pythonic :
`except AttributeError:`

jhillairet:
Thank you for your contribution. Do you think you could add a specific test for this improvement?

skrf/network.py
luar123: Thanks for the suggestion, I have changed it throughout the write_touchstone() method. And also fixed some white space formating warnings.

jhillairet: great, thank you.

jhillairet:
thank you for the improvement

Issue #434, opened 2021-03-11
Non Harmonic sweep


FranzForstmayr:
I'm using a non-harmonic sweep and wanted to calculate the TDR. I recognized, that the step_response function does not check if the sweep is harmonic and assumes it is. 
Wouldn't it be better, if the frequency class could check if the sweep is harmonic, so that the step_response could use a non uniform fourier transformation or interpolate the support points (probably easier)?
Can you confirm this behavior or did I just something wrong?

jhillairet:
What do you mean by non-harmonics: frequency points are not evenly spaced (changing frequency step size)? 

Indeed, the code does use standard fft for time gating, so it will not work properly on unevenly spaced data. 

You are right that a test should be made somewhere to indicate users that something is wrong. It has to be discussed if the data should be interpolated under the hood silently or not... could you provide a data example so that one could add this feature and forge a dedicated test?

FranzForstmayr:
Yes, I changed the step size using a segmented sweep in a R&S VNA. I think there would be an similar issue using an logarithmic sweep.

I would say either throw a Warning and interpolate or throw an Error that the feature is not implemented (yet).
Probably the mentioned non-uniform fourier transformation could help without interpolating. 

I can deliver test data for a segmented use case. Should I integrate it in skrf/tests and create a testcase?

jhillairet:
I would personally be in favour of throwing a not implement error, as interpolating may result in specific case-to-case problems... If you want to make a pull request, with a test that check an error is raiser in such case, you are very welcome!

FranzForstmayr:
I'll give it a try :)

jhillairet:
fixes with #435. thank you!

PR #435, opened 2021-03-17
Fix/434

FranzForstmayr:
Fixing #434 

Also merged code for step_response and impulse response

### DIFF
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -320,7 +320,6 @@ class Frequency(object):
         set the number of points in the frequency
         '''
 
-        print("self.sweep type")
         if self.sweep_type == 'lin':
             self.f = linspace(self.start, self.stop, n)
         elif self.sweep_type == 'log':
@@ -420,8 +419,6 @@ class Frequency(object):
         sets the frequency object by passing a vector in Hz
         '''
         self._f = npy.array(new_f)
-
-        print(new_f)
 
         if npy.allclose(    new_f,  
                             npy.linspace(new_f[0], new_f[-1], len(new_f))):
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -128,7 +128,7 @@ class Frequency(object):
 
         if sweep_type.lower() == 'lin':
             self._f = linspace(start, stop, npoints)
-        elif sweep_type.lower() == 'log':
+        elif sweep_type.lower() == 'log' and start > 0:
             self._f = geomspace(start, stop, npoints)
         else:
             raise ValueError('Sweep Type not recognized')
@@ -423,7 +423,7 @@ class Frequency(object):
         if npy.allclose(    new_f,  
                             linspace(new_f[0], new_f[-1], len(new_f))):
             self.sweep_type = 'lin'
-        elif npy.allclose(  new_f, 
+        elif new_f[0] and npy.allclose(  new_f, 
                             geomspace(new_f[0], new_f[-1], len(new_f))):
             self.sweep_type = 'log'
         else:
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -127,9 +127,9 @@ class Frequency(object):
         stop = self.multiplier * stop
 
         if sweep_type.lower() == 'lin':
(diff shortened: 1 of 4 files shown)


FranzForstmayr:
Is the Python2.7 build still relevant? 
unittest's assertFail was introduced in Python3 obviously.

JAnderson419:
Re: Python 2.7. The others can correct me if I am wrong, but I believe it was decided just a week or two ago to finally drop support for 2.7, so you should be good in that department. We obviously have a bit of work to do still with CI/documentation to make that official for the next version release.

jhillairet:
That's right, Python 2 is no more supported.

skrf/frequency.py
```diff
@@ -318,7 +319,14 @@ def npoints(self, n):
         '''
         set the number of points in the frequency
         '''
-        self.f = linspace(self.start, self.stop, n)
+
+        if self.sweep_type == 'lin':
+            self.f = linspace(self.start, self.stop, n)
```
jhillairet, line 324: should we used 'linear' (like in matplotlib) or just 'lin' as proposed ? I'm more in favor in 'linear' because I'm used to matplotlib, but tbd

skrf/frequency.py
```diff
@@ -32,7 +32,7 @@
 '''
 
 # from matplotlib.pyplot import gca,plot, autoscale
-from numpy import pi, linspace, logspace
+from numpy import pi, linspace, geomspace
```
jhillairet, line 35: nice conversion, I didn't know geomspace

skrf/frequency.py
```diff
@@ -412,6 +420,15 @@ def f(self,new_f):
         '''
         self._f = npy.array(new_f)
 
+        if npy.allclose(    new_f,  
+                            linspace(new_f[0], new_f[-1], len(new_f))):
+            self.sweep_typesweep_type = 'lin'
```
jhillairet, line 425: there is a typo here I think

skrf/network.py
```diff
@@ -3223,33 +3223,31 @@ def step_response(self, window='hamming', n=None, pad=1000):
...
+        NotImplementedError
+            If used with non equidistant sampled frequency vector
+
         See Also
         -----------
             impulse_response
             extrapolate_to_dc
         """
-        if self.nports != 1:
-            raise ValueError('Only one-ports are supported')
```
jhillairet, line 3232: why did you remove that part?

skrf/network.py
```diff
@@ -3223,33 +3223,31 @@ def step_response(self, window='hamming', n=None, pad=1000):
...
-        else:
-            t = npy.flipud(npy.linspace(.5 / fstep, -.5 / fstep, n + 1, endpoint=False))
-            t = t[:-1]
-        if window != None:
-            w = self.windowed(window=window, normalize=False, center_to_dc=True)
-        else:
-            w = self
-        return t, npy.cumsum(mf.irfft(w.s, n=n).flatten())
+        t, y = self.impulse_response(window=window, n=n, pad=pad, bandpass=False)
+        return t, npy.cumsum(y)
```
jhillairet, line 3250: Did you check this is correct is any case (I didn't)?

skrf/frequency.py
FranzForstmayr: I would prefer 'linear' too, however I didn't want to introduce a breaking change.
Personally I prefer Enums instead of strings in general.

skrf/network.py
FranzForstmayr: This constraint is checked in the impulse_response function.

skrf/network.py
FranzForstmayr: Only for an linear sweep starting at DC, otherwise an Exception is raised anyway.
Check this [gist](https://gist.github.com/FranzForstmayr/0f22a87c39aa80e7fb8b0ff3256b7a4a#file-step-py):

FranzForstmayr:
I think I have to investigate a small issue in @f.setter tomorrow. So please wait with merging.

skrf/network.py
jhillairet: ha yes, OK

PR #433, opened 2021-03-05
docs changes, Network(fid) fix, and some stringio half-ways

arsenovic:

PR #437, opened 2021-03-19
fix network contruction from file-like obj

vik-s:


jhillairet:
doesn'it collide with the PR #433?

vik-s:
It looks like this has already been fixed in #433 ... so i close out this PR.

jhillairet:
thank you for your proposal anyway!

PR #438, opened 2021-03-24
Use relative paths for import

luar123:
Fixes two imports from skrf.

### DIFF
--- a/skrf/plotting.py
+++ b/skrf/plotting.py
@@ -1444,7 +1444,7 @@ def stylely(rc_dict={}, style_file = 'skrf.mplstyle'):
     loads the rc-params from the specified file (file must be located in skrf/data)
     '''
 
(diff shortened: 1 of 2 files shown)


jhillairet:
did you run into problems before these changes?

luar123:
Yes, I have scikit-rf installed globally in my python installation and additionally, in one project, I use the latest git version. So I have this version in the directory of the project. This is working fine for everything except these two files which will import the global version. Therefore, if I need to change something in plotting or data/ this is ignored as the global version is used.
Additionally, using pyinstaller both module version are included in the created exe, which makes this file unnecessary large.

Also I think it is not good to rely on the module name within the module itself. E.g. if someone wants to rename the module to something other than skrf it will break.

jhillairet:
indeed, good points. Thank you.

PR #442, opened 2021-04-08
Add option for renaming files in data_grabber

wbarnha:
I wanted to add a feature to data_grabber.py so that I could rename the files by right-clicking on them rather than going to manually the plot legend.

Also some other changes are simple syntax reformatting.

### DIFF
--- a/qtapps/skrf_qtwidgets/networkListWidget.py
+++ b/qtapps/skrf_qtwidgets/networkListWidget.py
@@ -224,10 +224,9 @@ class NetworkListWidget(QtWidgets.QListWidget):
                 self.takeItem(self.row(item))
 
     def rename_item(self):
-        items = self.selectedItems()
-        if len(items) == 1:
-            item = items[0]
-            widgets.rename_file_label(self, item)
+        item = self.currentItem()
+        item.setFlags(item.flags() | QtCore.Qt.ItemIsEditable)
+        self.editItem(item)
 
     def get_save_which_mode(self):
         """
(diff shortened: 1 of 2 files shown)


wbarnha:
Apparently you could do this by just double-clicking the file name, this never occurred to me until now. Here's the code anyway if you still want it.

jhillairet:
seems OK to me, but eventually @dvincentwest could review as he is the author of the qt widget code.

dvincentwest:
I haven't spent a lot of time dealing with PyQt for almost 3 years now, so unfortunately I can't give you much direction.  I tested the changes, and they work, and I think that having a right click option is a good idea.  I think also adding an "F2" key press which is probably a Qt named constant somewhere would also be a good idea.

I think it would be better though if you can figure out how to route the editing through the standard QListWidgetItem ItemDelegate editor, which is what happens when you double click on the item currently.  Right now you pop up a separate line editor on top that is narrower and also offset which looks a little distracting, and also that you can't escape out of to cancel your edit.  Once you hit the right-clicked option, your only way out is to accept your changes, but sometimes you don't want to do that.

Basically, I think what you want is for the right click action to do the same thing that the double click action does, not a different thing.  I don't know how to do that though.  Been too long

wbarnha:
Got the requested change to have double click editing to be the same for right click rename. Apparently all you need to do is:
```
    def rename_item(self):
        item = self.currentItem()
        item.setFlags(item.flags() | QtCore.Qt.ItemIsEditable)
        self.editItem(item)
```

dvincentwest (approved):
I tested and the changes behave as expected

jhillairet:
> I tested and the changes behave as expected

OK, thank you.

jhillairet:
> Got the requested change to have double click editing to be the same for right click rename. Apparently all you need to do is:

@wbarnha is it OK to merge?

wbarnha:
Yeah, you can go ahead and merge now.

jhillairet:
Thank you.

PR #444, opened 2021-04-12
Lazy load vectorFitting to remove matplotlib dependency

FranzForstmayr:
The vectorFitting module introduced a matplotlib dependency, which is removed with this patch.

### DIFF
--- a/skrf/__init__.py
+++ b/skrf/__init__.py
@@ -26,7 +26,7 @@ from . import constants
 from . import util
 from . import io
 from . import instances
-# from . import vectorFitting
+from . import vectorFitting
 
 
 # Import contents into current namespace for ease of calling
@@ -43,7 +43,7 @@ from .io import *
 from .constants import * 
 from .taper import * 
 from .instances import *
-# from .vectorFitting import *
+from .vectorFitting import *
 
 # Try to import vi, but if except if pyvisa not installed
 try:
@@ -84,7 +84,6 @@ def setup_pylab():
         return False
 
     from . import plotting
-    from . import vectorFitting
     plotting.setup_matplotlib_plotting()
 
     global saf, stylely
--- a/skrf/__init__.py
+++ b/skrf/__init__.py
@@ -26,7 +26,7 @@ from . import constants
 from . import util
 from . import io
 from . import instances
-from . import vectorFitting
+# from . import vectorFitting
 
 
 # Import contents into current namespace for ease of calling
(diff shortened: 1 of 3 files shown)


skrf/__init__.py
```diff
@@ -43,7 +43,7 @@
 from .constants import * 
 from .taper import * 
 from .instances import *
-from .vectorFitting import *
+# from .vectorFitting import *
```
jhillairet, line 46: by doing so, the `VectorFitting` class is no more available from the root. Is it on purpose?

Vinc0110:
What exactly is the issue? I don't understand the phrase "lazy load" in this context, but maybe vectorfitting.py could be improved directly instead of un-importing it?

FranzForstmayr:
Ahh Sorry, this was not intended like that!

I just wanted to be able to import skrf without matplotlib installed.
I was pretty sure, that all my tests succeded without having matplotlib installed. I have to investigate!

FranzForstmayr:
Or would it be better to add matplotlib in setup.py for automated install?

Vinc0110:
I see. matplotlib is not stricly required for vectorfitting.py. Maybe we can test if matplotlib is available and make a conditional import, so the rest of the code would still work without matplotlib. Just like the plotting module is already being initialized, but I'd rather do this directly in vectorFitting.py:

https://github.com/scikit-rf/scikit-rf/blob/feb404d8fdc9de159ed223b00c130efc92ae4f00/skrf/__init__.py#L79-L106

Vinc0110 (approved):
Thanks for those changes. Seems like an elegant solution.

jhillairet:
For reference, a recent message has been posted on the mailing list from Ken Harris:
> it's better not to put mplt.figure() & mplt.show() calls in skrf.vectorFitting.plot_* calls.  Better to do like the other plot routines (eg : plot_s_smith) and allow an optional "ax" parameter.

Vinc0110:
I don't mind changing the plotting routines according to Ken's suggestion. It should be simple enough. But I think it's better to separate that from this PR, unless @FranzForstmayr likes to do that here. Otherwise, I will open a new PR later for the changes in the plotting output.

FranzForstmayr:
> I don't mind changing the plotting routines according to Ken's suggestion. It should be simple enough. But I think it's better to separate that from this PR, unless @FranzForstmayr likes to do that here. Otherwise, I will open a new PR later for the changes in the plotting output.

I would appreciate, if you could have a look at this! I'm not that deep in the vectorfitting module!

jhillairet:
> 
> 
> I don't mind changing the plotting routines according to Ken's suggestion. It should be simple enough. But I think it's better to separate that from this PR, unless @FranzForstmayr likes to do that here. Otherwise, I will open a new PR later for the changes in the plotting output.

OK. So I'm merging this PR. Thank you.

PR #446, opened 2021-04-16
Fixed a/an typo in Interpolating Network Sets

NikoDaGreat:
`Interpolating Network Sets.ipynb` had the following typo:

a old -> an old


jhillairet:
great , thank you for the correction !

Issue #384, opened 2020-08-08
ValueError when using rf.connect()


jhillairet:
The issue has been spotted in the documentation, where an error is generated inside the following notebook:
https://scikit-rf.readthedocs.io/en/latest/examples/circuit/Lumped%20Element%20Circuits.html

Minimum example:
```python
import skrf as rf

freq = rf.Frequency(1, 10, 10)
media = rf.DefinedGammaZ0(freq, z0=50)

line = media.line(d=1)
tee = media.tee()

rf.connect(tee, 0, line, 0)
```
which generates:
```
Traceback (most recent call last):

  File "C:\Users\JH218595\Documents\tests\sanstitre0.py", line 15, in <module>
    rf.connect(tee, 0, line, 0)

  File "C:\Users\JH218595\Documents\scikit-rf\skrf\network.py", line 3443, in connect
    to_ports=to_ports)

  File "C:\Users\JH218595\Documents\scikit-rf\skrf\network.py", line 2674, in renumber
    raise ValueError('from_ports and to_ports must have the same set of indices')

ValueError: from_ports and to_ports must have the same set of indices
```

sublimet:
I have a problem which is not the same as one discussed here but
may have the same source of bug.

My problem is:
Port numbering of networks created by connecting 3-ports network and
2-port-fixture network has been changed in recent version of scikit-rf.

```
(In scikit-rf 0.15.4)
1T2---1-2 
 3
=> 2T1
    3
(In scikit-rf 0.14.9)
=> 1T2
    3
```

It seems that the change in Network.py have critial impact..
```
(In scikit-rf 0.15.4) line 3443-
        to_ports.pop(k-1);
        to_ports.append(k-1)
```
```
(In scikit-rf 0.14.9) 
        to_ports.pop(k)
        to_ports.append(k)
```
I'm glad if this behaviour will be corrected.
Thank you.

jhillairet:
Hello, thanks for reporting

> ```
> (In scikit-rf 0.15.4)
> 1T2---1-2 
>  3
> => 2T1
>     3
> (In scikit-rf 0.14.9)
> => 1T2
>     3
> ```

Can you elaborate more on your setup? I don't understand your sketch.

> I found that the change is introduced intensionally (#330).

Indeed, the origin of this change is discussed in PR  #289, and was though at that time as a good idea ;) (but maybe not)

sublimet:
Thank you for the reply.  I have edited my sketch.
[...]

PR #448, opened 2021-04-19
Bug fix in delay function

hohe:
The delay function delays always port=0 independent from the function
argument port.
The submitted change not only fixes this but also allows to apply the
function to Networks with more than 2 ports.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2771,14 +2771,15 @@ class Network(object):
         if d ==0:
             return self
         d=d/2.
-        if self.nports >2:
-            raise NotImplementedError('only implemented for 1 and 2 ports')
         if media is None:
             from .media import Freespace
             media = Freespace(frequency=self.frequency,z0=self.z0[:,port])
 
         l =media.line(d=d, unit=unit,**kw)
-        return l**self
+        if port < 1:
+            return connect(l, 1, self, port)
+        else:
+            return connect(self, port, l, 0)
 
     def windowed(self, window=('kaiser', 6), normalize=True, center_to_dc=None):
         '''
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2776,7 +2776,7 @@ class Network(object):
             media = Freespace(frequency=self.frequency,z0=self.z0[:,port])
 
         l =media.line(d=d, unit=unit,**kw)
-        if port < 1:
+        if port == 0:
             return connect(l, 1, self, port)
         else:
             return connect(self, port, l, 0)
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2776,10 +2776,7 @@ class Network(object):
             media = Freespace(frequency=self.frequency,z0=self.z0[:,port])
 
         l =media.line(d=d, unit=unit,**kw)
-        if port == 0:
-            return connect(l, 1, self, port)
-        else:
-            return connect(self, port, l, 0)
+        return connect(self, port, l, 0)
 
     def windowed(self, window=('kaiser', 6), normalize=True, center_to_dc=None):
         '''
@@ -3497,8 +3494,8 @@ def connect(ntwkA, k, ntwkB, l, num=1):
     if ntwkB.nports == 2 and ntwkA.nports > 2 and num == 1:
         from_ports = list(range(ntwkC.nports))
         to_ports = list(range(ntwkC.nports))
-        to_ports.pop(k-1);
-        to_ports.append(k-1)
+        to_ports.pop(k);
+        to_ports.append(k)
 
         ntwkC.renumber(from_ports=from_ports,
                        to_ports=to_ports)
--- a/doc/source/examples/mixedmodeanalysis/Mixed Mode S and Impedance Transformation.ipynb
+++ b/doc/source/examples/mixedmodeanalysis/Mixed Mode S and Impedance Transformation.ipynb
@@ -78,10 +78,7 @@
     "\n",
     "# connect the transformer to the common mode port\n",
     "# common mode port impedance transformed from 50 Ω to 25 Ω\n",
-    "mm_ntwk_t = skrf.connect(mm_ntwk,2,transformer,1)\n",
-    "\n",
-    "# skrf.connect() re-ordered the ports.  Change them back so they are consistent with the schematic above.\n",
-    "mm_ntwk_t.renumber([0,1,2],[0,2,1])"
+    "mm_ntwk_t = skrf.connect(mm_ntwk,2,transformer,1)"
    ]
   },
   {
@@ -140,7 +137,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.7"
+   "version": "3.7.8"
   }
  },
  "nbformat": 4,
--- a/doc/source/skrf_Network_symbols.svg
+++ b/doc/source/skrf_Network_symbols.svg
@@ -11,12 +11,165 @@
    inkscape:version="1.0 (4035a4fb49, 2020-05-01)"
    id="svg13093"
    version="1.1"
-   viewBox="0 0 519.87408 188.13292"
-   height="188.13292mm"
-   width="519.87408mm">
+   viewBox="0 0 592.13145 450.3762"
+   height="450.37619mm"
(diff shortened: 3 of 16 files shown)


skrf/network.py
```diff
@@ -2771,14 +2771,15 @@ def delay(self, d, unit='deg', port=0, media=None,**kw):
...
         d=d/2.
-        if self.nports >2:
-            raise NotImplementedError('only implemented for 1 and 2 ports')
         if media is None:
             from .media import Freespace
             media = Freespace(frequency=self.frequency,z0=self.z0[:,port])
 
         l =media.line(d=d, unit=unit,**kw)
-        return l**self
+        if port < 1:
```
jhillairet, line 2779: Just wondering: do you have any good reason to use < 1 instead of == 0?
```suggestion
        if port == 0:
```

jhillairet:
Thank you for both noticing and correcting. It would be super nice if you could add some new dedicated tests as well.

skrf/network.py
hohe: I changed to ==1

hohe:
I created two tests to test the delay-function().

It's the first time I wrote any tests, I checked them and I think everything is working as expected, but it might still be a good idea to double check.

jhillairet:
I was wondering if this could be extanded to N-ports with N>2, but seems not. 

The following tests work for a 2-port:

```python
import numpy as np
import skrf as rf

#%% Two Port Tests
delay = 3e-9  # s
ntwk1 = rf.data.ring_slot
ntwk1_delay_0 = ntwk1.copy().delay(delay, unit='s', port=0)
ntwk1_delay_1 = ntwk1.copy().delay(delay, unit='s', port=1)

# testing (relative) differential group delay for S21 and S12
np.testing.assert_allclose((ntwk1_delay_0.group_delay - ntwk1.group_delay)[:,1,0] / (delay/2), 1)
np.testing.assert_allclose((ntwk1_delay_0.group_delay - ntwk1.group_delay)[:,0,1] / (delay/2), 1)
np.testing.assert_allclose((ntwk1_delay_1.group_delay - ntwk1.group_delay)[:,1,0] / (delay/2), 1)
np.testing.assert_allclose((ntwk1_delay_1.group_delay - ntwk1.group_delay)[:,0,1] / (delay/2), 1)
```

but seems not for a 3-ports like a T:

```python
#%% Three Port Test
delay = 3e-9  # s
tee = rf.data.tee
tee_delayed_0 = rf.data.tee.delay(delay, unit='s', port=0)

np.testing.assert_allclose((tee_delayed_0.group_delay - tee.group_delay)[:,1,0] / (delay/2), 1)  # fails
```

I didn't look into the theory, maybe someone here did one day? (@arsenovic ?). This could explain why the code was intended for two port only...

hohe:
Hi Julien, thanks for testing and reporting.

I think I found another bug in the connect function and I tried to fix it. It needs further testing. I will try to write further tests as soon as I get time to do it.
At least with these additions your test is also working.

hohe:
I see that now various other tests in test_circuit are failing.

I followed this, and it is related to the connect()-function.
To me it is not clear what to expect as the resulting port order of the connect()-function. Especially in the case of connecting one side of two-port to an N-port, I would expect that the result has the port order of the N-Port. But I'm not sure if that is intended. I can't find this info in the documentation, but perhaps I am missing something.

jhillairet:
> At least with these additions your test is also working.

I've just tried 
```python

delay = 3e-9  # s
tee = rf.data.tee
tee_delayed_0 = rf.data.tee.delay(delay, unit='s', port=0)

np.testing.assert_allclose((tee_delayed_0.group_delay - tee.group_delay)[:,1,0] / (delay/2), 1)  # fails
```
but still got an error. I obtain an array of `3.23529412e-10+0.j` values for `(tee_delayed_0.group_delay - tee.group_delay)[:,1,0]`
 
> 
> I see that now various other tests in test_circuit are failing.
> 
> I followed this, and it is related to the connect()-function.
> To me it is not clear what to expect as the resulting port order of the connect()-function. Especially in the case of connecting one side of two-port to an N-port, I would expect that the result has the port order of the N-Port. But I'm not sure if that is intended. I can't find this info in the documentation, but perhaps I am missing something.

Indeed, the tests for Circuit compare networks obtained with connect() vs networks obtained with Circuit. If the order of port is changed, the results will fail. 

This has also been discussed in #268. The  `k-1` comes from issue #289 (and implemented in #330)

hohe:
Thanks for the hints regarding the history.

I think the solution at least to me seems to be logical in all cases, even if it breaks backward compatibility.
The intention of the code is to preserve the original port order of ntwkA in case a twoport is connected to one of its ports. I wrote a tests which shows that this is the case for any nport network ntwkA with a twoport network ntwkB now.

I'm not sure why issue [289] was raised at all, since there `num=2` was used and the renumber code is only triggered for `num=1`now.

I understand that this breaks backward compatibility at some places and tests. Nevertheless, without this correction the code for renumbering the ports in this special case makes no sense.
So, I see two solutions:
a) remove this part of the code completely
b) use the proposed solution

I think both would probably break backwards compatibility.

Please let me know, what you think. If I find time, I can also help to fix the issues which are upcoming because of the break of backwards compatibility.

jhillairet:
> I think the solution at least to me seems to be logical in all cases, even if it breaks backward compatibility.
> The intention of the code is to preserve the original port order of ntwkA in case a twoport is connected to one of its ports. I wrote a tests which shows that this is the case for any nport network ntwkA with a twoport network ntwkB now.
>
>
> I'm not sure why issue #289 was raised at all, since there `num=2` was used and the renumber code is only triggered for `num=1`now.

Yes, I think you're right, and we missed something somewhere, me first. The issue #289 has been reported in Aug 2019, but it should have been prevented by the condition  `num=1` added before in #268 (May 2019). 

We didn't check if the user used the very last GitHub version (since no pip/conda version was released until Dec2019), so maybe a confusion arose here. 
 
> I understand that this breaks backward compatibility at some places and tests. Nevertheless, without this correction the code for renumbering the ports in this special case makes no sense. @arsenovic what do you think?

I think the proposed solution is logical. Maybe this would avoid the confusions reported in #384 (didn't check). We should eventually think twice and add more documentation and tests to make it very clear for the future (with various different characteristic impedances for each ports).

> Please let me know, what you think. If I find time, I can also help to fix the issues which are upcoming because of the break of backwards compatibility.

Thanks !

To fix the test in test_circuit, you can [revert to port numbers as before the PR](https://github.com/scikit-rf/scikit-rf/pull/330/files#diff-5dd3c7b052dbde95103ee595461f471ddd60b4cf545b412c6b9738f7da4af6fbL198-L613) in:

hohe:
I updated the test based on your input at [384] using z0 as a port indicator.
I also updated all other tests in test_circuit to be compliant with the fix.
This also fixes [384].

However, it might be worth to consider the comment from caspar-iqm in [384]. I basically like the fixed version now, as I think it is more intuitive if you are dealing with (nport, twoport)-networks. However it is different from (nport, mport)-behaviour and introduces therefore a special case.

I think we should somehow describe this in the documentation.

In the documentation I found that we probably also have to fix it in the example "Mixed Mode S-Parameters & Impedance Transformation", at [3] the following must be removed:
```
# skrf.connect() re-ordered the ports.  Change them back so they are consistent with the schematic above.
mm_ntwk_t.renumber([0,1,2],[0,2,1])
```

jhillairet (approved):
OK to me. I've added supplementary (illustrated) documentation (which you can eventually correct or complete) to your PR.

hohe:
Hello Julien,
thanks for the documentation, I think this is a great improvement.
I slightly amended one figure.

hohe:
Hello Julien,
thanks for the documentation, I think this is a great improvement.
I slightly amended one figure.

hohe:
Hello Julien,
thanks for the documentation, I think this is a great improvement.
I slightly amended one figure.

jhillairet:
@hohe, ready to be merged?

hohe:
Sure, I'd love to.

jhillairet:
Merged. 

A [special tutorial dedicated to Network connections ](https://scikit-rf.readthedocs.io/en/latest/tutorials/Connecting_Networks.html)has also been added.

PR #450, opened 2021-04-19
Vector Fitting Improvements

Vinc0110:
As discussed in https://github.com/scikit-rf/scikit-rf/pull/444, this PR is adding optional Axes parameters to the plotting functions in vectorFitting.py to control the output with matplotlib. The function docstrings and the example notebooks are adapted accordingly.

Some minor changes: Reorganization and correction of some docstrings.

### DIFF
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -607,7 +607,7 @@ class VectorFitting:
         return resp
 
     @check_plotting
-    def plot_s_db(self, i, j, freqs=None):
+    def plot_s_db(self, i, j, freqs=None, ax=None):
         """
         Plots the magnitude in dB of the response **S_(i+1,j+1)** in the fit.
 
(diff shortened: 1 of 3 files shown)


Vinc0110:
The code tests fail during the build of the vector fitting example notebooks because of a TypeError:
`TypeError: plot_s_mag() got an unexpected keyword argument 'ax'`.

That is because the notebooks are already updated to match the modified plotting functions, but of course the Juypter kernel is still importing `skrf` with the old versions of those functions without the `ax` parameter.

skrf/vectorFitting.py
```diff
@@ -689,46 +719,57 @@ def plot_pz(self, i, j):
...
 
+        ax : :class:`matplotlib.axes.AxesSubplot` object or None
+            matplotlib axes to draw on. If None, the current axes is fetched with gca().
+
         Returns
         -------
         None
         """
 
+        if ax is None and mplt is not None:
```
FranzForstmayr, line 730: The mplt check is unnecessary here, it's already checked by the @check_plotting decorator

Vinc0110: You're right, thanks! I updated those checks in https://github.com/scikit-rf/scikit-rf/pull/450/commits/4b8d25947e65e8cf6167886f1fb91300f5ae1c1a

Vinc0110:
@jhillairet How do we proceed here? The PR is basically finished, except for the problems with the code testing related to the example notebooks mentioned above. 

We could implement these changes in two steps to pass the code tests, or you just do ahead and merge it as is.

jhillairet:
> @jhillairet How do we proceed here? The PR is basically finished, except for the problems with the code testing related to the example notebooks mentioned above.

OK, I'm merging for now. Thank you.

Issue #449, opened 2021-04-19
Typehints in scikit-rf


FranzForstmayr:
I would like to add python typehints in skrf, to improve the quality of the autosuggestions in IDEs.
I use vscode, which takes the information out of typehints, I think most IDEs do it like that.

I would start a pull request, if this is ok?

arsenovic:
it is welcomed.

FranzForstmayr:
Need your opinion here...

Type annotations, especially with mypy, work very well with pure python types, but get tricky with numpy types.
For example this function:
```python
def complex_2_magnitude(input):
    '''
    returns the magnitude of a complex number.
    '''
    return abs(input)
```
returns a `float` when called with any Python `number` and returns a `npy.ndarray`, if called with an `npy.ndarray`. So the return type would basically be a `Union`, however `Union`s as return type is not recommended.
https://mypy.readthedocs.io/en/stable/kinds_of_types.html#union-types

The big players (Numpy/Scipy/Pandas) haven't adopted annotations completely, so there's less documentation about this right now.

**Inline Type Hints**
This annotations are interpreted correct when doint the following (in vscode at least)

```python
@overload
def complex_2_magnitude(input: complex) -> float: ...
@overload
def complex_2_magnitude(input: npt.ArrayLike) -> npy.ndarray: ...

## simple conversions
def complex_2_magnitude(input):
    '''
    returns the magnitude of a complex number.
    '''
    return abs(input)
```
This method works for most of the function in math_functions.py, however we would add lot of boilerplate code.

**Type Hints in a stub file**
To minimize boilerplate code within the functions, it's also possible to generate stub files, like in this repository:
https://github.com/python/typeshed/tree/master/stubs
Typeshed is a collection of stubfiles, especially to support static typing in IDEs and for mypy.
One yould also add and package the stubs directly in the skrf repository.
However another file containing stubs seem to me prone to get out of sync.

The first method seems reliable, with the drawback to add lot's of code.
Maybe anyone has some thoughts about this issue?

PS:
Most of the other files could work very well, but I wanted to start with the base functions, as a bottom up approach looks best to me. That's why the math_functions made me some headache.

jhillairet:
> **Inline Type Hints**
> This annotations are interpreted correct when doint the following (in vscode at least)
> 
> ```python
> @overload
> def complex_2_magnitude(input: complex) -> float: ...
> @overload
[...]

PR #451, opened 2021-04-19
Ready : Start to add type hints in network.py

FranzForstmayr:
Start to add type hints in this PR, this will probably take some time.


FranzForstmayr:
Link to #449

requirements.txt
```diff
@@ -13,5 +13,6 @@ networkx
 pathlib
 nose
 nose-exclude
+typing-extensions
 xlwt
 git+https://github.com/scikit-rf/scikit-rf.git@master
```
jhillairet, line 18: please update with the master branch, the list of requirements has changed since #459 and #460

skrf/media/media.py
```diff
@@ -31,11 +31,11 @@
 
 from ..constants import to_meters ,ZERO
 
-from abc import ABCMeta, abstractmethod, abstractproperty
+from abc import ABC, abstractmethod
 import re
 from copy import deepcopy as copy
 
-class Media(object):
+class Media(ABC):
```
jhillairet, line 38: I wonder if some users have instantiated directly Media. After that, it will throw an error. @arsenovic, what do you think?

skrf/network.py
```diff
@@ -1133,7 +1131,7 @@ def z0(self):
             return self._z0
 
     @z0.setter
-    def z0(self, z0):
+    def z0(self, z0: Union[Number, npy.ndarray]) -> None:
```
jhillairet, line 1134: could be a list of float/complex as weel as input

skrf/network.py
```diff
@@ -1198,7 +1196,7 @@ def frequency(self):
             return self._frequency
 
     @frequency.setter
-    def frequency(self, new_frequency):
+    def frequency(self, new_frequency: Union[Frequency, int, npy.ndarray]) -> None:
```
jhillairet, line 1199: frequencies could be set with a list also

skrf/networkSet.py
```diff
@@ -658,7 +658,7 @@ def cov(self, **kw) -> npy.ndarray:
         return npy.array([npy.cov(k.T) for k in smat])
 
     @property
-    def mean_s_db(self) -> 'Network':
+    def mean_s_db(self) -> Network:
```
jhillairet, line 661: for my knowledge: why do you remove the quotes around Network here, and, for example, not in network.py ?

jhillairet (commented):
great work, thank you. It's OK to me, only few minor comments and questions

skrf/constants.py
```diff
@@ -60,6 +62,8 @@
 S_DEFINITIONS = ['power', 'pseudo', 'traveling']
 S_DEF_DEFAULT = 'power'
 
+NumberLike = Union[Number, Sequence[Number], npy.ndarray]
+
```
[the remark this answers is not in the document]
FranzForstmayr: Indeed, but ArrayLike contains much more types, like strings- and byte (arrays).

```
(type alias) ArrayLike: Type[int] | Type[float] | Type[complex] | Type[str] | Type[bytes] | Type[generic] | Type[Sequence[int | float | complex | str | bytes | generic]] | Type[Sequence[Sequence]] | Type[_SupportsArray]
```

I choosed Number for every scalar value, complex is also a subtype of Number

skrf/media/media.py
FranzForstmayr: Wasn't the Media class intended to be an abstract class?
There are abstract properties and the `__metaclass__`. 
I just wonder, why python let's me instantiate the previous code when Media is clearly an abstract class

skrf/networkSet.py
FranzForstmayr: Within the `Network` class, `Network` is not (fully) defined yet, so I'll get an error.
One could import `annotations` from `__future__`, which will become the default behaviour in Python 3.10.

As we are still supporting Python3.6 there's no `from __future__ import annotation`. This would allow to use a class name as typehint within the class. For Python < 3.7 one should just use the stringified name.

See here for a better explanation:
https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class

skrf/network.py
FranzForstmayr: You are right of course! :)

skrf/media/media.py
jhillairet: I guess it was indeed intended to be an abstract class (@arsenovic could say more than me.). I think it's more "pythonic" the way you propose, I was just wondering myself

skrf/networkSet.py
jhillairet: ok, thank you for the explanation

skrf/media/media.py
arsenovic: this change makes sense. i do believe the media is an abstract class, not sure why i didnt subclass abc. 
i think the closest thing to instantiate directly is DefinedGammaZ0.

skrf/calibration/calibration.py
```diff
@@ -258,7 +258,7 @@ def __init__(self, measured, ideals, sloppy_input=False,
         # ensure that all ideals have same frequency of the measured
         # if not, then attempt to interpolate
         for k in list(range(len(self.ideals))):
-            if self.ideals[k].frequency != self.measured[0]:
```
arsenovic, line 261: why didnt this  print warnings all over the place before??

skrf/frequency.py
```diff
@@ -78,7 +80,8 @@ class Frequency(object):
             }
 
 
-    def __init__(self,start=0, stop=0, npoints=0, unit='ghz', sweep_type='lin'):
+    def __init__(self, start: float = 0, stop: float = 0, npoints: float = 0,
```
arsenovic, line 83: npoints should be int

skrf/media/media.py
```diff
@@ -142,7 +141,8 @@ def z0(self):
     def z0(self, val):
         self._z0 = val
 
-    @abstractproperty
+    @property
+    @abstractmethod
```
arsenovic, line 145: i am not sure the difference between these two.  are they equivalent?

skrf/util.py
```diff
@@ -39,14 +39,6 @@
 import sys
 from functools import wraps
 
-# globals
-
-try:
```
arsenovic, line 44: can you comment on this removal?

arsenovic:
i made several small comments. this is a great addition and will really help with bug catching and usability .  thanks a lot for adding this.

skrf/calibration/calibration.py
FranzForstmayr: The frequency class has a `f` property, the network class forwards this property of the underlying frequency object.
So comparing `self.f == other.f` returned true, even if the `other` instance was of type `network`

skrf/media/media.py
FranzForstmayr: According to python docs it's equivalent, and recommended since python3.3.
I think there should be another cleanup round to remove lot of python2 related code. I just came over this recently, so I fixed this directly.

skrf/util.py
FranzForstmayr: Basestring was removed with python3, I was not even aware of this before.
basestring is now just written as `str`.

FranzForstmayr:
> i made several small comments. this is a great addition and will really help with bug catching and usability . thanks a lot for adding this.

Thanks :)

jhillairet:
Merged. Thank you @FranzForstmayr

PR #453, opened 2021-04-22
Fix image links in documentation

jhillairet:
Should use the html `` in a notebook instead of the `` markdown, otherwise Sphinx takes only the last image into account. (Another solution is to use `xxx` in markdown, where xxx is different for all images)


jhillairet:
[Documentation ](https://scikit-rf.readthedocs.io/en/latest/) has been updated

thanks to @rragona for pointing me these errors

PR #455, opened 2021-04-24
Z0 - new setter/getter

FranzForstmayr:
While adding some typehints I struggled with the z0 setter/getter methods, so I decided to rewrite them.

I had to modify some tests, as I think they were wrong, because we pass two frequency points but the s-matrix has shape (1,x,x), so only one frequency point.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -765,8 +765,10 @@ class Network(object):
         else:
             name = self.name
 
-        if self.z0.ndim < 2:
-            z0 = self.z0
+        _z0 = self.z0
+
+        if _z0.ndim < 2:
+            z0 = _z0
         else:
             z0 = self.z0[0, :]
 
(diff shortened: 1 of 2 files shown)


skrf/network.py
```diff
@@ -400,6 +400,7 @@ def __init__(self, file=None, name=None, comments=None, f_unit=None, s_def=S_DEF
         self.deembed = None
         self.noise = None
         self.noise_freq = None
+        self._z0: npy.ndarray = npy.array(50, dtype=complex)
```
jhillairet, line 403:
```suggestion
        self._z0 = npy.array(50, dtype=complex)
```

skrf/network.py
```diff
@@ -1109,62 +1112,48 @@ def z0(self):
...
+        if z0.ndim == 1:
+            if len(z0) in (self.frequency.npoints, self.nports):
+                self._z0 = z0
+                return
+        elif z0.ndim == 2:
+            if z0.shape == (self.frequency.npoints, self.nports):
+                self._z0 = z0
+                return
+
+        raise AttributeError('Unable to broadcast z0 to s shape')
```
jhillairet, line 1156: Ok. would be nice if you could add some comments to explain the different cases

arsenovic: i agree that a few comments would go a  long way. looks great though. 
adding tests for the various z0  setters would be great since you have it fresh in mind, but up to you. 

thanks.

skrf/network.py
arsenovic: will this typeing  allow them to set a number  for z0?

skrf/network.py
```diff
@@ -764,12 +765,14 @@ def __str__(self):
         else:
             name = self.name
 
-        if len(npy.shape(self.z0)) == 0 or npy.shape(self.z0)[0] == 0:
-            z0 = str(self.z0)
+        _z0 = self.z0
+
+        if _z0.ndim < 2:
+            z0 = _z0
```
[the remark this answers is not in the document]
arsenovic: agree with jillairet

skrf/network.py
FranzForstmayr: Thanks for the feedback, I added some comments in both setter and gettter.

I will add some tests the next days :)

jhillairet:
Nice addition. I will not be picky asking what happen for a N-port with N frequency points ;). I think it's OK to merge if it's OK for you.

FranzForstmayr:
There's already a comment about this ambiguity:
https://github.com/scikit-rf/scikit-rf/pull/455/files#diff-e8681d39b5ae2d88950aff03c3a1f1ab9db446f7e08d77976a81777e55ad3c52R1155

It's more a theoretical issue, but if nports == npoints, and i pass z0 like [1,2], port1 will have 1, port 2 will have 2 for each frequency.
I'm not even sure, if someone wants to pass a vector with length == len(npoints) without specifying each port separately (As fxn array). However, I decided to handle this case like the previous implementation.
It's up to you if this feature is needed, otherwise I'm fine with merging :)

jhillairet:
merged. Thank you @FranzForstmayr

PR #457, opened 2021-04-28
Allow reading HFSS files from StringIO

FranzForstmayr:
This patch fixes reading hfss files from stringio.

See #456 

The `get_gamma_z0_fid` method is now called before closing the stringIO handle, the values are stored in the Touchstone class.
The original `get_gamma_z0` returns just the previously parsed values.

PR #458, opened 2021-05-01
Switch nbsmoke to nbval for notebook testing

jhillairet:
Since I didn't manage to make nbsmoke working without installing scikit-rf from git. By doing so, changes in PR are not seen during the notebooks evaluation.

Some note concerning the options used:
- `--nbval-lax`, which runs notebooks and checks for errors, but only compares the outputs of cells which include a `#NBVAL_CHECK_OUTPUT` marker comment.
- By default, each .ipynb file will be executed using the kernel specified in its metadata, a behaviour overridden by passing either `--current-env` to use a kernel in the same environment in which pytest itself was launched.
- Specify `-p no:python` if you would like to execute notebooks only. 
- ignore instrumentcontrol notebooks

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -23,7 +23,7 @@ jobs:
         pip install flake8 pytest
         sudo apt-get install xvfb
         pip install coverage
-        pip install nbsmoke
+        pip install nbval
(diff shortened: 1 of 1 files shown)


jhillairet:
one notebook is failing due to the previously merged PR #457. Merging for now, in order to be able to test correctly the notebooks from future PRs

PR #459, opened 2021-05-01
Add tests for NetworkSet, fix NetworkSet.from_zip and other things

jhillairet:
- fix `NetworkSet.from_zip` to be compatible with file streams following PR #457 
- add a complete set of tests for NetworkSet
- add `__eq__` method to NetworkSet to allow comparing two NetworkSet
- add `__len__`method to NetworkSet to get the length more easily
- add `inplace` parameter for the `sort()`method (default being `True `to not break compatibility)
- fix docstring style according to  [pycodestyle](https://pycodestyle.pycqa.org/en/latest/intro.html) (such as ''' -> """)
- complete docstrings when missing
- add type hint when possible in `networkSet.py`
- fix bug with missing extension when exporting Excel from NetworkSet
- fix the missing requirement (openpyxl) for Excel export
- remove the skrf git clone in requirements.txt (was for nbsmoke to work)...
- ...and finally managed to make pytest --nbval works! (that was the hard part).


conda.recipe/meta.yaml
```diff
@@ -26,6 +26,7 @@ requirements:
     - matplotlib
     - six 
     - future
+	- openpyxl
```
jhillairet, line 29:
```suggestion
    - future
    - openpyxl
```

conda.recipe/meta.yaml
```diff
@@ -41,6 +42,7 @@ requirements:
     - xlwt
     - xlrd
     - networkx
+	- openpyxl
```
jhillairet, line 46:
```suggestion
    - networkx
    - openpyxl
    
```

jhillairet:
OK, so in addition to the things concerning NetworkSet in this PR, it also fixes notebook testing. Notebook testing now works without having to clone the master, which led to undetected some new bugs introduced with some previous PRs. 

In fact, it was neither the fault of nbsmoke or nbval, but a strange behaviour or pytest which was not finding the skrf package in github Actions environment, despite skrf being in pythonpath. Despite working fine locally on my computer... the skrf package is now built and install in the github Actions from a push/PR and then it works...

Ready for review!

arsenovic:
that must have been a hard bug to find!  tests look good, and very needed.

PR #460, opened 2021-05-04
VectorFitting: Bugfix, new testcases and improved documentation

Vinc0110:
This PR fixes the generated API docs of the VectorFitting class, which were missing the plotting functions. 

This was probably due to the changes in PR https://github.com/scikit-rf/scikit-rf/pull/444: the wrapper for the plotting functions was masking them, so they were excluded from the doc builds by Sphinx. Calling [functools.wraps()](https://docs.python.org/3/library/functools.html#functools.wraps) in the wrapper seems to do the trick.


jhillairet:
Note that the official documentation page is currently built "manually" (when we trigger it manually), not triggered automatically after each merging (should eventually be done on day).

Vinc0110:
I see. But the latest doc build seems to be only 4 days old and already includes the changes from PRs https://github.com/scikit-rf/scikit-rf/pull/444 and https://github.com/scikit-rf/scikit-rf/pull/450. 
When I build the docs locally the plotting functions are also missing, which is now fixed by this PR. Not sure why I did not spot this problem early though...

Vinc0110:
As this PR is still open, I've pushed some more commits adding 3 more testcases. In the process, I also spotted and fixed a bug in the read/write functions.

I also modified requirements.txt and added the `coverage` module as it is required for coverage reports with `nosetests`.

Vinc0110:
@jhillairet: It looks like you did a great job fixing the CI bug. From my side, this PR is ready to be merged.

requirements.txt
```diff
@@ -13,5 +13,6 @@ networkx
 pathlib
 nose
 nose-exclude
+coverage
 xlwt
 git+https://github.com/scikit-rf/scikit-rf.git@master
```
jhillairet, line 18: this line has been removed in #459. Could you try updating your PR from the current master branch? (I don't understand why github is telling me there is no conflict)

Vinc0110: Sure. I just did that below. Should work now.

jhillairet:
the doc has been rebuilt : https://scikit-rf.readthedocs.io/en/latest/

Vinc0110:
Thanks!
But it seems like there is a big problem in some (all?) example notebooks: `ModuleNotFoundError: No module named 'skrf'`

jhillairet:
> 
> 
> Thanks!
> But it seems like there is a big problem in some (all?) example notebooks: `ModuleNotFoundError: No module named 'skrf'`

ho fantastic. Now that CI finally works, RTD not anymore... 😫  thanks for noticing!

PR #463, opened 2021-05-07
Fix docs

FranzForstmayr:
This shold fix the doc build, as runipy is executed from the root directory and is able to import skrf without installing

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -37,8 +37,8 @@ help:
 all: notebooks  html
 
 notebooks:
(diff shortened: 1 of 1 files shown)


jhillairet:
I've rebuilt the doc, but the problem is still here

FranzForstmayr:
This is strange, as I was able to reproduce the issue first and fixed it with that patch.
I tried it like described in the docs
```
cd docs
make html
```
Without this it requires skrf to be installed in the virtual environment. 
Maybe readthedocs performs different steps by default.

PR #465, opened 2021-05-10
Fix math typesetting in network.chopinhalf

NikoDaGreat:
Currently, the docs for network.chopinhalf look like this
image

This is fixed by removing the second `\\cdot` for the first eq. and adding escaping `\` for the `\frac`s in the latter eqs.


NikoDaGreat:
Hi, you can merge now

PR #469, opened 2021-05-18
Vector fitting: Bugfixes, example notebooks, testcases

Vinc0110:
This PR fixes two bugs in the vector fitting algorithm and adds some documentation and tests.

The first bug was related to the flags `fit_constant` and `fit_proportional` for the fitting function `vector_fit()`, which did not work in some cases. 
The second bug caused a poor convergence or even failure for larger networks, which works much better now. I'm actually surprised the algorithm worked at all with that bug in place, so it's good to have it fixed.

To show off the capabilities of vector fitting, I wrote another example notebook. The 4-port network from `./skrf/tests/Agilent_E5071B.s4p` is copied to the example folder and fitted in the new notebook, which is a good example for a large and noisy network (i.e. a real-world example).

After the bugfixes, some of the old tests failed, so I corrected them and even added another one.

Once this PR is merged, I propose to update the package release and RTD, so the buggy code gets fixed (even though I doubt that the vectorfitting feature has [m]any users).


jhillairet:
Great, thank you. Is it ready to merge for you?

Vinc0110:
Yes, it's ready.

jhillairet:
By the way @Vinc0110 , Read The Docs fails in building the documentation, because there is a cell that takes too much times to execute:
```
nbsphinx.NotebookError: CellTimeoutError in examples/vectorfitting/vectorfitting_Agilent_E5071B.ipynb:
A cell timed out while it was being executed, after 120 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
vf.vector_fit(n_poles_real=1, n_poles_cmplx=26)
-------------------


Notebook error:
CellTimeoutError in examples/vectorfitting/vectorfitting_Agilent_E5071B.ipynb:
A cell timed out while it was being executed, after 120 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
vf.vector_fit(n_poles_real=1, n_poles_cmplx=26)
```

PR #472, opened 2021-05-21
Fix port numbering, start with Port[1]

luar123:
There is a bug in write_touchstone when the port names are written: The first port should be Port[1] instead of Port[0].

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1968,7 +1968,7 @@ class Network(object):
                 if self.port_names and len(self.port_names) == self.number_of_ports:
                     ports = ''
                     for port_idx, port_name in enumerate(self.port_names):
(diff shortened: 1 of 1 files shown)


jhillairet:
OK, thank you. 

I didn't check: is reading is OK?

luar123:
Reading back is working fine:
```
freq = rf.Frequency(1, 10, 101, 'ghz')

s = rand(101, 2, 2) + 1j*rand(101, 2, 2)

ntw2 = rf.Network(frequency=freq, s=s, z0=50, name='same z0 for all ports')

ntw2.write_touchstone()

ntw3=rf.Network('same z0 for all ports.s2p')

ntw2 == ntw3
Out[36]: True

ntw2.port_names == ntw3.port_names
Out[37]: True
```

PR #473, opened 2021-05-24
Vector Fitting: Fixing RTD build timeout

Vinc0110:
This PR hopefully fixes the RTD build timeout of the new vector fitting example notebook, reported in https://github.com/scikit-rf/scikit-rf/pull/469#issuecomment-846389757.

Hesitating to disable the execution timeout entirely, I increased the maximum from 120 s to 240 s (only for this notebook).


jhillairet:
OK, it works now:

https://scikit-rf.readthedocs.io/en/latest/examples/vectorfitting/vectorfitting_Agilent_E5071B.html

PR #474, opened 2021-05-28
Remove pandas dependency

FranzForstmayr:
I added a pandas dependency with the last PR #451 . 
This commit should remove the dependency.

I did not recognize this earlier, as I had pandas installed almost always :)
Sorry for the inconvenience

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ qtapps/sandbox*
 *.iml
 
 .vscode/
+.coverage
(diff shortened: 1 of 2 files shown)


FranzForstmayr:
We should add a testcase only with the minimum required subset of packages installed. (without pandas and matplotlib)
However, I would recommend to use a testing tool like tox (or even better nox) to manage the versions.

FranzForstmayr:
> I agree concerning dependencies and nox. The dependencies in the conda config yml file should also be updated.
> 
> By the way, there is a couple of binary files associated to the PR. Not sure they are useful, isn't ?

Thanks for the hint :)
Fixed!

Issue #196, opened 2018-02-12
Suggestion: Implement De-Embedding class


climbit:
Hi all,

is there intrest in implementing a class for de-embedding?

Cheers,
Daniel

denzchoe:
I've interest to attend it too :)

arsenovic:
i am not sure what you are envisioning. could you describe what you are
thinking, and  show some examples of how it would be used?

climbit:
The class should implement functions for different de-embeding methods like open pad de-embedding (opd), open short de-embedding (osd), 2 port de-embedding (2PD)

arsenovic:
ok, i am not familiar with those features. it sounds like partial one-port calibration tailored to on-wafer measurements?
you might be able to re-use some of the Calibration code, (like subclass Calibration), but if its easier for you to start fresh feel free to create a new file. you could put it in skrk/calibration/   but we can decide later.

climbit:
It is not necessarily tailored to on wafer measurements. For calibration you need very perfect standards because you assume a perfect short, open.... right? But usually you can't produce that perfect standards yourself. Therefore you need de-embedding, where you think of the parasitics as a network with different elements. You try to measure them and remove them from the dut measurements.

I open a new branch 'deembedding' in my fork and work within this branch. Should I push the progress from time to time or should I first push if I am done?

arsenovic:
for generic   calibration you can use any  standards, so long as you know what they are.  if you can model parasitics then you can use 'non-perfect' standards with normal SOL calibration.  alternatively you can use some self-calibration techniques. 

do you have an references or websites describing what you are disusing?

climbit:
> ... alternatively you can use some self-calibration techniques.

By self-calibration, you mean internal calibration of the equipment?

> for generic calibration you can use any standards, so long as you know what they are. if you can model parasitics then you can use 'non-perfect' standards with normal SOL calibration.

How do you know how a standard look like, don't you need to measure it?
How would the procedure look like with non-perfekt standards?
It would be great if you can show me some references. That might be very helpful for me.

Regarding what I said about calibration and perfect standards, I reference to [1], 2nd page, last paragraph.
[2] describes the methods I would like to implement.

[1] http://literature.cdn.keysight.com/litweb/pdf/5980-2784EN.pdf
[2] https://www.researchgate.net/publication/2517390_On-Wafer_Calibration_Techniques_for_Giga-Hertz_CMOS_Measurements

[...]

PR #480, opened 2021-06-08
WIP: Implement a de-embedding class

vik-s:
Hey,

I've got some preliminary code to implement a de-embedding class which was suggested in #196. I figured it was easiest to start a deembedding.py file in skrf/calibration and put in a base deembedding class, based on which other deembedding methods can be implemented. I have started with OpenShort deembedding, since that is commonly used in on-wafer measurements.

Here are some questions to get discussions started:

1. Is this a good place to develop this capability?
2. How should we go about naming the different deembedding methods to easily help the user identify it?
3. What are some of the deembedding methods that should be implemented that are most useful?

Feedback is welcome!

### DIFF
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -1,0 +1,80 @@
+from ..frequency import *
+from ..network import *
+
+class Deembedding(object):
+    '''
+    Base Class for all de-embedding objects
+    '''
+    def __init__(self, dummy, name=None, *args, **kwargs):
+        '''
+        De-embedding Initializer
+
+        Notes
+        -----
+        Parameters
+        ----------
+        dummies: network objects of dummy structures used for de-embedding 
+        '''
+
+        # fill dummy with copied lists of input
+        self.dummy = [ntwk.copy() for ntwk in dummy]
+
+       # ensure all the dummy Networks' frequency's are the same
+        for dmyntwk in self.dummy:
+            if self.dummy[0].frequency != dmyntwk.frequency:
+                raise(ValueError('Dummy Networks dont have matching frequencies.')) 
+
+        # may attempt to interpolate if frequencies do not match
+
+        self.kwargs = kwargs
+        self.name = name
+
+    def __str__():
+        pass
+
+    def __repr_():
+        pass
+
+    def apply_cal(self,ntwk):
+        '''
+        Apply correction to a Network
+        '''
+        raise NotImplementedError('The Subclass must implement this')
+
+class OpenShort(Deembedding):
+    '''
+    2-step open-short de-embedding [1]_
+
+    [1] M. C. A. M. Koolen, J. A. M. Geelen and M. P. J. G. Versleijen, "An improved 
+    de-embedding technique for on-wafer high frequency characterization", 
+    IEEE 1991 Bipolar Circuits and Technology Meeting, pp. 188-191, Sep. 1991.
+    '''
+    def __init__(self, measured, *args, **kwargs):
+        '''
+        Docstring
+        '''
+        Deembedding.__init__(self, measured, *args, *kwargs)
+
+    def apply_cal(self, ntwk):
+        '''
+        Docstring
+        '''
+       
+        # first measured ntwk is open dummy
+        open = self.dummy[0].copy()
+
+        # second measured ntwk is short dummy
+        short = self.dummy[1].copy()
+        
+        # check if the frequencies match with dummy frequencies
+        if ntwk.frequency != open.frequency:
+            raise(ValueError('Network frequencies dont match dummy frequencies.')) 
+
+        caled = ntwk.copy()
+
+        # remove open parasitics
+        caled.y = ntwk.y - open.y
+        # remove short parasitics
+        caled.z = caled.z - short.z
+
+        return caled
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from ..frequency import *
 from ..network import *
 
@@ -34,7 +35,8 @@ class Deembedding(object):
 
     def __repr_():
         pass
-
+    
+    @abstractmethod
     def apply_cal(self,ntwk):
         '''
         Apply correction to a Network
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -1,29 +1,33 @@
 '''
-.. module:: skrf:calibration:deembedding
+.. module:: skrf.calibration.deembedding
 
 ====================================================
 deembedding (:mod:`skrf.calibration.deembedding`)
 ====================================================
 
 This module provides objects to implement de-embedding methods
-for on-wafer applications. Each de-embedding method inherits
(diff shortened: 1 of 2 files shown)


skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,80 @@
...
+    '''
+    def __init__(self, dummy, name=None, *args, **kwargs):
+        '''
+        De-embedding Initializer
+
+        Notes
+        -----
+        Parameters
+        ----------
+        dummies: network objects of dummy structures used for de-embedding
```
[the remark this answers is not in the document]
vik-s: It could be just a single dummy network object for just an open de-embed. I'll change it up to reflect that.

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,80 @@
+from ..frequency import *
+from ..network import *
+
+class Deembedding(object):
```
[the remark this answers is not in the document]
vik-s: I did try subclassing `calibration` but it has `ideals` as a compulsory input. De-embedding only requires a bunch of `network` objects that will be removed from the raw device measurement. So I decided to not subclass `calibration` and went with a separate class for de-embedding. What are your thoughts on this?

Great point on using Abstract classes. I will use that for the base class because it will never be instantiated directly.

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,80 @@
+from ..frequency import *
+from ..network import *
+
```
jhillairet, line 3: tests seems to fail because of missing  `abstractmethod`
```suggestion
from abc import ABC, abstractmethod
```

JAnderson419:
Hello Vikram, 

Nice work on implementing a native de-embedding class. One comment I have (that is personal preference/usability rather than a functional issue) is that it might be nice to have open and short as named parameters for the OpenShort method rather than relying on the user to put them in the correct position in a list of networks. I know you are using dummy to check that the frequencies all line up - perhaps these named networks could be loaded into dummy in the __init__ statement of OpenShort and then an internal method of Deembedding run to check the frequencies?

Cheers,
Jackson

skrf/calibration/deembedding.py
vik-s: I'm adopting `ABC` and `abstractmethod` into this de-embedding class. So hopefully it goes away.

vik-s:
Hi Jackson,

Thanks for the feedback. I agree it will be much more clear if I use named parameters for open/short networks. I will implement this in my next commit. I was also thinking that I should allow both `Network` objects and filename strings as valid inputs for de-embedding.

vik-s:
I thought I will provide a quick update on this. I've implemented the basic classes and methods needed for some standard de-embedding styles. I tried building the sphinx documentation locally, and it seems okay. I'm not too proficient in writing unit tests, so it might take me a bit to figure that out and write it in for this class.

My question is should I make this PR anyway, and write them as another PR? Or is it preferred to do both together?

vik-s:
@jhillairet @JAnderson419 would you folks have a moment to look at this PR? I think it's ready for review. Tests for this class coming later :)

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,452 @@
...
+        self.args = args
+        self.kwargs = kwargs
+        self.name = name
+
+    def __str__():
+        pass
+
+    def __repr_():
+        pass
```
FranzForstmayr, line 91: These two methods are required to have 'self' as first argument. str(instance) throws an error now.
However, as they are empty anyway I'd suggest to remove them completely.
If you want Child classes to override these methods, you should make them abstract.

vik-s: Thanks for pointing that out. I forgot to update those methods, I'll push an update shortly.

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,452 @@
...
+.. module:: skrf.calibration.deembedding
+
+====================================================
+deembedding (:mod:`skrf.calibration.deembedding`)
+====================================================
+
+This module provides objects to implement de-embedding methods
+that are often used for on-wafer applications. 
+Each de-embedding method inherits from the common abstract base 
+class :class:`Deembedding`.
```
jhillairet, line 11: It would be worth mentioning in the documentation somewhere why these objects differ from the calibration ones.

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,452 @@
...
+    '''
+    def __init__(self, dummies, name=None, *args, **kwargs):
+        '''
+        De-embedding Initializer
+
+        Notes
+        -----
+        Each de-embedding algorithm may use a different number of
+        dummy networks. We check that each of these dummy networks
+        have matching frequecies to perform de-embedding.
```
jhillairet, line 55: typo:
```suggestion
        have matching frequencies to perform de-embedding.
```

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,452 @@
...
+    Remove parasitics to get the actual device network
+        
+    >>> realdut = dm.deembed(dut)
+
+    References
+    ------------
+
+    .. [1] M. C. A. M. Koolen, J. A. M. Geelen and M. P. J. G. Versleijen, "An improved 
+        de-embedding technique for on-wafer high frequency characterization", 
+        IEEE 1991 Bipolar Circuits and Technology Meeting, pp. 188-191, Sep. 1991.
```
jhillairet, line 141: Detail: Generally, the References section should be before the Example section

skrf/calibration/deembedding.py
```diff
@@ -0,0 +1,452 @@
...
+    def deembed(self, ntwk):
+        '''
+        Perform the de-embedding calculation
+
+        Parameters
+        ----------
+        ntwk : :class:`~skrf.network.Network` object
+            Network data of device measurement from which
+            parasitics needs to be removed via de-embedding
+        '''
```
jhillairet, line 182: Could you add in the docstring the Return section with the description of `caled`? Same for the other methods `deembed`

jhillairet (commented):
for my knowledge: is there a big change between `OpenShort` and `ShortOpen` ?

vik-s:
> for my knowledge: is there a big change between `OpenShort` and `ShortOpen` ?

Yes, actually. In `OpenShort`, you remove shunt parasitics before series ones. In `ShortOpen`, it is the opposite- series before shunt. It all depends on the nature of the test fixture in which the device is being tested in. For example, is there series resistance to be removed before shunt capacitors, or is it the other way around.

Thank you for reviewing the code. I'll implement these changes shortly.

jhillairet:
OK for me. Any other things to add @vik-s  ?

jhillairet:
One dedicated notebook example would be very useful (in this PR, or in a next one if you prefer). 

If you can, with pictures (I'm very interesting into seeing more practical setups and photos in the documentation).

vik-s:
I'm ok to merge this for now. I'll write tests for this class and notebook examples in another PR.

jhillairet:
Merged. @vik-s it would be great if you could make a notebook example and write some tests. I would like to wait for them before making a new release.

PR #481, opened 2021-06-15
Modify widgets.py to accept .S*P files

wbarnha:
Some manufacturers provide their .s2p files with an extension as .S2P, and I got tired of changing them all the time. This should allow data_grabber.py to include the .S*P extensions as well for loading files.


jhillairet:
Sure, thank you.

PR #482, opened 2021-06-22
Add analyzer_keysight_n9918a.py for usage in data_grabber.py

wbarnha:
I added a file so that a Keysight FieldFox could be used in conjunction with data_grabber.py. However, I have not yet been able to test it since my development environment (Ubuntu 18.04) is not directly compatible with NI VISA.

I wonder if there's any incentive to move away from using NI VISA to communicate with these devices. I have been able to communicate with these Keysight devices without needing the VISA binaries setup on my computer. So perhaps we could make a fallback case for when a user doesn't have the proper VISA libraries installed?

I know it's possible, according to https://lightlab.readthedocs.io/en/master/_static/installation/sysadmin.html#ubuntu-installation, but do we expect users to go through this?

### DIFF
--- a/qtapps/skrf_qtwidgets/analyzers/analyzer_keysight_n9918a.py
+++ b/qtapps/skrf_qtwidgets/analyzers/analyzer_keysight_n9918a.py
@@ -1,0 +1,9 @@
+from skrf.vi.vna import keysight_fieldfox
+
+
+class Analyzer(keysight_fieldfox.FieldFox):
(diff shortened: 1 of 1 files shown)


jhillairet:
OK, thank you. I'll merge.

jhillairet:
> I wonder if there's any incentive to move away from using NI VISA to communicate with these devices. I have been able to communicate with these Keysight devices without needing the VISA binaries setup on my computer. So perhaps we could make a fallback case for when a user doesn't have the proper VISA libraries installed?

Eventually, to be discussed. You can open an issue to gather comments from other on this.

PR #484, opened 2021-06-24
Fix a lot of typos

nikosavola:
This PR fixes a lot of typos, mostly in comments, some in function and variable names.

The typos were scoped automatically with [scspell](https://pypi.org/project/scspell/)


skrf/media/media.py
```diff
@@ -950,7 +950,7 @@ def shunt_capacitor(self,C,*args,**kwargs):
         Returns
         --------
         shunt_capacitor : :class:`~skrf.network.Network` object
-                shunted capcitor(2-port)
+                shunted capaciton(2-port)
```
jhillairet, line 953:
```suggestion
                shunted capacitor (2-port)
```

jhillairet (approved):
Very nice, thank you. I've only spotted 1 thing. 

And the winner is... "characteristic" ;)

Vinc0110 (approved):
Wow, that's a lot of typos... Thanks for fixing them.

Just an idea for further work on this:
It would probably be even more important and useful to do this kind of check on the Jupyter notebooks for the examples, tutorials and API documentation in `scikit-rf/doc/source/`. That should be possible, right?

nikosavola:
> Wow, that's a lot of typos... Thanks for fixing them.
> 
> Just an idea for further work on this:
> It would probably be even more important and useful to do this kind of check on the Jupyter notebooks for the examples, tutorials and API documentation in `scikit-rf/doc/source/`. That should be possible, right?

Yes, should be similarly possible. I might even do that at some point.

jhillairet:
> Yes, should be similarly possible. I might even do that at some point.

OK. I will merge for now. Thank you !

PR #485, opened 2021-06-25
Add stdev function, organize file, allow read_all() to recursively check directories

wbarnha:
I created a function in network.py called stdev(), which was inspired by the adjacent average() function.

However, after realizing that networkSet.py exists, we may not need this.


wbarnha:
I think something that's worth noting is that this library has a lot more functionality than I realized. I spent a long time manually trying to check what's going on with calls to network.s vs network.s_sb for deriving standard deviations and mean values of network sets when i didn't realize uncertainty_ntwk_triplet was a thing. Probably because my IDE didn't have the means to make that immediately clear to me and I didn't RTFM. 

Maybe we should change things so network.py doesn't have average() and stdev(). Instead we could defer to using the methods in networkSet.

jhillairet:
> Maybe we should change things so network.py doesn't have average() and stdev(). Instead we could defer to using the methods in networkSet.

Yes, I do agree. 

In fact, I was about to ask if it wouldn't be redundant with the `mean_s_db` and `std_s_db` provided with `NetworkSet`, but then you made me realize the existence of the function `average` ;-)

In any case, the recursive search is a good idea.

wbarnha:
Another issue I encountered was after using the uncertainty_ntwk_triplet function, there was a mistake in its implementation:
```
        upper_bound = (ntwk_mean + ntwk_std)
        lower_bound = (ntwk_mean - ntwk_std)
```
This got me unusual results where the upper and lower bounds would occasionally intersect with each other. I just tested this using generic s-parameters from a manufacturer for the same device with repeated tests. Changing this to:
```
        upper_bound = ntwk_mean.copy()
        lower_bound = ntwk_mean.copy()
        upper_bound.s = (ntwk_mean.s + ntwk_std.s)
        lower_bound.s = (ntwk_mean.s - ntwk_std.s)
 ```
 got me proper results where the upper and lower bounds are properly defined. We may also want to hardcode it to only use 's' as an attribute.

**Comment by wbarnha** [comment]
> In fact, I was about to ask if it wouldn't be redundant with the `mean_s_db` and `std_s_db` provided with `NetworkSet`, but then you made me realize the existence of the function `average` ;-)

Funny enough, uncertainty_ntwk_triplet does everything I've been looking for, in a nice one-liner. I really would like to really give this function more attention.

> In any case, the recursive search is a good idea (but it currently fails on notebook examples).

I'd like to fix this before doing a merge, then. It's certainly a feature I need.

**Comment by wbarnha** [comment]
I forgot that we shouldn't use lists as default kwargs otherwise we run into the issue of scope inside function calls. Oops.

**Comment by wbarnha** [comment]
Who knew changing read_all would require this much effort, good test cases. Feel free to merge, I have nothing further to commit for now.

File: skrf/io/general.py [hunk 1]
```diff
@@ -242,11 +243,13 @@ def write(file, obj, overwrite = True):
         pickle.dump(obj, fid, protocol=2)
         fid.close()
 
-def read_all(dir='.', contains = None, f_unit = None, obj_type=None, files=None):
+def read_all(dir: str ='.', contains = None, f_unit = None, obj_type=None, files: list=None, recursive=False) -> dict:
```
jhillairet, line 246: could you add in the docstring the optional parameters `recursive`?

skrf/io/general.py
```diff
@@ -300,19 +303,27 @@ def read_all(dir='.', contains = None, f_unit = None, obj_type=None, files=None)
 
     out={}
 
-    filelist = files
-    if files == None:
-        filelist = sorted(os.listdir(dir))
+    if not os.path.isdir(dir):
+        dir = os.path.join(os.getcwd(), dir)
```
jhillairet, line 307: if `dir` is not a directory, why `join(os.getcwd(), dir)` would be a directory? Do you have an example ?

skrf/networkSet.py
```diff
@@ -800,8 +800,10 @@ def uncertainty_ntwk_triplet(self, attribute: str, n_deviations: int = 3):
         ntwk_std = self.__getattribute__('std_'+attribute)
         ntwk_std.s = n_deviations * ntwk_std.s
 
-        upper_bound = (ntwk_mean + ntwk_std)
-        lower_bound = (ntwk_mean - ntwk_std)
+        upper_bound = ntwk_mean.copy()
+        lower_bound = ntwk_mean.copy()
+        upper_bound.s = (ntwk_mean.s + ntwk_std.s)
```
jhillairet, line 805: I don't understand why you see a difference with this and the previous implementation. Adding two Networks do:

https://github.com/scikit-rf/scikit-rf/blob/d1f60e701cac7276784510a4aaba74a68cc0f0a5/skrf/network.py#L581-L594

which looks like similar to what you propose. Do have some examples ?

skrf/network.py
```diff
@@ -4052,6 +4052,34 @@ def average(list_of_networks: Sequence[Network], polar: bool = False) -> Network
     return out_ntwk
 
 
+def stdev(list_of_networks: Sequence[Network], attr: str = 's_db') -> npy.ndarray:
```
FranzForstmayr, line 4055: Is it a good idea, to let users calculate statistics on logarithmic values like `s_db`?
mean(s_db) differs from 20 * log10(mean(s)).

wbarnha: It's not, that was something I forgot to change. Thank you for pointing this out. I think defaulting to 's' instead of 's_db' is a safer practice for now.

skrf/io/general.py
wbarnha: I originally had a reason for this, but I cannot remember. After further testing, it seems unnecessary to be there.

wbarnha: This must have been something I did in my test configuration where I forgot to correct the current path.

skrf/networkSet.py
wbarnha: Apparently I was looking at a bad set of data. I'll find something better to present.

wbarnha: I think the solution here is to create a method to optionally extract the minimums and maximums from the upper_bound and lower_bound network objects. This could create two new objects containing the minimum dB values and maximum dB values for each frequency.

wbarnha: Apparently the issue was that I was calling `uncertainty_ntwk_triplet('s')` instead of `uncertainty_ntwk_triplet('s_mag')`. I'm all set.

wbarnha:
After further tests, it turns out I was using the library incorrectly. However, this may useful to note in the future to prevent future mistakes from other users. The method of calling `uncertainty_ntwk_triplet` is a bit unclear.

Also I'm not sure why that last PR build failed.

jhillairet:
> After further tests, it turns out I was using the library incorrectly. However, this may useful to note in the future to prevent future mistakes from other users. The method of calling `uncertainty_ntwk_triplet` is a bit unclear.

OK. Feel free to improve the documentation if you think you can make it more clear ;)

So, does it change something to add the `.s` below :
```python
        upper_bound = (ntwk_mean + ntwk_std.s)
        lower_bound = (ntwk_mean - ntwk_std.s)
``` 

Note the order of the mean/std operators with respoect to dB operator is discussed in the docstring here:
https://github.com/scikit-rf/scikit-rf/blob/d1f60e701cac7276784510a4aaba74a68cc0f0a5/skrf/networkSet.py#L661-L678

wbarnha:
> Note the order of the mean/std operators with respect to dB operator is discussed in the docstring here:

Learned that one the hard way. I was staring at it but wondering what gets called. :frowning_face: 
Anyway, I followed your request and left a note on uncertainty_ntwk_triplet so that the next person doesn't encounter the confusion I did.

skrf/network.py
```diff
@@ -166,7 +166,7 @@
 
 import numpy as npy
 from numpy.linalg import inv as npy_inv
-from numpy import fft, gradient, ndarray, reshape, shape, ones
+from numpy import fft, gradient, ndarray, reshape, shape, ones, std, dstack
```
jhillairet, line 169: `dstack` seems not use after. 

```suggestion
from numpy import fft, gradient, ndarray, reshape, shape, ones, std
```

skrf/networkSet.py
```diff
@@ -794,14 +797,15 @@ def uncertainty_ntwk_triplet(self, attribute: str, n_deviations: int = 3):
...
 
         """
         ntwk_mean = self.__getattribute__('mean_'+attribute)
         ntwk_std = self.__getattribute__('std_'+attribute)
         ntwk_std.s = n_deviations * ntwk_std.s
 
-        upper_bound = (ntwk_mean + ntwk_std)
-        lower_bound = (ntwk_mean - ntwk_std)
+        upper_bound = (ntwk_mean + ntwk_std.s)
+        lower_bound = (ntwk_mean - ntwk_std.s)
```
jhillairet, line 808: I've tested it and found it was the same as the previous implementation. I'm interested if you have some counter-example.
```suggestion
        upper_bound = (ntwk_mean + ntwk_std)
        lower_bound = (ntwk_mean - ntwk_std)
```

wbarnha: Committed

wbarnha:
Added an additional commit so I could import ro,1.s1p and its files and recreate it in the LaTeX code.

jhillairet:
> Added an additional commit so I could import ro,1.s1p and its files and recreate it in the LaTeX code. 

OK 

> Sorry for the inconvenient reviews.

No problem, everything's fine and thank you for your contribution. OK for merging ?

wbarnha:
You're good to merge now, cheers!

Issue #299, opened 2019-10-29
Offset in `time` vector returned by `Network.impulse_response`


mbastani:
I think there is a one-sample offset in the time vector returned in [`Network.impulse_response` ](https://github.com/scikit-rf/scikit-rf/blob/bba57718e898be96ecbd316ee5f7c1584671ebff/skrf/network.py#L2901) function.

Suppose for simplicity that the function is invoked with `n` satisfying `n % 2 == 0`. [`mf.irfft()`](https://github.com/scikit-rf/scikit-rf/blob/bba57718e898be96ecbd316ee5f7c1584671ebff/skrf/mathFunctions.py#L692) first invokes `numpy.fft.irfft()` that returns the samples of impulse response in the following order: h[0], h[1], ..., h[n-1].  Next an `fftshift` is performed that reorders the output samples to h[n/2], h[n/2+1], ..., h[0], h[1], ..., h[n/2-1].  Because the impulse response is assumed to be periodic with period n, the latter is equivalent to h[-n/2], h[-n/2+1], ... h[n/2-1].   Therefore, the `t` vector created [here](https://github.com/scikit-rf/scikit-rf/blob/bba57718e898be96ecbd316ee5f7c1584671ebff/skrf/network.py#L2951)

            t = npy.flipud(npy.linspace(.5 / fstep, -.5 / fstep, n, endpoint=False))

should actually include `-.5/fstep` point but exclude `.5/fstep`.

FranzForstmayr:
I'm pretty sure that @mbastani  is right here.
Just double checked with an optimum transmission line which should result in an perfect dirac impulse respnse.

```python
import skrf as rf
import numpy as np

freq = rf.Frequency.from_f(np.arange(11), unit='Hz')
net = rf.Network(frequency=freq, s=np.ones(11))

t_vec, y_vec = net.impulse_response('boxcar', pad=0)

for t, y in zip(t_vec, y_vec):
    print(t, y)
```
Results in
```python
-0.45000000000000007 0.0
-0.4 0.0
-0.3500000000000001 0.0
-0.30000000000000004 0.0
-0.25 0.0
-0.20000000000000007 0.0
-0.15000000000000002 0.0
-0.10000000000000009 0.0
-0.050000000000000044 0.0
0.0 0.0
0.04999999999999999 1.0
0.09999999999999998 0.0
0.14999999999999997 0.0
0.19999999999999996 0.0
0.25 0.0
0.3 0.0
[...]

PR #488, opened 2021-06-30
Fix/299

FranzForstmayr:
This PR addresses #299 to avoid a time lag in impulse/step response calculation,

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3201,7 +3201,7 @@ class Network(object):
             w = self.windowed(window=window, normalize=False, center_to_dc=center_to_dc)
         else:
             w = self
-        return t, mf.irfft(w.s, n=n).flatten()
+        return t, mf.irfft(w.s, n=n).ravel()
 
     def step_response(self, window: str = 'hamming', n: int = None, pad: int = 1000) -> Tuple[npy.ndarray, npy.ndarray]:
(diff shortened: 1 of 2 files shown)


jhillairet:
I suggest adding the kind of dedicated test in `test_network.py` (after `test_time_transform` by example). Feel free to improve it. This test does not pass currently, but it passes with this PR.

```python
    def test_impulse_response_dirac(self):
        """
        Test if the impulse response of a perfect transmission line is pure Dirac
        """
        # test for an even and an odd number of frequency points
        n_points = [10, 11]
        for n_point in n_points:
            freq = rf.Frequency.from_f(npy.arange(n_point), unit='Hz')
            # perfect line with s=1 for all frequencies
            net = rf.Network(frequency=freq, s=npy.ones(n_point))
            # check the value at t=0 is 1 and 0 elsewhere
            t, y = net.impulse_response('boxcar', pad=0)
            npy.testing.assert_equal(y[npy.argmin(npy.abs(t))], 1)
            npy.testing.assert_equal(y[:npy.argmin(npy.abs(t))], 0)
            npy.testing.assert_equal(y[npy.argmin(npy.abs(t))+1:], 0)
```

FranzForstmayr:
Ready when you are!

jhillairet:
Great. Thank you

PR #489, opened 2021-07-03
Add functionality to plot group delay for multiple files simultaneously

wbarnha:
When I added group delay plotting for data_grabber.py, I was focused on only making it work for an individual file. I'd like to extend this so that it can plot multiple files simultaneously. This recent commit allows for me to do it, however, I'd like to rework on cleaning up the code in networkPlotWidget since there seems to be some redundant calls.


wbarnha:
I'm satisfied with the changes up to this point, feel free to merge when you get the opportunity.

wbarnha:
Thanks to the recent clean up, it's pretty easy to add new features to data_grabber now, so I went ahead and added VSWR plotting as a feature. I think a nice future feature would be dynamic plotting so that way we don't need to manually add features.

jhillairet:
Do you want to make more modifications or is it ready for merging?

wbarnha:
It's all set for merging!

PR #490, opened 2021-07-05
Documentation Improvements

jhillairet:
- Added a part in the tutorial on how to find min/max and associated frequency
- Fix documentation for the `time` functions
- Fix Slack invitation link on the main page

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ conda install -c conda-forge scikit-rf
 ## Questions or comments? Ask the Community
 
 [![Mailing List](https://img.shields.io/badge/mail-Mailing%20list-brightgreen)](http://groups.google.com/group/scikit-rf)
-[![Slack](https://img.shields.io/badge/Slack-join%20chat-blueviolet?style=flat&logo=slack)](https://scikit-rf.slack.com)
+[![Slack](https://img.shields.io/badge/Slack-join%20chat-blueviolet?style=flat&logo=slack)](https://join.slack.com/t/scikit-rf/shared_invite/zt-d82b62wg-0bdSJjZVhHBKf6687V80Jg)
 [![Matrix](https://img.shields.io/badge/Matrix-join%20chat-blueviolet?style=flat&logo=matrix)](https://app.element.io/#/room/#scikit-rf:matrix.org)
 
 * [Mailing list](http://groups.google.com/group/scikit-rf): You can suscribe to our mailing list to ask (and answer) questions and to receive scikit-rf announcements.
-* Chat: scikit-rf have channels on both [Slack](https://scikit-rf.slack.com) and [Matrix](https://app.element.io/#/room/#scikit-rf:matrix.org)  
+* Chat: scikit-rf have channels on both [Slack](https://scikit-rf.slack.com) (invitation link is above) and [Matrix](https://app.element.io/#/room/#scikit-rf:matrix.org)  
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -35,11 +35,11 @@ conda install -c conda-forge scikit-rf
 ## Questions or comments? Ask the Community
 
 [![Mailing List](https://img.shields.io/badge/mail-Mailing%20list-brightgreen)](http://groups.google.com/group/scikit-rf)
-[![Slack](https://img.shields.io/badge/Slack-join%20chat-blueviolet?style=flat&logo=slack)](https://scikit-rf.slack.com)
+[![Slack](https://img.shields.io/badge/Slack-join%20chat-blueviolet?style=flat&logo=slack)](https://join.slack.com/t/scikit-rf/shared_invite/zt-d82b62wg-0bdSJjZVhHBKf6687V80Jg)
 [![Matrix](https://img.shields.io/badge/Matrix-join%20chat-blueviolet?style=flat&logo=matrix)](https://app.element.io/#/room/#scikit-rf:matrix.org)
 
 * [Mailing list](http://groups.google.com/group/scikit-rf): You can suscribe to our mailing list to ask (and answer) questions and to receive scikit-rf announcements.
-* Chat: scikit-rf have channels on both [Slack](https://scikit-rf.slack.com) and [Matrix](https://app.element.io/#/room/#scikit-rf:matrix.org)  
+* Chat: scikit-rf have channels on both [Slack](https://scikit-rf.slack.com) (invitation link is above) and [Matrix](https://app.element.io/#/room/#scikit-rf:matrix.org)
```
Vinc0110, line 42: Why not directly link to the correct url with the invitation link?

doc/source/conf.py
```diff
@@ -62,7 +62,7 @@
 nbsphinx_allow_errors = True
 nbsphinx_kernel_name = 'python'
 numpydoc_show_class_members = False
-nbsphinx_timeout = 120
+nbsphinx_timeout = 360
```
Vinc0110, line 65: Is this required? The difference is not huge, but in my opinion it's better to leave a smaller timeout in the general configuration and only change the configuration of individual notebooks if they need to run longer (like I had to do for one of the vector fitting examples).

Vinc0110: This is the notebook I meant: https://raw.githubusercontent.com/scikit-rf/scikit-rf/master/doc/source/examples/vectorfitting/vectorfitting_Agilent_E5071B.ipynb
 At the very bottom, the timeout for this notebook is set to 240 seconds:
`"nbsphinx": {
   "timeout": 240
  }`

jhillairet: You're right, it would be better to define this timeout for a single notebook. I'll do it.

README.md
jhillairet: yes, that's right.

doc/source/conf.py
Vinc0110: Is it actually the same vector fitting notebook that's causing a timeout error again? Because it has worked before when it got merged with the 240 s timeout.

jhillairet: > 
> 
> Is it actually the same vector fitting notebook that's causing a timeout error again? Because it has worked before when it got merged with the 240 s timeout.

yes, it's the same:
```
vf.vector_fit(n_poles_real=1, n_poles_cmplx=26)
```
my machine is maybe is bit slower than yours ;)

Vinc0110 (approved):
Thanks for fixing the notebook timeout and the link.

jhillairet:
> Thanks for fixing the notebook timeout and the link.

thanks for the review!

PR #494, opened 2021-07-15
Update & Streamline requirements

JAnderson419:
This is to update the setup.py (closes #493) and also makes an effort to move some of the development/build dependencies out of requirements.txt and into optional settings in setup.py. Github actions may need updating to grab the optional dependencies from setup.py for tests to run.

PR #501, opened 2021-07-23
Fixes ModuleNotFound error in docs

JAnderson419:
skrf install in RTD environment was broken in #494. This edits the rtd yml file to reintroduce the installation of skrf and fix that oversight.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -8,6 +8,7 @@ build:
 
 python:
   version: 3.7
+  pip_install: true
   install:
     - requirements: requirements.txt
     - requirements: requirements_dev.txt
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -8,8 +8,9 @@ build:
 
 python:
   version: 3.7
-  pip_install: true
   install:
     - requirements: requirements.txt
     - requirements: requirements_dev.txt
+    - method: pip
+      path: ./skrf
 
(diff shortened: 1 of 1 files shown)


jhillairet:
> 
> 
> skrf install in RTD environment was broken in #494. This edits the rtd yml file to reintroduce the installation of skrf and fix that oversight.

What error did you see? [I've compiled the documentation yesterday](https://readthedocs.org/projects/scikit-rf/builds/14306784/) and it worked.

JAnderson419:
I missed this as well because it looked like the docs build succeeded, but if you look in the ipynb you will see the below error. I added "pip install ." for the repo so that it installed skrf as well as the requirements, such that it can be imported during doc build. It would be good to have the doc build fail if an error is thrown during notebook execution. Perhaps this could be done through further updates to the rtd yml?

image

jhillairet:
> I missed this as well because it looked like the docs build succeeded, but if you look in the ipynb you will see the below error. 

ho damn, I missed it (again) too. I found that the build process was quicker than usual, but didn't investigate ;)

> It would be good to have the doc build fail if an error is thrown during notebook execution.

It's possible, I can turn the parameter `nbsphinx_allow_errors` to `False`in: 

https://github.com/scikit-rf/scikit-rf/blob/f9e36e24cdcef5a01ac35a18c6f6474ba0e36c00/doc/source/conf.py#L62

but then it will fail for sure due to notebooks using `pyvisa`

jhillairet:
it's ok now, i've rebuilt the doc

PR #504, opened 2021-07-27
Docstring improvements

jhillairet:
- improve the network.renumber() method doc
- add Network.renumbered() method (to mirror .flipped() or .cropped())
- various docstrings improvements to solve most of Sphinx warnings

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -847,7 +847,7 @@ class Network(object):
         """
         Scattering parameter matrix.
 
-        The s-matrix[#]_ is a 3-dimensional :class:`numpy.ndarray` which has shape
+        The s-matrix [#]_ is a 3-dimensional :class:`numpy.ndarray` which has shape
         `fxnxn`, where `f` is frequency axis and `n` is number of ports.
         Note that indexing starts at 0, so s11 can be accessed by
         taking the slice s[:,0,0].
@@ -856,7 +856,7 @@ class Network(object):
         Returns
         ---------
         s : complex :class:`numpy.ndarray` of shape `fxnxn`
-                the scattering parameter matrix.
+            The scattering parameter matrix.
 
         See Also
         ------------
@@ -875,8 +875,14 @@ class Network(object):
     @s.setter
     def s(self, s: npy.ndarray) -> None:
         """
-        the input s-matrix should be of shape fxnxn,
-        where f is frequency axis and n is number of ports
+        Scattering parameter matrix.
+        
+        Parameter
+        --------
+        s : :class:`numpy.ndarray`
+            The input s-matrix should be of shape `fxnxn`,
+            where f is frequency axis and n is number of ports.
+
         """
         s_shape = npy.shape(s)
         if len(s_shape) < 3:
@@ -1874,10 +1880,8 @@ class Network(object):
         to_archive : zipfile.Zipfile
             opened ZipFile object to place touchstone file in
         form : 'db','ma','ri'
-            format to write data,
-            * db = db, deg
-            * ma = mag, deg
-            * ri = real, imag
+            format to write data:
+            'db': db, deg. 'ma': mag, deg. 'ri': real, imag.
         format_spec_A : string, optional
             Any valid format specifying string as given by https://docs.python.org/3/library/string.html#format-string-syntax
             This specifies the formatting in the resulting touchstone file for the A part of the S parameter, (e.g. the dB magnitude for 'db' format, the linear
@@ -2212,7 +2216,7 @@ class Network(object):
 
         This interpolates  a given `basis`, ie s, z, y, etc, in the
         coordinate system defined by `coord` like polar or cartesian.
-         Different interpolation types ('linear', 'quadratic') can be used
+        Different interpolation types ('linear', 'quadratic') can be used
         by passing appropriate `\*\*kwargs`. This function `returns` an
         interpolated Network. Alternatively :func:`~Network.interpolate_self`
         will interpolate self.
@@ -2221,17 +2225,19 @@ class Network(object):
         -----------
         freq_or_n : :class:`~skrf.frequency.Frequency` or int or listlike
             The new frequency over which to interpolate. this arg may be
-            one of the following
-              * a new `Frequency` object
-              * if an  int, the current frequency span is resampled linearly.
-              * if a listlike, then create its used to create a new frequency
+            one of the following:
+            * a new `Frequency` object
+
+            * if an  int, the current frequency span is resampled linearly.
+
+            * if a listlike, then create its used to create a new frequency
                 object using `Frequency.from_f`
-        basis : ['s','z','y','a'],etc
+
+        basis : ['s','z','y','a'], etc
             The network parameter to interpolate
-        coords : ['cart','polar']
-            coordinate system to use for interpolation.
-             * 'cart' is cartesian is Re/Im
-             * 'polar' is unwrapped phase/mag
+        coords : str
+            Coordinate system to use for interpolation: 'cart' or 'polar':
+            'cart' is cartesian is Re/Im. 'polar' is unwrapped phase/mag
         return_array: bool
             return the interpolated array instead of re-assigning it to
             a given attribute
@@ -2256,12 +2262,13 @@ class Network(object):
         for duts with slowly changing magnitude. try them all.
 
         See  :func:`scipy.interpolate.interpolate.interp1d` for useful
-        kwargs. For example
-            **kind** : str or int
-                Specifies the kind of interpolation as a string ('linear',
-                'nearest', 'zero', 'slinear', 'quadratic, 'cubic') or
-                as an integer specifying the order of the spline
-                interpolator to use.
+        kwargs. For example:
+
+        kind : str or int
+            Specifies the kind of interpolation as a string ('linear',
+            'nearest', 'zero', 'slinear', 'quadratic, 'cubic') or
+            as an integer specifying the order of the spline
+            interpolator to use.
 
 
         See Also
@@ -2504,11 +2511,9 @@ class Network(object):
 
             `d` kwarg controls the degree of rational polynomials
             when `kind` is 'rational'. Defaults to 4.
-
-        coords : ['cart','polar']
-            coordinate system to use for interpolation.
-             * 'cart' is cartesian is Re/Im
-             * 'polar' is unwrapped phase/mag
+        coords : str
+            Coordinate system to use for interpolation: 'cart' or 'polar'. 
+            'cart' is cartesian is Re/Im, 'polar' is unwrapped phase/mag.
              Passed to :func:`Network.interpolate`
 
         Returns
@@ -2600,7 +2605,6 @@ class Network(object):
 
         No interpolation is done.
 
-
         Parameters
         -----------
         f_start : number
@@ -2612,6 +2616,11 @@ class Network(object):
         unit : string
             Units that `f_start` and `f_stop` are described in. This must be a string recognized by the Frequency
             class, e.g. 'Hz','MHz', etc. A value of `None` assumes units are same as `self`
+
+        See Also
+        --------
+        cropped
+
         '''
         if f_start == None:
             f_start = -npy.inf
@@ -2650,11 +2659,29 @@ class Network(object):
 
     def cropped(self, f_start: float, f_stop: float, unit: str = None) -> 'Network':
         '''
-        returns a cropped network, leaves self alone.
+        Returns a cropped network, leaves self alone.
+
+        Parameters
+        -----------
+        f_start : number
+            start frequency of crop range, in units of self.frequency.unit.
+            If `f_start` is lower than the lowest frequency, no change to the network is made by the lower bound.
+        f_stop : number
+            stop frequency of crop range, in units of self.frequency.unit
+            If `f_stop` is higher than the highest frequency, no change to the network is made by the higher bound.
+        unit : string
+            Units that `f_start` and `f_stop` are described in. This must be a string recognized by the Frequency
+            class, e.g. 'Hz','MHz', etc. A value of `None` assumes units are same as `self`
+
+        Returns
+        -------
+        ntwk : :class:`Network` object
+            Resulting cropped network
 
         See Also
         ---------
         crop
+
         '''
         out = self.copy()
         out.crop(f_start=f_start, f_stop=f_stop,unit=unit)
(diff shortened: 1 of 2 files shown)


Vinc0110 (commented):
It's good that you do these improvements to the docstrings.
You are probably not yet finished, but I already have two comments:
 - There are several occasions with too many underscores below "Parameters" and "Returns", e.g. in the noise_*() functions. Sphinx will tell you.
 - You added underscores after many references. e.g. `[4]_`. This is correct, but is it always the same reference, or are the reference numbers reused? The latter will cause problems and results in missing hyperlinks. I found out the hard way: the reference numbers are unique for the class functions and cannot be redefined for each function. Be sure to check the html build.

skrf/network.py
```diff
@@ -467,12 +485,12 @@ def from_z(cls, z: npy.ndarray, *args, **kw) -> 'Network':
             Network, `f` and `z0`.
 
         Return
```
Vinc0110, line 487: Sorry to annoy you, but I just noticed that it should be `Returns` (not `Return`) according to the NumPy convention (multiple occasions). Also there sometimes still is the old `:return:` above the new `Returns` statement.

jhillairet: no problem, you are right. Since I'm on it, it's the right time to make all these small corrections !

skrf/network.py
```diff
@@ -2591,16 +2751,15 @@ def subnetwork(self, ports: Sequence[int], offby: int = 1) -> 'Network':
...
     def crop(self, f_start: float, f_stop: float, unit: str = None) -> None:
-        '''
+        """
         Crop Network based on start and stop frequencies.
 
         No interpolation is done.
 
-
         Parameters
         -----------
```
Vinc0110, line 2764: One too many underscores here for `Parameters` (line 2764)

skrf/network.py
```diff
@@ -2649,28 +2813,56 @@ def crop(self, f_start: float, f_stop: float, unit: str = None) -> None:
         self.frequency, self.s, self.z0 = ntwk.frequency, ntwk.s, ntwk.z0
 
     def cropped(self, f_start: float, f_stop: float, unit: str = None) -> 'Network':
-        '''
-        returns a cropped network, leaves self alone.
+        """
+        Returns a cropped network, leaves self alone.
+
+        Parameters
+        -----------
```
Vinc0110, line 2820: One too many underscores here for `Parameters`

skrf/network.py
```diff
@@ -2680,21 +2872,28 @@ def flip(self) -> None:
...
     def flipped(self) -> 'Network':
-        '''
-        returns a flipped network, leaves self alone.
+        """
+        Returns a flipped network, leaves self alone.
+
+        Returns
+        -------
+        ntwk : :class:`Network` object
+            Resulting flipped Network
```
Vinc0110, line 2881: One too many underscores for `See Also` in `flipped()`.

skrf/network.py
```diff
@@ -2680,21 +2872,28 @@ def flip(self) -> None:
...
         out = self.copy()
         out.flip()
         return out
 
     def renormalize(self, z_new: NumberLike, s_def: str = S_DEF_DEFAULT) -> None:
-        '''
-        Renormalize s-parameter matrix given a new port impedances
-
+        """
+        Renormalize s-parameter matrix given a new port impedances.
```
Vinc0110, line 2896: Way too many underscores for `Parameters` in `renormalize()`.

skrf/network.py
```diff
@@ -2712,30 +2911,102 @@ def renormalize(self, z_new: NumberLike, s_def: str = S_DEF_DEFAULT) -> None:
...
-        '''
-        renumbers  ports of a  Network
+        """
+        Renumbers ports of a Network (inplace).
 
         Parameters
         -----------
         from_ports : list-like
-        to_ports: list-like
+            List of port indices to change. Size between 1 and N_ports.
```
Vinc0110, line 2925: One too many underscore for `Parameters` in `renumber()`.

skrf/network.py
```diff
@@ -2822,9 +3132,9 @@ def windowed(self, window: Union[str, float, Tuple[str, float]]=('kaiser', 6),
 
         References
         -------------
-        .. [1] Agilent Time Domain Analysis Using a Network Analyzer Application Note 1287-12
+        .. [#] Agilent Time Domain Analysis Using a Network Analyzer Application Note 1287-12
```
Vinc0110, line 3135: Way too many underscores here for `References`.

skrf/network.py
```diff
@@ -2994,45 +3335,56 @@ def nonreciprocity(self, m: int, n: int, normalize: bool = False) -> 'Network':
...
-        port renumbering can be used.
+        .. note:: Odd Number of Ports
 
+            In the case where there are an odd number of ports (such as a 3-port network
+            with ports 0, 1, and 2), se2gmm() assumes that the last port (port 2) remains
+            single-ended and ports 0 and 1 are converted to differential mode and common
+            mode, respectively. For networks in which the port ordering is not suitable,
+            port renumbering can be used.
+
+        Example
```
Vinc0110, line 3362: Plural:
```
Examples
--------
```

skrf/network.py
```diff
@@ -4252,14 +4619,14 @@ def subnetwork(ntwk: Network, ports: int, offby:int = 1) -> Network:
     Network.subnetwork, n_twoports_2_nport
 
     Example
-    --------
+    -------
```
Vinc0110, line 4622: Plural:
```
Examples
--------
```

skrf/network.py
```diff
@@ -4606,51 +4973,51 @@ def innerconnect_s(A: npy.ndarray, k: int, l: int) -> npy.ndarray:
...
     References
     ----------
-    .. [1] http://en.wikipedia.org/wiki/S-parameters
-    .. [2] http://en.wikipedia.org/wiki/impedance_parameters
-    .. [3] Kurokawa, Kaneyuki "Power waves and the scattering matrix", IEEE Transactions on Microwave Theory and Techniques, vol.13, iss.2, pp. 194–202, March 1965.
-    .. [4] Marks, R. B. and Williams, D. F. "A general waveguide circuit theory", Journal of Research of National Institute of Standard and Technology, vol.97, iss.5, pp. 533–562, 1992.
+    .. [#] http://en.wikipedia.org/wiki/S-parameters
+    .. [#] http://en.wikipedia.org/wiki/impedance_parameters
+    .. [#Kurokawa] Kurokawa, Kaneyuki "Power waves and the scattering matrix", IEEE Transactions on Microwave Theory and Techniques, vol.13, iss.2, pp. 194–202, March 1965.
+    .. [#Marks] Marks, R. B. and Williams, D. F. "A general waveguide circuit theory", Journal of Research of National Institute of Standard and Technology, vol.97, iss.5, pp. 533–562, 1992.
```
Vinc0110, line 5018: Out of interest: how does this work? Will Sphinx just generate ascending label numbers for these references? And what is the difference between `[#]` with and without a label name? How does Sphinx know which one is which without a label name?

I did not find much information about this in the Sphinx docs... It would be great if you could send a link

skrf/network.py
```diff
@@ -5458,7 +5829,7 @@ def t2s(t: npy.ndarray) -> npy.ndarray:
             scattering parameter matrix.
 
     See Also
-    ---------
+    -------
```
Vinc0110, line 5832: You deleted one too many

Vinc0110 (changes requested):
Great job on this extremely tedious task! It will certainly improve the API docs a lot.

I'm taking the opportunity to report some additional very minor errors that I spotted while scrolling through the changes. Would be great if you could fix them while you are at it ;)

skrf/network.py
```diff
@@ -196,14 +197,13 @@
 
 class Network(object):
     """
-    A n-port electrical network [#]_.
+    A n-port electrical network.
```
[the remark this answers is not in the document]
jhillairet: This first line of docstring appears on the [Table Of Content of the Network documentation](https://scikit-rf.readthedocs.io/en/latest/api/network.html). But then,fails in linking  correctly to the reference here. So I moved the footnote in the paragraph below.

skrf/network.py
jhillairet: `[#]` are auto-numbered footnotes. So there are numbered in the order by which then appear. If you need to re-use a reference, then you should provide a label other than `#`. 

source : https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#footnotes

skrf/network.py
```diff
@@ -2243,12 +2243,14 @@ def write(self, file: str = None, *args, **kwargs) -> None:
         \*args, \*\*kwargs :
             passed through to :func:`~skrf.io.general.write`
 
-        Notes
-        -----
+
+        Note
+        ----
```
Vinc0110, line 2248: I think `Notes` is actually correct, so you should change it back (multiple occasions in this commit):
https://numpydoc.readthedocs.io/en/latest/format.html#notes

Vinc0110 (changes requested):

skrf/network.py
Vinc0110: Thanks a lot!

skrf/network.py
jhillairet: Notes is indeed correct and produces a section "Notes", while "Note" produce a pretty CSS bloc of note

Vinc0110: I see, thanks! So it's both fine and just a matter of taste?

jhillairet: I guess so. I like the CSS one, it's more stylish ;)

skrf/time.py
```diff
@@ -127,100 +123,132 @@ def indexes(y, thres=0.3, min_dist=1):
...
 
-def find_n_peaks(x,n, thres=.9, **kw):
-    '''
-    Find a given number of peaks in a signal
-    '''
-    
-    for dummy  in range(10):
-        
-        idx = indexes(x, **kw)
+def find_n_peaks(x: npy.ndarray, n: int, thres: float = 0.9, **kwargs):
```
Vinc0110, line 130: Don't forget the type hint for the return value. Probably `-> List[int]`

skrf/time.py
```diff
@@ -127,100 +123,132 @@ def indexes(y, thres=0.3, min_dist=1):
...
-    raise ValueError('Couldnt find %i peaks'%n)
-    
-def detect_span(ntwk):
-    '''
-    detect the correct time-span between two largest peaks
-    '''
+    raise ValueError('Couldnt find %i peaks' % n)
+
+
+def detect_span(ntwk) -> float:
```
Vinc0110, line 168: You can add a type hint for the `ntwk` argument (several occasions in methods of this class):
 `ntwk: Network`

However, this requires you to import the `Network` class at the top: `from .network import Network`

Vinc0110 (commented):
I hope I'm not bothering you, but I'm adding a few more comments.

skrf/time.py
jhillairet: ok. Often the return value is NumberLike (scalar, list or numpy aray), which I remember is problematic (not here).

skrf/time.py
jhillairet: I didn't because it created a circular import here... if you have a solution...

Vinc0110: You're right. I was not aware of that problem, but I found a nice article proposing a solution:
https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/

So, what you can do in `time.py` is a conditional import of `Network` only for the type checking:
```
from __future__ import annotations
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .network import Network
```
Then, you can do the type hinting for the methods in `time.py`:
```
def detect_span(ntwk: Network) -> float:
```

Vinc0110: @FranzForstmayr seems to be an expert for type hinting, maybe he can confirm this is a good solution.

jhillairet: > I didn't because it created a circular import here... if you have a solution...

using 'Network' works fine. Sphinx even knows how to link to the Network page!

Vinc0110: Great. No import required at all?

jhillairet: hum, flake8 doesn't like finally. OK, I drop it for now.

Vinc0110: Well, it worked for me in PR #505.

I ended up using a mixture of our tricks above:
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .network import Network
```
But no `from __future__ import annotations`, because it only works for Python >= 3.7. 

So, I had to put `ntwk: 'Network'` as a string.

jhillairet:
I've done most of the files at the root of skrf/, there is still so work to do into plotting.py... 
Interestingly, I discovered or rediscovered some stuffs in the codebase, that were undocumented. Like for example the Taper & Co object that deserves a tutorial (next time...)

skrf/time.py
FranzForstmayr: I think that's the best solution right now. When 3.6 support get's dropped in the future we can start to use  `__annotations__` from `futures`. The `TYPE_CHECKING` is a great way to avoid circular imports.

doc/source/conf.py
```diff
@@ -64,6 +64,9 @@
 numpydoc_show_class_members = False
 nbsphinx_timeout = 120
 
+# Autodoc settings
+autodoc_typehints = "description"
```
Vinc0110, line 68: This worked as expected on my local html build. 

However, I had to upgrade the package version of Sphinx to see the effect of this relatively new option (I was using the Ubuntu's ancient package python3-sphinx version 1.8.5-7ubuntu3). It now works with Sphinx upgraded to version 4.1.2. Not sure if this has be to considered in `requirements/docs.pip`.

jhillairet:
OK, except for calibrations files (next time...), most of the codebase should be OK now. Quite large PR finally!

jhillairet:
Tests are OK, so let's merge! We will see after if we need to impose a specific version of sphinx for RTD.

jhillairet:
OK, the documentation is compiled and online. 

Interestingly, I've found some differences wrt to the doc I've compiled locally. For example, the `Media` base class page does not exist online while it exists locally (?). There is probably other stufs. 

I'm preparing a fix PR for some typos as well, so if you spot errors (as uncorrect section levels in circuit), let me know.

Vinc0110:
Regardless of any remaining issues, you did a great job rectifying this huge number of docstrings!

doc/source/conf.py
jhillairet: I've forced RTD to use a sphinx version > 4.0. 
https://scikit-rf.readthedocs.io/en/latest/

What about now ?

PR #505, opened 2021-07-31
Vector Fitting: Docstring improvements, type hints and bugfix

Vinc0110:
In the spirit of PR #504 and issue #449 this PR improves the docstrings for the `VectorFitting` class and adds type hints for its methods.
A tiny bug in the number formatting in `write_spice_subcircuit_s()` is also fixed.

### DIFF
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -7,14 +7,11 @@ Vector Fitting (:mod:`skrf.vectorFitting`)
     :member-order: groupwise
 
 """
-
-from __future__ import annotations  # according to PEP 563 (will become default in Python 3.11)
-
 import numpy as np
 import os
 
 # imports for type hinting
-from typing import Any, TYPE_CHECKING
+from typing import Any, Tuple, TYPE_CHECKING
 if TYPE_CHECKING:
     from .network import Network
 
@@ -93,7 +90,7 @@ class VectorFitting:
     .. [#vectfit_website] Vector Fitting website: https://www.sintef.no/projectweb/vectorfitting/
     """
 
-    def __init__(self, network: Network):
+    def __init__(self, network: 'Network'):
         """
         Creates a VectorFitting instance based on a supplied :class:`skrf.network.Network` containing the frequency
         responses of the N-port.
@@ -517,7 +514,7 @@ class VectorFitting:
 
         logging.info('\n### Vector fitting finished.\n')
 
-    def _get_ABCDE(self) -> np.ndarray:
+    def _get_ABCDE(self) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
         """
         Private method.
         Returns the real-valued system matrices of the state-space representation of the current rational model, as
@@ -537,6 +534,11 @@ class VectorFitting:
         E : ndarray
             State-space matrix E holding the proportional coefficients (usually 0 in case of fitted S-parameters)
 
+        Raises
+        ------
+        ValueError
+            If the model parameters have not been initialized (by running :func:`vector_fit()` or :func:`read_npz()`).
+
         References
         ----------
         .. [#] B. Gustavsen and A. Semlyen, "Fast Passivity Assessment for S-Parameter Rational Models Via a Half-Size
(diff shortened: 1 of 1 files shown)


skrf/vectorFitting.py
```diff
@@ -45,35 +48,49 @@ def wrapper(*args, **kwargs):
...
+
+    Examples
+    --------
+    Load the `Network`, create a `VectorFitting` instance, perform the fit with a given number of real and
+    complex-conjugate starting poles:
+
+    >>> nw_3port = skrf.Network('my3port.s3p')
+    >>> vf = skrf.VectorFitting(nw_3port)
+    >>> vf.vector_fit(n_poles_real=1, n_poles_cmplx=4)
+
```
jhillairet, line 68: Question : is it a good practice to put the constructor documentation in the class docstring?

skrf/vectorFitting.py
```diff
@@ -112,8 +129,8 @@ def __init__(self, network):
         self.delta_max_history = []
         self.history_max_sigma = []
 
-    def vector_fit(self, n_poles_real=2, n_poles_cmplx=2, init_pole_spacing='lin', parameter_type='S',
-                   fit_constant=True, fit_proportional=False):
+    def vector_fit(self, n_poles_real: int = 2, n_poles_cmplx: int = 2, init_pole_spacing: str = 'lin',
+                   parameter_type: str = 'S', fit_constant: bool = True, fit_proportional: bool = False) -> None:
```
jhillairet, line 133: in the rest of the code, 's' (lowercase) is used of s-parameters.

jhillairet (approved):

skrf/vectorFitting.py
Vinc0110: Apparently, that is the correct way to di it. See https://developer.lsst.io/python/numpydoc.html#documenting-classes:
```The __init__ method never has a docstring since the class docstring documents the constructor.```
(I did it differently before and the constructor still has its own docstring. So, I still have to fix that)

skrf/vectorFitting.py
Vinc0110: It does not matter, both lower and upper case works:
https://github.com/scikit-rf/scikit-rf/blob/8c30a735dd1539b26c5d6bfff8baaa81a2dceae1/skrf/vectorFitting.py#L202

jhillairet: yes I did see, that was just a matter of uniformization in the documentation

skrf/vectorFitting.py
jhillairet: ok, thanks. So I will fix that too for the class I've updated

skrf/vectorFitting.py
Vinc0110: Ok. It's changed for lower case 's' in https://github.com/scikit-rf/scikit-rf/pull/505/commits/3c6554b86f46add3171325adb84a037cdbd3b434

Vinc0110:
Should be ready for merging now.

skrf/vectorFitting.py
jhillairet: I found various recommendations about docstring and class: 

In the [sphinx extension Napoleon documentation](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

>    The __init__ method may be documented in either the class level
    docstring, or as a docstring on the __init__ method itself.
>
>    Either form is acceptable, but the two should not be mixed. Choose one
    convention to document the __init__ method and be consistent with it. 

or advise to put public class attributes in the class documentation. 

[Numpy's Style guide](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes) essentially say the same:

> The constructor (__init__) should also be documented here, the Parameters section of the docstring details the constructor’s parameters.

So I guess it's a matter choice. 

As most of the codebase has the `__init__()` already documented, I will not change it for the moment.

jhillairet:
The [documentation has been compiled](https://scikit-rf.readthedocs.io/en/latest/api/vectorFitting.html)

Vinc0110:
Looks good, thanks. I just created issue #506 for a more general suggestion.

PR #507, opened 2021-08-04
Docstring fixes after PR #504

jhillairet:
Some missed things during PR #504 :
- fixing the level of the sections in circuit.py
- removing the description of the constants in `mathFunction` and linking to `skrf.constants` where they are documented.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -7,7 +7,7 @@ build:
   image: latest
 
 python:
-  version: 3.7
+  version: "3.8"
   install:
     - requirements: requirements.txt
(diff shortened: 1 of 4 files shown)


jhillairet:
if anyone spots some problem in the RTD doc, please let me know before tomorrow.

Vinc0110:
Is the online version on RTD actually updated? Because the type hints are still included in the function signatures. If it has been updated, this might be related to the problem I had with my local test build. See https://github.com/scikit-rf/scikit-rf/pull/504#discussion_r682010520

PR #510, opened 2021-08-05
Single Ended to Mixed Mode conversion with reference impedance different from 50 Ohms

FranzForstmayr:
se2gmm and back conversion assumed having a 50 ohm reference impedance. This PR get's the diff/common reference impedance directly out of the original z0.


Vinc0110 (approved):
That was a good idea!

jhillairet (approved):

Issue #518, opened 2021-09-14
DC Extrapolation fails with noisy network


FranzForstmayr:
`extrapolate_to_dc` fails with noisy networks, when using a interpolation method different to `rational`.
The DC value is not inserted into the noise array, so the interpolation raises a Value error, as a `x` value is outside the interpolation range.

I'll add an PR with a failing test to show the behaviour.

`rational` interpolation seems to ignore the boundaries. Don't know if this should be an seperate issue?

PR #519, opened 2021-09-14
Fix/518

FranzForstmayr:
This PR enables DC extrapolation for noisy networks


jhillairet:
fixes #518 

Thank you

Issue #517, opened 2021-09-14
Handling of duplicate frequency values 


FranzForstmayr:
Is there a recommended way how to handle S-Parameter measurements with duplicate frequency entries?
I think these measurements are taken by accident most of the time, when  specifying a segemented sweep or something in VNAs, but maybe there's another possibility.

An example for this behaviour is in the reposity:
[ntwk_noise.s2p](https://github.com/scikit-rf/scikit-rf/blob/master/skrf/tests/ntwk_noise.s2p)

With the duplicate x-value you'll get strange issues, for example scipy's interpolation functions require the x values to be unique. 

Should we create a method, to drop duplicate values? Or is there another recommended way?

jhillairet:
No, there is no recommended way to handle duplicate frequency entries. 

And indeed, a lot of Numpy functions would not like that....

A workaround could be to convert the s parameters to pandas, drop duplicates with pandas `drop_duplicates` and then recreate a new Network from this cleaned data. 

Eventually, why not a  `drop_duplicates` or, more explicit, a `drop_frequency_duplicates` method in Network?

FranzForstmayr:
> Eventually, why not a drop_duplicates or, more explicit, a drop_frequency_duplicates method in Network?

Yeah, I would prefer something like this too. Should we throw a Warning when creating a Network with duplicates to let the user decide if the duplicates should be dropped?

jhillairet:
yes, eventually, and promoting the drop_frequency_duplicates at the same time. 

I wonder if having duplicate frequency values could be a "normal" case for some users..

FranzForstmayr:
Touchstone specification states:
> All data sets must be arranged in increasing order of frequency.

So it's probably not even a legal touchstone file with two equal frequency points. 

Would you suggest to drop the duplicates silently or with Warning automatically in this case?
I would go for the warning and suggest to drop method.

jhillairet:
No, I'm not in favour to drop it silently. The warning message should just state that there is some frequency duplicate point, and if this is unexpected, one can use the xxxx method to drop them.

FranzForstmayr:
fixed with #523

PR #523, opened 2021-09-16
Warn if invalid frequency data is detected and method to fix the data

FranzForstmayr:
See issue #517


jhillairet:
Wouldn't that make sense to separate dropping duplicates and non-increasing frequency values? `drop_invalid `can do both of course, calling `drop_duplicates() `and `drop_non_monotonic_increasing()`. 

And instead of `check_values()` which name is rather unexplicit, I propose (like in pandas) `is_monotonic_increasing()` and `duplicated()` (which [returns the duplicate values](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.duplicated.html#pandas.DataFrame.duplicated)? It can be use as a test using `duplicated().any ` 

what do you think?

FranzForstmayr:
Good Idea to go similar to pandas API!
I'll change the code in a few days.

FranzForstmayr:
Ok, thought again about this issue. 
Splitting `drop_invalid` into `drop_duplicates` and `drop_non_monotonic_increasing` would generate a lot more code, without any gain for the user. In addition we would need two seperate Warnings, to distinguish these two cases. The methods would be needed in frequency and network classes, so I don't think it's worth to change.

However, I'm open to rename the `drop_invalid` to `drop_non_strictly_monotic_increasing` or something else. Duplicates are not possible anyway, if the function is strictly monotonic increasing.

Would this be ok for you?

jhillairet:
> Splitting `drop_invalid` into `drop_duplicates` and `drop_non_monotonic_increasing` would generate a lot more code, without any gain for the user. In addition we would need two seperate Warnings, to distinguish these two cases. The methods would be needed in frequency and network classes, so I don't think it's worth to change.

OK, I got your point. 
 
> However, I'm open to rename the `drop_invalid` to `drop_non_strictly_monotic_increasing` or something else. Duplicates are not possible anyway, if the function is strictly monotonic increasing.

Yes, I think a more meaningful function name is preferable.  `drop_non_monotonic` is maybe sufficient to understand what the function does.

jhillairet:
merged. Thank you.

PR #526, opened 2021-09-21
logging error is never thrown if passivity enforcement does not converge.

FranzForstmayr:
The last iteration of passivity enforcement sets `t` to `max_iterations`, so the warning condition was not triggered.


Vinc0110 (approved):
Good catch, thanks!

PR #530, opened 2021-09-26
LRRM calibration

Ttl:
LRRM is probably the most widely used calibration that is still missing from scikit-rf. It's often used for on-wafer calibration at mm-wave frequencies but it's usable also in other situations. This implementation combines the best parts from few different LRRM papers and the functionality looks good so far. Example notebook is included to show how to use it and what assumptions are made of the calibration standards.

Three alternatives for match determination are provided. 'none' uses independent inductance for each frequency, this doesn't currently work correctly with non-z0 match. 'l' which is the default fits a single inductance value to the match. 'lc' fits inductor and parallel capacitor assuming second reflect is open with some unknown capacitance that is the same at all frequencies. Tests are included for 'l' and 'none'. No test is included for 'lc' right now as it doesn't work too well using the waveguide media without low frequency measurements. The iterative approach it uses is also not accurate enough for the equality test of the calibration parameters. Example using it is however included in the example notebook and it should be usable with real measurements.

### DIFF
--- a/doc/source/examples/metrology/LRRM.ipynb
+++ b/doc/source/examples/metrology/LRRM.ipynb
@@ -1,0 +1,555 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# LRRM"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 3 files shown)


jhillairet:
Looks great!

jhillairet:
Is it ready to merge ?

jhillairet:
@Ttl, is it OK to merge?

PR #532, opened 2021-09-26
LRM calibration

Ttl:
Not sure how useful this is but it was very easy to implement based on the LRRM calibration.


jhillairet:
Ready to merge?

jhillairet:
@Ttl , there are few conflicts after merging #530. Could you please check?

PR #548, opened 2021-10-29
Zipped touchstone fix

stevehamblin:
Fix Network.zipped_touchstone() and add unit test.

This broke due to changes in Touchstone.load_file after 0.17 release.  Unit test coverage for zipped files was only present in test_networkSet.py, though NetworkSet.from_zip did not leverage Network.zipped_touchstone, so the issue escaped unit testing.

### DIFF
--- a/skrf/io/general.py
+++ b/skrf/io/general.py
@@ -70,7 +70,6 @@ import pickle
 from pickle import UnpicklingError
 import sys
 import warnings
-import zipfile
 
 import numpy as npy
 
--- a/skrf/io/touchstone.py
+++ b/skrf/io/touchstone.py
@@ -26,6 +26,7 @@ Functions related to reading/writing touchstones.
 """
 import re
 import os
+import typing
 import zipfile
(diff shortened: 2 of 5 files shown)


jhillairet:
Great PR, thank you!
It fails for Python 3.6 however. Can you have a look?

stevehamblin:
Ah, now I understand why StringIO was used in NetworkSet.from_zip... ZipFile was not seekable in Python 3.6.
I'll change my Network.zipped_touchstone to use the StringIO technique.

jhillairet:
> 
> 
> Ah, now I understand why StringIO was used in NetworkSet.from_zip. ZipExtFile is not seekable in Python 3.6. I'll change my Network.zipped_touchstone to use the StringIO technique.

yes indeed, Python 3.6 is still supported for the moment. Thanks for the update.

jhillairet:
Super 1st contribution @stevehamblin! Thank you and welcome!

PR #552, opened 2021-10-31
Multiline TRL documentation improvements

Ttl:
Add reference plane shift and impedance renormalization examples. Improved explanation of the existing examples. Added more noise to the synthetic files so that the calibration accuracy is better visible in the plots.


jhillairet:
Great addition, thank you ! It is ready for you to be merged?

Ttl:
Should be good to merge.

jhillairet:
The documentation has been built and updated.

Issue #553, opened 2021-11-01
Vector Fitting: difficult to fit data with a large negative slope


alexatadi:
Matlab has an example for this, fitting improved by introduced a "delay factor". Any way to deal with it in scirf? Thanks.
https://www.mathworks.com/help/rf/ug/using-the-delayfactor-parameter.html

jhillairet:
@Vinc0110, could you have a look?

Vinc0110:
Hi @alexatadi,
unlike the Matlab routine, the implementation in scikit-rf's `vector_fit()` does not include this kind of delay "improvement".

Out of curiosity: Why would you need such a feature? Can you maybe share some example data that you need to fit?

But regardless of its use cases, you should be able to achieve a similar result with the delay function in `Network.delay()`:
Instead of delaying the vector fitted results, you can "accelerate" the frequency responses before the fitting by applying a negative delay to them. 
If I understood the operation of the Matlab function correctly (see also https://de.mathworks.com/help/rf/ref/rationalfit.html), the effect should be the same, even though a network with negative delay is unphysical.

Example code for the delay:
```
import skrf
import matplotlib.pyplot as mplt

# using the ring slot as example 2-port
nw = skrf.data.ring_slot
nw.name = 'original'

# 300 micron delay at port 1
nw_delayed = nw.delay(300, unit='um', port=0)
nw_delayed.name = 'delayed'

# -300 micron delay at port 1
nw_accel = nw.delay(-300, unit='um', port=0)
nw_accel.name = 'acclerated'

# plot magnitude and phase
fig = mplt.figure(figsize=(6, 4))
ax = fig.subplots(2, 1)
nw.plot_s_db(1, 0, ax=ax[0])
nw.plot_s_deg_unwrap(1, 0, ax=ax[1])
nw_delayed.plot_s_db(1, 0, ax=ax[0])
nw_delayed.plot_s_deg_unwrap(1, 0, ax=ax[1])
nw_accel.plot_s_db(1, 0, ax=ax[0])
nw_accel.plot_s_deg_unwrap(1, 0, ax=ax[1])
fig.tight_layout()
mplt.show()
```
![delay](https://user-images.githubusercontent.com/63644854/139705653-5f19b2eb-54ff-41e9-97ac-274ce604262d.png)

alexatadi:
Hi @Vinc0110 ,
[...]

PR #561, opened 2021-11-09
Fix DefinedAEpTandZ0 example

Ttl:
This example had a same problem as the one I fixed in PR #555 caused by PR #553. Explicitly set the TRL reference plane to center of thru to match the existing behaviour. Also fix the phase offset in calibration by setting the correct reflect offset.


jhillairet (approved):

jhillairet:
the online documentation has been rebuilt and updated.

PR #563, opened 2021-11-12
Warn users about future change of 'extrapolate_to_dc' default value from rational to cubic

### LINKED ISSUE #522: 'extrapolate_to_dc' seems to process very slowly
Hi, I am sort of new to this, so hope this is ok to post here.

It seems like the `extrapolate_to_dc` method takes a very long time (~30 s per 5,000 point dataset) on Google colab.  Is there a particular computational step in `extrapolate_to_dc` that could be done much more efficiently/quickly?

Before finding scikit-rf, I wrote amateur code that did similar extrapolations to dc which ran much faster. 

If you are reading this, thank you for your time.  Any thoughts or help here would be greatly appreciated!!
-bob

jhillairet:
Fixes #522. 

Keep the default value to `rational`, but warns users this choice is now considered as deprecated and will be changed to `cubic` by default in future versions.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2671,8 +2671,9 @@ class Network(object):
         # freq = Frequency.from_f(f,**kwargs)
         # self.interpolate_self(freq, **interp_kwargs)
 
-    def extrapolate_to_dc(self, points: int = None, dc_sparam: NumberLike = None, kind: str = 'rational',
-            coords: str = 'cart', **kwargs) -> 'Network':
+    def extrapolate_to_dc(self, points: int = None, dc_sparam: NumberLike = None, 
+                          kind: str = 'warn', coords: str = 'cart',
+                          **kwargs) -> 'Network':
         """
         Extrapolate S-parameters down to 0 Hz and interpolate to uniform spacing.
 
@@ -2691,7 +2692,7 @@ class Network(object):
         dc_sparam : class:`npy.ndarray` or None
             NxN S-parameters matrix at 0 Hz.
             If None S-parameters at 0 Hz are determined by linear extrapolation.
-        kind : str or int
+        kind : str or int, default is 'rational'
             Specifies the kind of interpolation as a string ('linear',
             'nearest', 'zero', 'slinear', 'quadratic, 'cubic') or
             as an integer specifying the order of the spline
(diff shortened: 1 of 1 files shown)


skrf/network.py
```diff
@@ -2691,7 +2692,7 @@ def extrapolate_to_dc(self, points: int = None, dc_sparam: NumberLike = None, ki
         dc_sparam : class:`npy.ndarray` or None
             NxN S-parameters matrix at 0 Hz.
             If None S-parameters at 0 Hz are determined by linear extrapolation.
-        kind : str or int
+        kind : str or int, default is 'rational'
```
Vinc0110, line 2695: Maybe also write here, that this default is going to be changed to 'cubic' in the future.

skrf/network.py
```diff
@@ -2717,6 +2718,14 @@ def extrapolate_to_dc(self, points: int = None, dc_sparam: NumberLike = None, ki
         impulse_response
         step_response
         """
+        if kind == 'warn':
```
Vinc0110, line 2721: Why did you introduce `kind='warn'`? Can't you just keep `kind='rational'` and check against this value to print the warning? For example:
```
if kind == 'rational':
    warnings.warn('...')
```

skrf/network.py
```diff
@@ -2717,6 +2718,14 @@ def extrapolate_to_dc(self, points: int = None, dc_sparam: NumberLike = None, ki
...
+            warnings.warn("The default `kind` parameter will change "
+                          "from `rational` to `cubic` in future versions."
+                          "Use `kind='rational'` to keep the old default."
+                          "To silent this warning, explitly define `kind`.",
+                          category=DeprecationWarning, stacklevel=2)
```
Vinc0110, line 2726: Did you figure out what `stacklevel=2` does? I know, [the Python doc](https://docs.python.org/3/library/warnings.html#warnings.warn) says *"The stacklevel argument can be used by wrapper functions written in Python"*, but it that important?

Vinc0110 (changes requested):
Looks good, but I made a few suggestions above.

skrf/network.py
FranzForstmayr: The warning would be printed every time, even if you select `kind='rational'` excplicit.

Vinc0110: Oh, you're right.

skrf/network.py
jhillairet: I found out that `stacklevel=1` (default) didn't display any message in ipython... so I've used 2.

skrf/network.py
jhillairet: done

PR #566, opened 2021-11-14
Fixing coverage

jhillairet:
pytest documentation (https://pypi.org/project/pytest-cov/) seems to indicate that the project name and directory must be mentioned

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -56,20 +56,30 @@ jobs:
       run: |
         python -m pytest --nbval-lax --current-env -p no:python --ignore=doc/source/examples/instrumentcontrol/ doc/source/examples/
 
-    - name: Coveralls Parallel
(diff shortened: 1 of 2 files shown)


FranzForstmayr:
@jhillairet, the issue is not about having pytest as test executot, but the usage of tox.
Tox creates an own environment for each run (py36 - py310). This can be seen in coverall which is reporting a coverage in python's site packages directory.

Try to add the following line below [testenv]

```
[testenv]
usedevelop=True
```

jhillairet:
> Try to add the following line below [testenv]
> 
> ```
> [testenv]
> usedevelop=True
> ```

I tried but it did not change the coverage. 

NB : the `--cov-append` increases a little the coverage.  I suppose some of the code used in the notebooks is not tested, so testing the notebooks examples increases a little the coverage.

Issue #565, opened 2021-11-14
Coverage statistics not working


Vinc0110:
The [badge from Coveralls](https://coveralls.io/github/scikit-rf/scikit-rf) with the current coverage of the tests in scikit-rf, which is linked on the front page, is showing a surprisingly low value of 38%.
When you go to that website, all values for "skrf" are at 0%. Only when you expand a massive tree for each Python release in `/opt/hostedtoolcache/Python/...`, you will eventually find some coverage values for the modules. However, it's not showing any details about hits and misses within the code, which used to be there. Instead, this error is shown:
```
Source Not Available

The file "/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/scikit_rf-0.20.0-py3.7.egg/skrf/vectorFitting.py" isn't available on github. Either it's been removed, or the repo root directory needs to be updated.
```

Is this maybe related to the changes from Nose to Pytest in PR #560 ? Or is it just about a missing update of the test reports after the last package release?

jhillairet:
Yes indeed. The coverage dropped from  after https://github.com/scikit-rf/scikit-rf/commit/624bb2f748bf17dc3538b8e706b8373753e04620 which indeed indicates it comes from the change to pytest. Looks like the directory skrf is not included.

FranzForstmayr:
I'll have a look later, sorry for the inconvenience!

jhillairet:
with the previous version of the CI (nose), the coverage file tree looked like that:
![image](https://user-images.githubusercontent.com/4642848/141683273-f7893903-0908-41da-a1fb-b70ae7ba09f3.png)

while now it looks like: 
![image](https://user-images.githubusercontent.com/4642848/141683312-921ce3ee-5a81-44b5-81b8-5eaa2bfe0517.png)

jhillairet:
ok, just have seen your message in #566 
let's try.

PR #568, opened 2021-11-14
Try to fix coverall stats

FranzForstmayr:


jhillairet:
OK I see... So, what is the point of the "--cov-append" option? Would be nice to combine the coverage of the notebooks and the one of the tests, if possible.

FranzForstmayr:
Now all tests are run within a single tox session. Is this fine? 
The duration is slightly longer of course, however, it's easier for developers to test both tests and notebooks.

jhillairet:
Few comments to be discussed:
- with the latest commit, running `pytest` also run the tests for all the notebooks, which takes of course muuuch more time. This is what bothers me the most: I think this is not very convenient for daily development, when users want to launch often the test suite on their computer to check their code... What do you think?
- the solution you found in 08a1594 of inverting the order of the test (putting the tests of the code at last) [also worked](https://coveralls.io/builds/44249987). In such case, I understand that the tests of the notebooks are not included in the coverage (61% vs 75%). This solution would allow using the `pytest` command locally to test only the code test suite, but reduces a bit the coverage percentage.
- question : how does tox deals with this multiple python versions? in the tox.ini they are defined twice. On my computer, running tox only runs the tests with the python version it found and produces an error for the other versions.

FranzForstmayr:
> * with the latest commit, running `pytest` also run the tests for all the notebooks, which takes of course muuuch more time. This is what bothers me the most: I think this is not very convenient for daily development, when users want to launch often the test suite on their computer to check their code... What do you think?

Yes, that's true. I'll see what's possible. 

> * the solution you found in [08a1594](https://github.com/scikit-rf/scikit-rf/commit/08a15941e655cceb306aa738deeeae5758bd455b) of inverting the order of the test (putting the tests of the code at last) [also worked](https://coveralls.io/builds/44249987). In such case, I understand that the tests of the notebooks are not included in the coverage (61% vs 75%). This solution would allow using the `pytest` command locally to test only the code test suite, but reduces a bit the coverage percentage

`pytest` alone still works as expected, however if you have `tox` installed, you can run the tests again multiple python versions and the installer is tested too. 

> * question : how does tox deals with this multiple python versions? in the tox.ini they are defined twice. On my computer, running tox only runs the tests with the python version it found and produces an error for the other versions.
`tox` uses the python versions specified in the `envlist` if they are installed. Unfortunately github runner specifies python3.9 as `py3.9` whereas `tox` uses `py39`. `tox-github-actions` allows to create a translation table.

FranzForstmayr:
Coverage seems to be pretty good, as the `vi` part is excluded from statistics.
Maybe we can find a solution to test the `vi` functions with mocked components later.

.github/workflows/testing.yml
```diff
@@ -38,23 +30,12 @@ jobs:
...
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-
-    # tests suites & creates coverage report
-    - name: Test with pytest
-      if: ${{ always() }}    
-      run: |
-        pytest
-
+  
     # test notebooks
```
jhillairet, line 34:
```suggestion
    # test code and notebooks (tutorials and examples)
```

.github/workflows/testing.yml
```diff
@@ -38,23 +30,12 @@ jobs:
...
-    - name: Run tutorials notebooks with nbval
-      if: ${{ always() }}    
-      run: |
-        python -m pytest --nbval-lax --current-env -p no:python doc/source/tutorials/
-
-    - name: Run examples notebooks with nbval
+    - name: Run tutorial & example notebooks with nbval
```
jhillairet, line 35:
```suggestion
    - name: Test the code, tutorials and examples
```

tox.ini
```diff
@@ -1,25 +1,32 @@
...
 commands =
     pytest {posargs}
 
 [pytest]
 testpaths = 
     skrf
-addopts = "--cov=skrf"
+    doc/source/examples
+    doc/source/tutorials
+addopts = --cov
```
jhillairet, line 25: I don't understand why the command `pytest` test only the `skrf` directory despite the doc is also indicated, but it works... ! ;) could you explain to me this magic ?

jhillairet:
> Coverage seems to be pretty good, as the vi part is excluded from statistics.
>Maybe we can find a solution to test the vi functions with mocked components later.

OK, super nice. We will see for the vi module later, as it requires anyhow a full update per se.

jhillairet (approved):

.github/workflows/testing.yml
jhillairet: just to update the comments

tox.ini
FranzForstmayr: I'm not hundred percent sure, I think `.ipynb` files aren't even recognized without setting the option `--nbval`.

FranzForstmayr: This is the regex pytest is using by default.
[pytest ini parser](https://github.com/pytest-dev/pytest/blob/062d91ab474881b58ae1ff49b4844402677faf45/src/_pytest/python.py#L108)

jhillairet: OK. So the test of the notebook is performed via tox because of the `nbval` option.

FranzForstmayr:
Remove --cov-append as it isn't necesary anymore. I'm fine with this PR :)

jhillairet:
fixes #565

PR #571, opened 2021-11-16
Fix/dc extrapolation array

FranzForstmayr:
`extrapolate_to_dc` was not able to handle a numpy array for `dc_s_param`. This patch fixes the comparison against None in this case and in several other files.


jhillairet (approved):
I've learned with your PR that `is None` is faster than `== None` !

OK to me.

FranzForstmayr:
Not only faster, comparing a numpy array against None with '==' will fail.
```
a = np.arange(10)
if a == None: # Will raise ValueError
```

PR #576, opened 2021-11-28
Documentation fixes

jhillairet:
- details how to use pytest with a particular test file or directory
- fixes some links of the references in the tutorials
- VectorFitting higher in the TOC list

NB : Sphinx-like references (like `[1]_`) are not processed correctly in notebook markdown cells. One option is to use instead markdown footnote (`[^1]`) but this requires the reference part to be in the same cell (not very convenient). I found a warkaround using html references (`href=#something` pointing on div anchors)


doc/source/contributing/index.rst
```diff
@@ -106,15 +106,23 @@ Or, to run tests for every installed python installation in an isolated environm
 
     tox 
 
-Or, to run test a single module or single test, 
+
+If you want to test a single file or directory, you need to override the defaukt pytest configuration (defined in the tox.ini file) and to indicate the test path. By example, to run only the tests associated to the Network object (-v to increase the verbosity):
```
Vinc0110, line 110: Typo in `override the defaukt pytest'

doc/source/tutorials/VectorFitting.ipynb
```diff
@@ -167,17 +167,17 @@
    "metadata": {},
    "source": [
     "## References\n",
-    "<a id='link_ref1'>[1]</a> B. Gustavsen and A. Semlyen, \"Rational approximation of frequency domain responses by vector fitting,\" in IEEE Transactions on Power Delivery, vol. 14, no. 3, pp. 1052-1061, July 1999, DOI: [10.1109/61.772353](https://doi.org/10.1109/61.772353).\n",
+    "<div id='link_ref1'></div>[1] B. Gustavsen and A. Semlyen, \"Rational approximation of frequency domain responses by vector fitting,\" in IEEE Transactions on Power Delivery, vol. 14, no. 3, pp. 1052-1061, July 1999, DOI: [10.1109/61.772353](https://doi.org/10.1109/61.772353).\n",
```
Vinc0110, line 170: Thanks for fixing.
Still, I find it strange that this syntax was wrong, because I created the link with the Jupyter GUI. Or is this more like a Sphinx thing?

doc/source/examples/index.rst
```diff
@@ -70,6 +70,15 @@ Mixed Mode
 
     mixedmodeanalysis/*
 
+Vector Fitting
```
Vinc0110, line 73: I don't mind the change, but why is the order important?

Vinc0110 (commented):
I'm just adding a few random comments. But what's going on with the tests failing? Is it about pytest again?

doc/source/tutorials/VectorFitting.ipynb
jhillairet: it works indeed under Jupyter, but no more after the conversion into html... so maybe, yes.

doc/source/examples/index.rst
jhillairet: Well, the Instrument control and the Interactive things are not as well documented as VectorFitting functionalities. So I think it's better to put VectorFitting above the former.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 matplotlib
 scipy
-numpy
+numpy>=1.9.5
```
FranzForstmayr, line 3: you mean 1.19.5 I assume, 1.9.5 is quite old

jhillairet: ho yes of course!

jhillairet:
> I'm just adding a few random comments. But what's going on with the tests failing? Is it about pytest again?

Yes, seems so: running the tests of the notebooks locally with or without coverage triggers the same error:
`pytest -c /dev/null --nbval-lax doc/source/tutorials/Introduction.ipynb`
works, while
`pytest -c /dev/null --nbval-lax --cov doc/source/tutorials/Introduction.ipynb`
doesn't. 

In the previous days, PRs used `coverage` 6.1.2 while now it's 6.2 ([released 2 days ago](https://github.com/nedbat/coveragepy/releases/tag/6.2)). And indeed, going back to 6.1.2 locally works fine...

It's not the first time I see such incompatibility arising between coverage and notebook (like [here](https://github.com/computationalmodelling/nbval/issues/116)). This is why I've pinned the version of coverage for the moment.

doc/source/tutorials/Deembedding.ipynb
```diff
@@ -167,7 +167,7 @@
    "source": [
     "## Through-only de-embedding\n",
     "\n",
-    "At higher frequencies above 10 GHz, the fringe capacitance of the open dummy and the parasitic inductance of the short dummy cause errors. For high frequencies, several methods using a through (thru) dummy have been proposed, such as `SplitPi`[^1], `SplitTee`[^2], `AdmittanceCancel`[^3], and `ImpedanceCancel`[^4]. The thru dummy is a direct connection of the left and right test pads. The example of the test device and thru dummy is shown in the figure below, where G and S indicate ground pads and signal pads.\n",
+    "At higher frequencies above 10 GHz, the fringe capacitance of the open dummy and the parasitic inductance of the short dummy cause errors. For high frequencies, several methods using a through (thru) dummy have been proposed, such as `SplitPi`[[1](#ref1)] , `SplitTee`[[2](#ref2)]_ , `AdmittanceCancel`[[3](#ref3)], and `ImpedanceCancel`[[4](#ref4)]. The thru dummy is a direct connection of the left and right test pads. The example of the test device and thru dummy is shown in the figure below, where G and S indicate ground pads and signal pads.\n",
```
keikawa, line 170: Thank you for fixing the references.
There's an unnecessary underscore.
`SplitTee`[[2](#ref2)]_

jhillairet:
The problem with coverage has been reproduced and confirmed in https://github.com/nedbat/coveragepy/issues/1286

jhillairet:
> The problem with coverage has been reproduced and confirmed in [nedbat/coveragepy#1286](https://github.com/nedbat/coveragepy/issues/1286)

ok so now it works again with the latest version of coverage. Seems like the `--cov` flag requires the (eventually multiple) path(s) to cover: `--cov=skrf --cov=doc` for example.

The [coverage hence obtained](https://coveralls.io/builds/44686054) is slightly reduced [from the previous one](https://coveralls.io/builds/44584270) as the vi/ directory has been included. 

tried to add in tox.ini the section:
```
[coverage:run]
omit = 
    skrf/vi
    skrf/src
```
but doesn't help.

FranzForstmayr:
Don't you thin the issue will be fixed from coverage?
We could look at this issue later, and just tag the coverage version to 6.1.2

jhillairet:
> Don't you thin the issue will be fixed from coverage? We could look at this issue later, and just tag the coverage version to 6.1.2

Well, in fact from the discussion on the coverage issue, I'm not sure if this is an issue or a functionality ;-) 

The [coverage documentation indicates](https://pytest-cov.readthedocs.io/en/latest/config.html) the parameter `--cov`  should take a path as `--cov=PATH`. 

I don't understand what is going on if no path is passed. For example, locally, using `--cov=.` does not produce the 75% coverage results but much less.

Vinc0110:
I'm almost afraid to ask: 
Why and how are the docs a valid test that increases the coverage? Sure, one can check if certain code lines in an example notebook get executed properly, but what does that tell? Unlike the "real" unit tests, the results and actions of code in the docs is not checked against a (correct) reference value, right?

jhillairet:
> I'm almost afraid to ask: Why and how are the docs a valid test that increases the coverage? Sure, one can check if certain code lines in an example notebook get executed properly, but what does that tell? Unlike the "real" unit tests, the results and actions of code in the docs is not checked against a (correct) reference value, right?

I think you are right, and probably having only the skrf/ directory in the coverage scope is representative enough. 

Anyway, adding the doc/ directory to the scope does not change the coverage score, so I'm not even sure coverage look at the notebooks in any case.

Vinc0110:
Ok, thanks for clarifying. So this is only about the `vi` module, which is not really testable?
>The coverage hence obtained is slightly reduced from the previous one as the vi/ directory has been included.

Would it make sense to remove the docs from the coverage scan in order to reduce the runtime of the tests? I think @FranzForstmayr mentioned that it takes much longer with the notebooks included.

FranzForstmayr:
It takes much longer, indeed, but this is due to the execution time of the notebooks itself. Maybe the coverage increases, as we add much more lines, which are executed and the lines which are not executed stay the same.

If we would remove the notebooks, we could run in a issue, that an example from the docs won't work. However, it's not needed for any python version, i think it's enough to run the notebooks only when building the docs.

@jhillairet, 
You trigger the sphinx build manually, would it be possible to set up an CI job for this?
We could run docs including notebooks only on master branch, and go with the unit tests for coverage and tests.

Vinc0110:
That sounds more reasonable to me.

In general, I have often been annoyed by these longs CI tests that get executed every time I push a commit to a work-in-progress branch. Not sure if this is feasible with GitHub, but why not only run all these tests occasionally when I'm done and want to have them checked? Then, maybe one more test triggered by a maintainer to be sure it's fine before merging it into the master branch.

jhillairet:
> It takes much longer, indeed, but this is due to the execution time of the notebooks itself. Maybe the coverage increases, as we add much more lines, which are executed and the lines which are not executed stay the same.
> 
> If we would remove the notebooks, we could run in a issue, that an example from the docs won't work. However, it's not needed for any python version, i think it's enough to run the notebooks only when building the docs.

I think it is wise to keep testing the notebooks at each PR/push but yes, testing the notebooks for only a single Python version is probably sufficient and would be faster. 

> @jhillairet, You trigger the sphinx build manually, would it be possible to set up an CI job for this? We could run docs including notebooks only on master branch, and go with the unit tests for coverage and tests.

Well, eventually. 

Or for each merge. 

But we may miss something as well. I would more be in favour of keeping the test of the notebooks for each PR/push, but with a single python version. Can be a separate GitHub action test.

jhillairet:
will modify CI as proposed in a future PR

FranzForstmayr:
There's an option to run actions only when the PR is not in draft state. Should we run notebooks only when draft == False?
So the ci would be quite fast for development and slow if you finalize your PR.

jhillairet:
> There's an option to run actions only when the PR is not in draft state. Should we run notebooks only when draft == False? So the ci would be quite fast for development and slow if you finalize your PR.

yeah, eventually. As most of the time the work is on the code rather than on the notebooks. PR is coming.

PR #577, opened 2021-12-02
Vector Fitting: Notation and other minor improvements

### LINKED ISSUE #570: Vector Fitting: Error in equation?
Is it possible that there is an description mistake in the [first Vector Fitting equation](https://scikit-rf.readthedocs.io/en/latest/tutorials/VectorFitting.html) of the docs? The one to calculate the transfer function of a system.

My suggestion is that instead of the zeros (z_k) in the numerator it should be the residues (c_k) because it's a sum of partial fractions and therefore has no zeros in this form. Sure you can calculate the zeros from that partial form but I'm wondering if that should be the residues instead of the zeros there.

So, instead of this:
aa
is should be this?
bb

Please let me know if I'm wrong here.
Thanks!

[Github Permalink to docs](https://github.com/scikit-rf/scikit-rf/blob/c50c7cc9444d03145be22d40e3c72d863b0abdf8/doc/source/tutorials/VectorFitting.ipynb?short_path=5e8554d#L19)

Vinc0110:
Fixes #570


Vinc0110:
Looks like the failed tests are related to PR #576. The poor thing was trying for 59 minutes :hankey:

jhillairet:
OK, I will merge the #576 for now, then I will update the CI after

Vinc0110:
It works now.
This PR is done, so feel free to merge if you are okay with it.

PR #578, opened 2021-12-02
Separate the test of the notebooks from the test of the code 

jhillairet:
split notebook tests into an independent github action: 

- the code is tested against multiple python versions (as before)
- while the notebooks are tested for only one python version, and independently of the code tests, and only for non-draft PR


jhillairet:
seems to work. Tests are made in ~ 5 min.

PR #582, opened 2021-12-04
Fix citation formatting that was a run-on line

temmeand:
The citation for the IEEE article was formatted as code. Sphinx displayed this as a very long line. This is now formatted as a block quote.


jhillairet:
doc is updated:
https://scikit-rf.readthedocs.io/en/latest/citations_ack/index.html

PR #583, opened 2021-12-04
DOI hyperlink and minor fixes in docs

Vinc0110:
Made a few minor changes:
- Adding a hyperlink for the DOI of the new Microwave Magazine article.
- Abbreviating journal name according to [IEEE-style guide](http://ieeeauthorcenter.ieee.org/wp-content/uploads/Magazine-Titles-and-Abbreviations.pdf): `Microw. Mag.` 
- Fixing broken hyperlink in Vector Fitting example 2
- Changing another occurrence of `Zeros` to `Residues` (missing in PR #577)


Vinc0110:
The abbreviation of the journal name is obviously still open for discussion. Now that I think about it again, it might be better to leave it intact.

jhillairet:
Well, even IEEE seems to not agree with itself : if you click on 
image
and ask for the BibTeX reference, the name of the magazine is not abbreviated.

Vinc0110:
Right. I've often noticed that the editing team at IEEE creating the meta data on IEEE Xplore doesn't do a great job with these things, especially not with the BibTex. 
Also, the style guide only applies for the references in IEEE journals, while someone might be citing our paper in a non-IEEE journal.

So maybe it's best to undo this change and leave the full name?

jhillairet:
I think more is better than less, people who wants to use abbreviation will handle it themselves, but it's not a big deal anyway.

Vinc0110:
So this PR is ready, I think. Please merge.

Issue #584, opened 2021-12-04
Multi-parameters NetworkSet 


jhillairet:
Currently `NetworkSet` does not explicitly support the definition of multiple parameters/dimensions (issue #535). While it is possible to create a NetworkSet and interpolate the results with respect to a parameter (from the [interpolate_from_network](https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.networkSet.NetworkSet.interpolate_from_network.html) method), it only concerns a unique parameter/dimension. 

## Purpose and objectives

A desired feature would be to allow creating NetworkSet in which the parameters for which each Network has been obtained (like temperature, voltage or whatever can be changed between the Networks of the NetworkSet) can be set, then to extract or interpolate data for specific parameters.  

prerequisites: 
 * retrocompatibility (do not create a new object NetworkSetX)

below we use the same nomenclature like in [xarray](http://xarray.pydata.org/en/stable/index.html), as xarray is now a defacto  standard for N-dimensional data. So the parameters for which a network can   be stored in a NetworkSet are _dimensions_ and the values of these parameters are the _coordinates_. 

Let be M(=5) Networks which depends on 2 dimensions `a` and `b`:
 * a = 0, b = 10 -> ntwk_0
 * a = 1, b = 10 -> ntwk_1
 * a = 2, b = 10 -> ntwk_2
 * a = 0, b = 20 -> ntwk_3
 * a = 1, b = 20 -> ntwk_4

## Initializations
multi-dimensional parameters NetworkSet _desired_ behaviour pseudo-code
```python
# the parameter definition is made for each Network, prior to make a NetworkSet
all_params = [
        dict(a=0, b=10, c='A'),
        dict(a=1, b=10, c='A'),
        dict(a=2, b=10, c='A'),
        dict(a=0, b=20, c='A'),
        dict(a=1, b=20, c='A')
    ]
# create M dummy networks which all contain a 'params' dictionnary
ntwks_params = [rf.Network(frequency=freq1, 
                                s=np.random.rand(len(freq1),2,2), 
                                name=f'ntwk_{m}',
                                params=params) \
                     for (m, params) in enumerate(all_params) ]

# Create NetworkSet from a list of Network (as usual)
ns_params = rf.NetworkSet(ntwks_params)  
```
the created NetworkSet contains the new additional parameters
```python
ns.dims -> tuple
ns.coords -> dict
```
### from (and to) a (G)MDIF file (ADS)
issue #359 

[...]

PR #585, opened 2021-12-08
multi-parameters NetworkSet

jhillairet:
Requirements as described in #584

# Status 
- [x] add `params` in Network
- [x] deduce `.dims` and `.coords` NetworkSet properties
- [x] sub-NetworkSet selection from a given parameter values: `ns.sel({'a': 1})` return a len(2) NetworkSet
- [x] sub-NetworkSet selection from a given range of parameter values
- [x] sub-NetworkSet selection from a dictionnary 
- [x] Interpolation from parameter 
- [x] from MDIF
- [x] <strike> from MDL </strike>
- [x] from CITI
- [x] Documentation


FranzForstmayr:
At the moment all the dims are stored inside the NetworkSet. Wouldn't it be better to use a 'params' dictionary property in each Network and create the dims in the NetworkSet dynamically? NetworkSet holds it's Networks in a simple list, and is able to query by parameter.

jhillairet:
will continue in January... have good holidays!

jhillairet:
As I didn't received any MDIF or etc kind of file. So I may merge as it is, and we will see later for the support of these files.

arsenovic:
julien i dont have any mdif files but i found some here. 
https://github.com/hftools/hftools/tree/master/hftools/file_formats/tests/testdata/mdif

i had forgotten about this package, but i also noticed that hftools has some great file-format support here, https://github.com/hftools/hftools/tree/master/hftools/file_formats

maybe some of this can be re-used ?

jhillairet:
> maybe some of this can be re-used ?

I have looked to it, but the implementation is rather obfuscated by many intermediate Objects, and I wasn't able to run it locally properly. But I'll have a look at the file examples.

jhillairet:
>  But I'll have a look at the file examples.

Woah, some files are formatted very differently with very unconventional headers...

jhillairet:
Here is a working version. Only miss the documentation yet.

PR #587, opened 2021-12-16
fix reading ISO-8859-1 encoded Touchstone file and add encoding param to Network() and Touchstone()

### LINKED ISSUE #586: Touchstone parser does not accept alternative encodings
Any encoding that is an alternative to default UTF-8 makes touchstone parsing fail.

This is because encoding should be an argument to skrf.Network and in /skrf/io/touchstone.py the line 74 that is
`fid = get_fid(file)`
should be replaced by 
`fid = get_fid(file,encoding)`

jhillairet:
Fixes #586

Guessing the encoding used in a file seems is a complicate problem per se. 

Here this fix only fixes bug #586, but the problem may still raise for other kinds of encoding... 

Is a better way exist?


jhillairet:
damned, for under Windows but not on Linux. Will continue later.

Ttl:
`UnicodeDecodeError` is also a `ValueError`. If you swap the order of excepts the tests pass.

jhillairet:
> `UnicodeDecodeError` is also a `ValueError`. If you swap the order of excepts the tests pass.

ho great, thanks.

skrf/io/touchstone.py
```diff
@@ -105,7 +117,34 @@ def __init__(self, file: typing.Union[str, typing.TextIO]):
...
+            self.load_file(fid)
+
+        except ValueError:
+            # Assume Microsoft UTF-8 variant encoding with BOM
+            fid = get_fid(file, encoding='utf-8-sig')
+            self.filename = fid.name
+            self.load_file(fid)
+
+        except Exception as e:
+                raise ValueError(f'Something went wrong by the file openning: {e}')
```
jhillairet, line 147:
```suggestion
        except Exception as e:
            raise ValueError(f'Something went wrong by the file openning: {e}')
```

jhillairet:
by the way I have also took the opportunity to add into the docstring that `Touchstone()` (and thus  `Network()`) can take an `io.StringIO` object as input, following the discussion in Slack.

skrf/io/touchstone.py
```diff
@@ -105,7 +117,34 @@ def __init__(self, file: typing.Union[str, typing.TextIO]):
         self.port_names = None
 
         self.comment_variables = None
-        self.load_file(fid)
+        
+        # open the file depending on encoding
+        # Guessing the encoding by trial-and-error, unless specified encoding
```
FranzForstmayr, line 122: If a encoding is passed and causes an Exception, there still an trial-and error approach with the two latest encodings, however the default encoding is skipped.
Why not adding all encodings, plus the custom one into a list and try all of them by default.

```python
encodings = ['latin1', 'ISO-8859-1', 'utf-8-sig']
encodings.insert(0, encoding) # Default encoding is None anyway

fid = None
for enc in encodings:
    try:
        fid = get_fid(file, encoding=encoding)
        break
    except ValueError:
        pass

if not fid:
    raise ValueError(f'Something went wrong by opening the file opening: {e}')

self.filename = fid.name
self.load_file(fid)
```

FranzForstmayr: Maybe it's even easier to try enough encodings by default and omit the additional parameter

FranzForstmayr (changes requested):

skrf/io/touchstone.py
jhillairet: While it's more elegant, it would still fail if one tries to open a Windows-like UTF-8 file (for example). Because `get_fid` would returns a fid when reading the file as 'latin-1', but `load_file` would fail after due to the BOM of the file. And if it's not `load_file`, then it would be `get_gamma_z0_from_fid` to fail.

PR #588, opened 2021-12-16
Correct error when asking db/ma from ri file

edumur:


jhillairet:
wow, nice catch. Seems like a few users use the `ri` format! It also means this is not correctly tested.

edumur:
To be fare, it was caught by one of the lab student.
We were both impressed that no one saw it before (we checked and the code had been written 11 years ago).

jhillairet:
Could you add a test in `test_touchstone.py` and a simple touchstone file in that format?

edumur:
I quickly added something that I hope is enough (no idea how these tests work).

If it is not enough, I will be back the 3rd of January ;) .

Live long and prosper 🖖.

jhillairet:
> I quickly added something that I hope is enough (no idea how these tests work).

Great, many thanks!

 > Live long and prosper vulcan_salute.

Have nice holidays!

jhillairet:
and thanks to lab student!

Issue #572, opened 2021-11-22
How to use UnknownThru


YasuoKawahara:
Hello,

I compared UNknownTHRU and SOLT.
I focused on the return loss of LINE1, but the result of UNknownTHRU (result_UNknownTHRU.jpg) is about 20 dB, while the result of SOLT (result_SOLT.jpg) is about 40 dB.
I thought both would give similar results, but is this result reasonable?

I have attached the archive file (UNknownTHRUvsSOLT.zip) .
Its contents are a set of my Python script files and measurement data files.
I would be very grateful if you could check my Python script for any mistakes.

Thank you.

![result_SOLT](https://user-images.githubusercontent.com/94821303/142806561-0d1b5ed8-9295-4be6-bb11-e71bf4f29990.jpg)
![result_UNknownTHRU](https://user-images.githubusercontent.com/94821303/142806591-07fb197e-1b83-4d1b-b129-b0c558f2ce8b.jpg)
[UNknownTHRUvsSOLT.zip](https://github.com/scikit-rf/scikit-rf/files/7578448/UNknownTHRUvsSOLT.zip)

jhillairet:
Dear @YasuoKawahara , 

How did you generate your ideal components? And measured ones?

YasuoKawahara:
Dear @jhillairet,

Thank you for your inquiry.

The calkit I used is 85052D.
The standard of this calkit is defined as follows.
Open:
C0 = 49.433e-15
C1 = -310.13e-27
C2 = 23.168e-36
C3 = -0.15966
Delay = 2.9243E-11
Loss = 2.2
Z0 = 50
Short:
L0 = 2.0765e-12
L1 = -108.54e-24
L2 = 2.1705e-33
L3 = -0.01e-42
Delay = 3.1785E-11
Loss = 2.36
Z0 = 50
Load:
Delay = 0
Loss = 0
Z0 = 50
Thru:
Delay = 0
Loss = 0
Z0 = 50

I entered the above values into ADS to create ideal data.

The measurement data is the measurement result of PNA-X.
I measured each standard of 85052D in UnCal state and saved it in TouchStone format.

Please let us know if you have any additional information you need.
Thank you.

JAnderson419:
[...]

PR #589, opened 2021-12-16
Fix coefs_8term and coefs_12term

Ttl:
Fix the issue reported by @JAnderson419 in #572. Accessing `coefs_8term` of 8-term calibration raised an Exception. Added tests to cover this case.


arsenovic:
great thanks !

PR #591, opened 2021-12-21
Bugfixes

jjoonathan:
Hi! I've been using scikit-rf recently and ran into a few bugs. Here are the fixes that got things moving with my setup -- though I'm pretty sure I have root-caused both of them and I think the fixes should apply to everyone. Cheers!

### DIFF
--- a/skrf/vi/vna/__init__.py
+++ b/skrf/vi/vna/__init__.py
@@ -37,6 +37,7 @@ Available VNAs
 """
 
 from .abcvna import VNA
+from .hp8510c import HP8510C
 from .keysight_pna import PNA, PNAX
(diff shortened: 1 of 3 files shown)


jhillairet:
Dear @jjoonathan, welcome and thank you for your contribution! Indeed, the VI part of the code certainly requires some updates. 

I have a limited access to internet, so unless @arsenovic look at your PR befire I can, I will not be able to review it until January.

jjoonathan:
Hi! No rush, January sounds good. Also, the bugfixes are independent of each other and the new 8510 driver, so we can break those up if needed.

jhillairet (approved):
That's very fine to me, thanks for your contribution

PR #601, opened 2022-01-19
MTRL fixes

### LINKED ISSUE #600: NISTMultilineTRL Discrepancy
Possible issue with NistMultilineTRL in calibration.py, reported to the scikit-rf google group by @ZiadHatab. NISTMultilineTRL inherits from EightTerm - is this potentially related to #572?

https://github.com/scikit-rf/scikit-rf/blob/2457bc4cb8ef2a188ecde89bfa373ca6b0ac775d/skrf/calibration/calibration.py#L2486

> Hi there!
> 
> I was testing your NIST multiline TRL (MultiCal) implementation using 150Ghz on-wafer measurements and I noticed that the result I get from your implementing differ from the one implemented I get from WinCal (https://www.formfactor.com/product/probes/calibration-tools/wincal-xe/)
> 
> I also programmed NIST MultiCal mTRL myself on python, and I get the same result as in WinCal, but different from the one in skrf. And to confirm this, I even tested it with my own original mTRL implementation, and its the same. All implemention agree with each other, but skrf differ?!
> 
>image
>
> You can check my github repo:
> https://github.com/ZiadHatab/multiline-trl-calibration
> You also find the on-wafer measurements i used. Also, attached is a screenshot comparing skrf calibration with the others.
> 
> best,
> Ziad

Ttl:
Fixes #600. While the noiseless case was correct, the lines weren't combined optimally. In Vc and Vb matrix calculations the lines were flipped which resulted in conjugated matrices, this seems to be a problem also in the reference implementation. Also fixes problems in the multical propagation constant determination heurestic and makes it the default method.

I wrote a script for checking the accuracy of the calibration with measurement noise: https://gist.github.com/Ttl/7dadb41a56eabdcc94bdd5f93c49980b. The results are below, it's the standard deviation of S-parameters of calibrated minus ideal DUT, lower is better:

```
master:
[[0.00479818 0.00304272]
 [0.00321142 0.00515514]]

this PR:
[[0.00361597 0.00302612]
 [0.00318306 0.00368045]]

MultiCal by ZiadHatab:
[[0.00361597 0.0030259 ]
 [0.00318309 0.00368045]]
```

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -2491,6 +2491,8 @@ class NISTMultilineTRL(EightTerm):
     calibration. Different line measurements are combined in a way that minimizes
     the error in calibration.
 
+    Calibration reference plane is at the edges of the lines.
+
     At every frequency point there should be at least one line pair that has phase
     difference that is not 0 degrees or a multiple of 180 degrees otherwise
     calibration equations are singular and accuracy is very poor.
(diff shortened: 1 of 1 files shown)


jhillairet (approved):

jhillairet:
OK to merge it, but before doing it, is your latest PR #603 affects this PR ? I mean, I need to merge the PRs in a correct order to avoid cancelling some of your work.

PR #603, opened 2022-01-22
Remove confusing transposes

Ttl:
Calibration used `.transpose().reshape(-1,2,2)` to generate some arrays from lists of array coefficients. The arrays had frequency as the last axis which needed to be moved to the first axis, but the problem with the previous code was that it also transposed the matrix. All the input matrices were confusingly written in transposed order to work around it. This PR replaces them with `.transpose(2,0,1)` which just moves the frequency axis from the last position to the first one without transposing the other axes.

Also includes some other cleanup such as replacing for loops with matrix products.


jhillairet (approved):
great upgrade

PR #605, opened 2022-01-29
Speed up vector fitting

FranzForstmayr:
I tried to use vectorized numpy operations in the vector fitting function and gained a significant performance gain. I did not measure exactly, but it's about 10 times faster.
The example in the docs finishes under two seconds on my machine.

This patch rearranges some data structures so that numpys vectorized operations can be used. I did not change anything else, the result _should_ be exactly the same.

Happy Review :)


skrf/vectorFitting.py
```diff
@@ -1,3 +1,4 @@
+from line_profiler import line_profiler
```
jhillairet, line 1: tests fail because of this import. I guess this is for testing purpose, so can be removed
```suggestion
```

jhillairet (changes requested):

dxxx9:
Yes!! The execution time (compared to the original MATLAB implementation) was really slowing me down.
Thank you very much! 👍

jhillairet (approved):

jhillairet:
Thank you @FranzForstmayr  for this new optimization work ! Once @dvincentwest is OK, I'm OK to merge and include it in the next release.

FranzForstmayr:
You mean @Vinc0110, don't you?

jhillairet:
> You mean @Vinc0110, don't you?

ho yes, indeed!

Vinc0110:
Sounds cool, but I'd like to check the changes in some more detail. Probably tomorrow

Vinc0110 (changes requested):
Awesome improvements, well done!

My local tests:
- Example4 (1+26): `master`: 19.6 s; `this`: 1.1 s
- Example4 (1+25): `master`: 14.9 s; `this`: 1.0 s
- Example2 (4+4): `master`: 4.8 s; `this`: 0.3 s
- Example2 (3+4): `master`: 2.2 s; `this`: 0.16 s
- Nasty 4-port (0+200): `master`: 2151 s; `this`: 249 s

However, I have a few questions, comments and suggestions regarding the implementation (also see inline comments):
- Readability: My previous implementation might not have been the cleanest, but I did my best to make things clear with comments. It would be great if you could add verbose comments to your indexing magic and those one-line wonders (which some Python hobbyists or dummies like myself probably won't understand).
- Complex versus real-valued data structures: Doing the conversions analytically and implementing separate real-valued equations gave quite a speed improvement in my previous optimization. Isn't this also an easy option in your vectorized implementation? For example, calling `np.imag(pole)` was fairly expensive. Is `poles.imag` not that bad?
- How exactly is the data organized? `poles` and `freq_response` are complex-valued ndarrays, and then you write things like `freq_response[:, None]`. How does that even work, i.e. what is the second index and why is it `None`? Are the real and imaginary parts stored separately, so they can be accessed directly?

It would be nice if you could explain these things some more. Maybe we can improve things even more, either in terms of speed or at least in terms of comprehensibility. After all, a simple and easy to understand implementation will increase the chance of other people adding new features or improvements.

FranzForstmayr:
> My local tests:
> 
> * Example4 (1+26): `master`: 19.6 s; `this`: 1.1 s
> * Example4 (1+25): `master`: 14.9 s; `this`: 1.0 s
> * Example2 (4+4): `master`: 4.8 s; `this`: 0.3 s
> * Example2 (3+4): `master`: 2.2 s; `this`: 0.16 s
> * Nasty 4-port (0+200): `master`: 2151 s; `this`: 249 s

I'll try the nasty 4-port example, I'm curious what's the exact bottlneck there. 

> * Readability: My previous implementation might not have been the cleanest, but I did my best to make things clear with comments. It would be great if you could add verbose comments to your indexing magic and those one-line wonders (which some Python hobbyists or dummies like myself probably won't understand).

Yeah, I'll add some more comments for better understandibilty. The numpy indexing is a little bit hard first, but when you get it, you'll love it :).

> * Complex versus real-valued data structures: Doing the conversions analytically and implementing separate real-valued equations gave quite a speed improvement in my previous optimization. Isn't this also an easy option in your vectorized implementation? For example, calling `np.imag(pole)` was fairly expensive. Is `poles.imag` not that bad?

I was not aware, that you have splitted the complex values in the last optimization round. You compared the speed of an local variable access against a function or method call. In every programming language the function call will be slower than the local variable access. So you trade some speed against readabilty. But the main problem was not the function call itself, but the fact that it was called millions of times. (I found the exact number with [kernprof](https://github.com/pyutils/line_profiler)). With the new implementation, a function call is still slower, but it happens only a few times per iteration. I don't expect any wonders if we split up Re/Im again.

> * How exactly is the data organized? `poles` and `freq_response` are complex-valued ndarrays, and then you write things like `freq_response[:, None]`. How does that even work, i.e. what is the second index and why is it `None`? Are the real and imaginary parts stored separately, so they can be accessed directly?

This is numpy broadcasting magic. The ':'-dimension is taken from the array, the `None`-Dimension is broadcasted if possible.
So if you multiply `omega[:,None] * poles`, and `omega` is a vector with 100 frequency points and `poles` is a vector with 10 elements, you'll end up with an array shape `[100, 10]`. Every pole is multiplied with every frequency point. 

> It would be nice if you could explain these things some more. Maybe we can improve things even more, either in terms of speed or at least in terms of comprehensibility. After all, a simple and easy to understand implementation will increase the chance of other people adding new features or improvements.

I'll explain things in more detail if you like. I hope I find some time tomorrow to answer the open points from the review.

Vinc0110:
Thanks for the explanations. I guess I'll have to do get familiar with the broadcasting syntax.

By the way, the "nasty 4-port" was provided by a user in issue: https://github.com/scikit-rf/scikit-rf/issues/553#issuecomment-957003759. The 200 starting poles are just a random guess and the relocation loop does not converge properly, stopping after 100 iterations, but it's a good benchmark. The long runtime is due to the large number of poles, I think.

> But the main problem was not the function call itself, but the fact that it was called millions of times. (I found the exact number with [kernprof](https://github.com/pyutils/line_profiler)). With the new implementation, a function call is still slower, but it happens only a few times per iteration. I don't expect any wonders if we split up Re/Im again.

So, your vectorized version obviously runs much faster compared to my nested loops (running `N_responses * N_freqs * N_poles` times), but doesn't numpy still has to calculate/extract the imaginary parts (and other things) the same number of times? After all, it still has to calculate all the complex coefficients for all poles at every frequency, building the `[(N_responses * N_freqs) x N_poles]` coefficient matrix for the solver. It might also be obvious, but I still don't fully understand why the vectorized version is that much faster. Of course, it's not your job to explain this to me, but maybe you have an idea and can tell me (and others?).

FranzForstmayr:
You are right, the number of mathematical operations stays exactly the same. The difference is the execution time. When you call an multiplication `n` times in python, the interpreter will need a few nanoseconds per iteration. The interpreter does not know beforehand, that the command is the same for each iteration. If you multiply two arrays with size `n` with numpy, numpy knows that there's only one operation, and can use efficient methods inside the CPU. In addition to that, the numpy multiplication is written in C (or Fortran, don't know), which is much faster than python in general. With only vectorized methods, you can reach the speed of an C implementation for some well optimized tasks.

jhillairet:
@Vinc0110 if you're OK I'll merge.

jhillairet:
Ok, i will wait for it then

Le ven. 4 févr. 2022 17:45, FranzForstmayr ***@***.***> a
écrit :

FranzForstmayr:
I profiled the nasty four port example and the most time is spent in QR decomposition:

```
Total time: 231.246 s
File: /home/franz/Workspaces/python/scikit-rf/skrf/vectorFitting.py
Function: vector_fit at line 151

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   151                                               @profile
   152                                               def vector_fit(self, n_poles_real: int = 2, n_poles_cmplx: int = 2, init_pole_spacing: str = 'lin',
   153                                                              parameter_type: str = 's', fit_constant: bool = True, fit_proportional: bool = False) -> None:
   154                                                   """
   155                                                   Main work routine performing the vector fit. The results will be stored in the class variables
   156                                                   :attr:`poles`, :attr:`residues`, :attr:`proportional_coeff` and :attr:`constant_coeff`.
   157                                           
   158                                                   Parameters
   159                                                   ----------
   160                                                   n_poles_real : int, optional
   161                                                       Number of initial real poles. See notes.
   162                                           
   163                                                   n_poles_cmplx : int, optional
   164                                                       Number of initial complex conjugate poles. See notes.
   165                                           
   166                                                   init_pole_spacing : str, optional
   167                                                       Type of initial pole spacing across the frequency interval of the S-matrix. Either linear (lin) or
   168                                                       logarithmic (log).
   169                                           
   170                                                   parameter_type : str, optional
   171                                                       Representation type of the frequency responses to be fitted. Either *scattering* (:attr:`s` or :attr:`S`),
   172                                                       *impedance* (:attr:`z` or :attr:`Z`) or *admittance* (:attr:`y` or :attr:`Y`). As scikit-rf can currently
   173                                                       only read S parameters from a Touchstone file, the fit should also be performed on the original S
   174                                                       parameters. Otherwise, scikit-rf will convert the responses from S to Z or Y, which might work for the fit
   175                                                       but can cause other issues.
   176                                           
   177                                                   fit_constant : bool, optional
   178                                                       Include a constant term **d** in the fit.
   179                                           
   180                                                   fit_proportional : bool, optional
   181                                                       Include a proportional term **e** in the fit.
   182                                           
   183                                                   Returns
   184                                                   -------
   185                                                   None
   186                                                       No return value.
   187                                           
   188                                                   Notes
   189                                                   -----
   190                                                   The required number of real or complex conjugate starting poles depends on the behaviour of the frequency
   191                                                   responses. To fit a smooth response such as a low-pass characteristic, 1-3 real poles and no complex conjugate
   192                                                   poles is usually sufficient. If resonances or other types of peaks are present in some or all of the responses,
   193                                                   a similar number of complex conjugate poles is required. Be careful not to use too many poles, as excessive
   194                                                   poles will not only increase the computation workload during the fitting and the subsequent use of the model,
   195                                                   but they can also introduce unwanted resonances at frequencies well outside the fit interval.
   196                                                   """
   197                                           
   198         1          2.0      2.0      0.0          timer_start = timer()
   199                                           
   200                                                   # create initial poles and space them across the frequencies in the provided Touchstone file
   201                                                   # use normalized frequencies during the iterations (seems to be more stable during least-squares fit)
   202         1         35.0     35.0      0.0          norm = np.average(self.network.f)
   203         1          8.0      8.0      0.0          freqs_norm = np.array(self.network.f) / norm
   204                                           
   205         1         14.0     14.0      0.0          fmin = np.amin(freqs_norm)
   206         1          9.0      9.0      0.0          fmax = np.amax(freqs_norm)
   207         1          2.0      2.0      0.0          if init_pole_spacing == 'log':
   208                                                       pole_freqs_real = np.geomspace(fmin, fmax, n_poles_real)
   209                                                       pole_freqs_cmplx = np.geomspace(fmin, fmax, n_poles_cmplx)
   210         1          1.0      1.0      0.0          elif init_pole_spacing == 'lin':
   211         1         32.0     32.0      0.0              pole_freqs_real = np.linspace(fmin, fmax, n_poles_real)
   212         1         35.0     35.0      0.0              pole_freqs_cmplx = np.linspace(fmin, fmax, n_poles_cmplx)
   213                                                   else:
   214                                                       warnings.warn('Invalid choice of initial pole spacing; proceeding with linear spacing.', UserWarning,
   215                                                                     stacklevel=2)
   216                                                       pole_freqs_real = np.linspace(fmin, fmax, n_poles_real)
   217                                                       pole_freqs_cmplx = np.linspace(fmin, fmax, n_poles_cmplx)
   218                                           
   219                                                   # init poles array of correct length
   220         1          3.0      3.0      0.0          poles = np.zeros(n_poles_real + n_poles_cmplx, dtype=complex)
   221                                           
   222                                                   # add real poles
   223         1          6.0      6.0      0.0          for i, f in enumerate(pole_freqs_real):
   224                                                       omega = 2 * np.pi * f
   225                                                       poles[i] = - omega
   226                                           
   227                                                   # add complex-conjugate poles (store only positive imaginary parts)
   228         1          1.0      1.0      0.0          i_offset = len(pole_freqs_real)
   229       201        228.0      1.1      0.0          for i, f in enumerate(pole_freqs_cmplx):
   230       200        237.0      1.2      0.0              omega = 2 * np.pi * f
   231       200        240.0      1.2      0.0              poles[i_offset + i] = (-0.01 + 1j) * omega
   232                                           
   233                                                   # save initial poles (un-normalize first)
   234         1          4.0      4.0      0.0          self.initial_poles = poles * norm
   235         1          1.0      1.0      0.0          max_singular = 1
   236                                           
   237         1         56.0     56.0      0.0          logging.info('### Starting pole relocation process.\n')
   238                                           
   239                                                   # stack frequency responses as a single vector
   240                                                   # stacking order (row-major):
   241                                                   # s11, s12, s13, ..., s21, s22, s23, ...
   242         1          2.0      2.0      0.0          freq_responses = []
   243         5         10.0      2.0      0.0          for i in range(self.network.nports):
   244        20         26.0      1.3      0.0              for j in range(self.network.nports):
   245        16         21.0      1.3      0.0                  if parameter_type.lower() == 's':
   246        16         28.0      1.8      0.0                      freq_responses.append(self.network.s[:, i, j])
   247                                                           elif parameter_type.lower() == 'z':
   248                                                               freq_responses.append(self.network.z[:, i, j])
   249                                                           elif parameter_type.lower() == 'y':
   250                                                               freq_responses.append(self.network.y[:, i, j])
   251                                                           else:
   252                                                               warnings.warn('Invalid choice of matrix parameter type (S, Z, or Y); proceeding with scattering '
   253                                                                             'representation.', UserWarning, stacklevel=2)
   254                                                               freq_responses.append(self.network.s[:, i, j])
   255         1         25.0     25.0      0.0          freq_responses = np.array(freq_responses)
   256                                           
   257                                                   # ITERATIVE FITTING OF POLES to the provided frequency responses
   258                                                   # initial set of poles will be replaced with new poles after every iteration
   259         1          1.0      1.0      0.0          iterations = self.max_iterations
   260         1          1.0      1.0      0.0          self.d_res_history = []
   261         1          1.0      1.0      0.0          self.delta_max_history = []
   262         1          1.0      1.0      0.0          self.history_cond_A = []
   263         1          1.0      1.0      0.0          converged = False
   264                                           
   265         1          4.0      4.0      0.0          omega = 2 * np.pi * freqs_norm
   266                                                   
   267                                           
   268       101        192.0      1.9      0.0          while iterations > 0:
   269       100        565.0      5.7      0.0              logging.info('Iteration {}'.format(self.max_iterations - iterations + 1))
   270                                           
   271                                                       # count number of rows and columns in final coefficient matrix to solve for (c_res, d_res)
   272                                                       # (ratio #real/#complex poles might change during iterations)
   273                                           
   274       100       2532.0     25.3      0.0              n_cols_unused = np.sum((poles.imag != 0) + 1)
   275                                           
   276       100        198.0      2.0      0.0              n_cols_used = n_cols_unused
   277       100        271.0      2.7      0.0              n_cols_used += 1
   278       100        189.0      1.9      0.0              if fit_constant:
   279       100        215.0      2.1      0.0                  n_cols_unused += 1
   280       100        179.0      1.8      0.0              if fit_proportional:
   281                                                           n_cols_unused += 1
   282       100        264.0      2.6      0.0              n_rows_A = n_cols_used * len(freq_responses)
   283                                           
   284                                                       # generate coefficients of approximation function for each target frequency response
   285                                                       # responses will be reduced independently using QR decomposition
   286                                                       # simplified coeff. matrices of all responses will be stacked in matrix A for least-squares solver
   287       100      29439.0    294.4      0.0              A = np.empty((n_rows_A, n_cols_used))
   288       100        693.0      6.9      0.0              b = np.zeros(n_rows_A)
   289                                           
   290      1700       6191.0      3.6      0.0              for i_response, freq_response in enumerate(freq_responses):
   291                                                           # calculate coefficients for each frequency response
   292                                                           # A_sub which will be reduced first (QR decomposition) and then filled into the main
   293                                                           # coefficient matrix A
   294                                           
   295      1600      99795.0     62.4      0.0                  A_sub = np.empty((len(freqs_norm), n_cols_unused + n_cols_used), dtype=complex)
   296      1600       5614.0      3.5      0.0                  A_row_extra = np.empty(n_cols_used)
   297                                           
   298                                                           # responses will be weighted according to their norm;
   299                                                           # alternative: equal weights with weight_response = 1.0
   300                                                           # or anti-proportional weights with weight_response = 1 / np.linalg.norm(freq_response)
   301      1600      33067.0     20.7      0.0                  weight_response = np.linalg.norm(freq_response)
   302                                           
   303                                                           
   304                                           
   305                                                           # Split up real and complex poles and store the correspondend column from A_sub
   306      1600      14219.0      8.9      0.0                  real_mask = poles.imag == 0
   307      1600      19926.0     12.5      0.0                  poles_real = poles[np.nonzero(real_mask)]
   308      1600      15063.0      9.4      0.0                  poles_cplx = poles[np.nonzero(~real_mask)]
   309                                           
   310      1600     723819.0    452.4      0.3                  A_sub_real_idx = np.nonzero(self._get_real_pole_mask(poles))[0]
   311      1600     697873.0    436.2      0.3                  A_sub_cplx_idx = np.nonzero(~self._get_real_pole_mask(poles))[0][::2]
   312                                           
   313                                           
   314                                                           # add coefficients for a pair of complex conjugate poles
   315                                                           # part 1: first sum of rational functions (residue variable c)
   316                                                           # merged with
   317                                                           # part 3: second sum of rational functions (variable c_res)
   318      1600      60013.0     37.5      0.0                  denom = (omega[:, None] ** 2 + poles_real ** 2)
   319      1600     121731.0     76.1      0.1                  coeff = - (poles_real + 1j * omega[:, None]) / denom
   320                                           
   321                                                           # part 1: coeff = 1 / (s_k - p') = coeff_re + j coeff_im
   322      1600      70153.0     43.8      0.0                  A_sub[:, A_sub_real_idx] = coeff
   323                                           
   324                                                           # part 3: coeff = -1 * H(s_k) / (s_k - pole)
   325                                                           # Re{coeff} = -1 * coeff_re * resp_re + coeff_im * resp_im
   326                                                           # Im{coeff} = -1 * coeff_re * resp_im - coeff_im * resp_re
   327      1600      88203.0     55.1      0.0                  A_sub[:, A_sub_real_idx + n_cols_unused] = - coeff * freq_response[:, None]
   328                                           
   329                                                           # extra equation to avoid trivial solution:
   330                                                           # coeff += Re(1 / (s_k - pole)) = coeff_re
   331      1600      77497.0     48.4      0.0                  A_row_extra[A_sub_real_idx] = np.sum(coeff.real, axis=0)
   332                                           
   333                                           
   334                                                           # coefficient for a complex pole of a conjugated pair: p = p' + jp''
   335      1600     969119.0    605.7      0.4                  denom1 = poles_cplx.real ** 2 + (omega[:, None] - poles_cplx.imag) ** 2
   336      1600     888191.0    555.1      0.4                  denom2 = poles_cplx.real ** 2 + (omega[:, None] + poles_cplx.imag) ** 2
   337                                           
   338                                                           # row 1: add coefficient for real part of residue
   339                                                           # part 1: coeff = 1 / (s_k - pole) + 1 / (s_k - conj(pole))
   340      3200    1516168.0    473.8      0.7                  coeff = (-1 * poles_cplx.real * (1 / denom1 + 1 / denom2)
   341      1600    2074845.0   1296.8      0.9                                  +1j * ((poles_cplx.imag - omega[:, None]) / denom1 - (poles_cplx.imag + omega[:, None]) / denom2))
   342      1600    3264703.0   2040.4      1.4                  A_sub[:, A_sub_cplx_idx] = coeff
   343                                           
   344                                                           # extra equation to avoid trivial solution:
   345                                                           # coeff += Re{1 / (s_k - pole) + 1 / (s_k - conj(pole))}
   346                                           
   347      1600     586688.0    366.7      0.3                  A_row_extra[A_sub_cplx_idx] = np.sum(coeff.real, axis=0)
   348                                                           # part 3: coeff = -1 * H(s_k) * [1 / (s_k - pole) + 1 / (s_k - conj(pole))]
   349      1600    5732488.0   3582.8      2.5                  A_sub[:, A_sub_cplx_idx + n_cols_unused] = - coeff * freq_response[:, None]
   350                                           
   351                                                           # part 1: coeff = 1j / (s_k - pole) - 1j / (s_k - conj(pole))
   352                                                           #               = (p'' - omega + j p') * (1 / denom2 - 1 / denom1)
   353      3200    2313011.0    722.8      1.0                  coeff = ((omega[:, None] - poles_cplx.imag) / denom1 - (omega[:, None] + poles_cplx.imag) / denom2
   354      1600    1276170.0    797.6      0.6                              + 1j * poles_cplx.real * (1 / denom2 - 1 / denom1)
   355                                                                   )
   356      1600    3182325.0   1989.0      1.4                  A_sub[:, A_sub_cplx_idx + 1] = coeff
   357                                           
   358                                                           # extra equation to avoid trivial solution:
   359                                                           # coeff += Re(1j / (s_k - pole) - 1j / (s_k - conj(pole)))
   360                                           
   361      1600     612946.0    383.1      0.3                  A_row_extra[A_sub_cplx_idx + 1] = np.sum(coeff.real, axis=0)
   362                                                           # part 3: coeff = -1 * H(s_k) * [1j / (s_k - pole) - 1j / (s_k - conj(pole))]
   363      1600    5553308.0   3470.8      2.4                  A_sub[:, A_sub_cplx_idx + 1 + n_cols_unused] = -coeff * freq_response[:, None]
   364                                           
   365                                                           # part 4: constant (variable d_res)
   366                                                           # coeff = -1 * H(s_k)
   367      1600      22726.0     14.2      0.0                  A_sub[:,-1] = - freq_response
   368                                           
   369                                                           # part 2: constant (variable d) and proportional term (variable e)
   370      1600       5779.0      3.6      0.0                  offset = n_cols_unused - 1
   371      1600       3097.0      1.9      0.0                  if fit_constant:
   372                                                               # coeff = 1 + j0
   373      1600      19541.0     12.2      0.0                      A_sub[:, offset] = 1
   374      1600       3803.0      2.4      0.0                      offset -=1
   375                                                               
   376      1600       2912.0      1.8      0.0                  if fit_proportional:
   377                                                               # coeff = s_k = j omega_k
   378                                                               A_sub[:, offset] = 1j * omega
   379                                                               
   380                                                           
   381      1600       6106.0      3.8      0.0                  A_row_extra[-1] = len(freqs_norm)
   382                                           
   383                                                           # QR decomposition
   384      1600      86654.0     54.2      0.0                  A_ri = np.empty((2 * A_sub.shape[0], A_sub.shape[1]))
   385      1600    2327198.0   1454.5      1.0                  A_ri[::2] = A_sub.real
   386      1600    1858140.0   1161.3      0.8                  A_ri[1::2] = A_sub.imag
   387                                           
   388      1600  174886563.0 109304.1     75.6                  Q, R = np.linalg.qr(A_ri, 'reduced')
   389                                           
   390                                                           # only R22 is required to solve for c_res and d_res
   391      1600      25923.0     16.2      0.0                  R22 = R[n_cols_unused:, n_cols_unused:]
   392                                           
   393                                                           # similarly, only right half of Q is required (not used here, because RHS is zero)
   394                                                           # Q2 = Q[:, n_cols_unused:]
   395                                           
   396                                                           # apply weight of this response and add coefficients to the system matrix
   397      1600     425804.0    266.1      0.2                  A[i_response * n_cols_used:(i_response + 1) * n_cols_used, :] = np.sqrt(weight_response) * R22
   398                                                           # multiplication of Q2 by rhs=0 omitted; right-hand side would also require weighting
   399                                                           # b[i_response * n_cols_used:(i_response + 1) * n_cols_used] = np.matmul(np.transpose(Q2), rhs)
   400                                           
   401                                                           # add extra equation to avoid trivial solution
   402      1600      75323.0     47.1      0.0                  weight_extra = np.linalg.norm(weight_response * freq_response) / len(freq_response)
   403      1600      15947.0     10.0      0.0                  A[(i_response + 1) * n_cols_used - 1, :] = np.sqrt(weight_extra) * A_row_extra
   404      1600       8797.0      5.5      0.0                  b[(i_response + 1) * n_cols_used - 1] = np.sqrt(weight_extra) * len(freq_response)
   405                                           
   406       100    5883020.0  58830.2      2.5              cond_A = np.linalg.cond(A)
   407       100       2486.0     24.9      0.0              logging.info('Condition number of coeff. matrix A = {}'.format(cond_A))
   408       100        402.0      4.0      0.0              self.history_cond_A.append(cond_A)
   409                                           
   410                                                       # solve least squares for real parts
   411       100    7141107.0  71411.1      3.1              x, residuals, rank, singular_vals = np.linalg.lstsq(A, b, rcond=None)
   412                                           
   413                                                       # assemble individual result vectors from single LS result x
   414       100        633.0      6.3      0.0              c_res = x[:-1]
   415       100        254.0      2.5      0.0              d_res = x[-1]
   416                                           
   417                                                       # check if d_res is suited for zeros calculation
   418       100        180.0      1.8      0.0              tol_res = 1e-8
   419       100       1100.0     11.0      0.0              if np.abs(d_res) < tol_res:
   420                                                           # d_res is too small, discard solution and proceed the |d_res| = tol_res
   421                                                           d_res = tol_res * (d_res / np.abs(d_res))
   422                                                           warnings.warn('Replacing d_res solution as it was too small. This is not a good sign and probably '
   423                                                                         'means that more starting poles are required', RuntimeWarning, stacklevel=2)
   424                                           
   425       100        347.0      3.5      0.0              self.d_res_history.append(d_res)
   426       100       2269.0     22.7      0.0              logging.info('d_res = {}'.format(d_res))
   427                                           
   428                                                       # build test matrix H, which will hold the new poles as eigenvalues
   429       100       5111.0     51.1      0.0              H = np.zeros((len(c_res), len(c_res)))
   430       100        250.0      2.5      0.0              i = 0
   431     20210      37618.0      1.9      0.0              for i_pole in range(len(poles)):
   432                                                           # fill diagonal with previous poles
   433     20110      44127.0      2.2      0.0                  pole_re = poles.real[i_pole]
   434     20110      42744.0      2.1      0.0                  pole_im = poles.imag[i_pole]
   435     20110      39899.0      2.0      0.0                  if pole_im == 0.0:
   436                                                               # one row for a real pole
   437       220        543.0      2.5      0.0                      H[i, i] = pole_re
   438       220       1196.0      5.4      0.0                      H[i] -= c_res / d_res
   439       220        467.0      2.1      0.0                      i += 1
   440                                                           else:
   441                                                               # two rows for a complex pole of a conjugated pair
   442     19890      40652.0      2.0      0.0                      H[i, i] = pole_re
   443     19890      41597.0      2.1      0.0                      H[i, i + 1] = pole_im
   444     19890      46634.0      2.3      0.0                      H[i + 1, i] = -1 * pole_im
   445     19890      41339.0      2.1      0.0                      H[i + 1, i + 1] = pole_re
   446     19890     147602.0      7.4      0.1                      H[i] -= 2 * c_res / d_res
   447     19890      38324.0      1.9      0.0                      i += 2
   448       100    6060767.0  60607.7      2.6              poles_new = np.linalg.eigvals(H)
   449                                           
   450                                                       # replace poles for next iteration
   451       100        681.0      6.8      0.0              poles_lst = []
   452                                           
   453     40100      79344.0      2.0      0.0              for k, pole in enumerate(poles_new):
   454     40000      87150.0      2.2      0.0                  if pole.imag >= 0.0:
   455                                                               # complex poles need to come in complex conjugate pairs; append only the positive part
   456     20111      39980.0      2.0      0.0                      poles_lst.append(pole)
   457                                           
   458       100       1717.0     17.2      0.0              poles = np.array(poles_lst)
   459                                                       # flip real part of unstable poles (real part needs to be negative for stability)
   460       100       1023.0     10.2      0.0              poles.real = - np.abs(poles.real)
   461                                           
   462                                           
   463                                                       # calculate relative changes in the singular values; stop iteration loop once poles have converged
   464       100       1840.0     18.4      0.0              new_max_singular = np.amax(singular_vals)
   465       100        703.0      7.0      0.0              delta_max = np.abs(1 - new_max_singular / max_singular)
   466       100        269.0      2.7      0.0              self.delta_max_history.append(delta_max)
   467       100       1562.0     15.6      0.0              logging.info('Max. relative change in residues = {}\n'.format(delta_max))
   468       100        195.0      1.9      0.0              max_singular = new_max_singular
   469                                           
   470       100        187.0      1.9      0.0              stop = False
   471       100        237.0      2.4      0.0              if delta_max < self.max_tol:
   472                                                           if converged:
   473                                                               # is really converged, finish
   474                                                               logging.info('Pole relocation process converged after {} iterations.'.format(
   475                                                                   self.max_iterations - iterations + 1))
   476                                                               stop = True
   477                                                           else:
   478                                                               # might be converged, but do one last run to be sure
   479                                                               converged = True
   480                                                       else:
   481       100        187.0      1.9      0.0                  if converged:
   482                                                               # is not really converged, continue
   483                                                               converged = False
   484                                           
   485       100        198.0      2.0      0.0              iterations -= 1
   486                                           
   487       100        197.0      2.0      0.0              if iterations == 0:
   488         1         22.0     22.0      0.0                  max_cond = np.amax(self.history_cond_A)
   489         1          2.0      2.0      0.0                  if max_cond > 1e10:
   490                                                               msg_illcond = 'Hint: the linear system was ill-conditioned (max. condition number = {}). ' \
   491                                                                             'This often means that more poles are required.'.format(max_cond)
   492                                                           else:
   493         1          2.0      2.0      0.0                      msg_illcond = ''
   494         1          2.0      2.0      0.0                  if converged and stop is False:
   495                                                               warnings.warn('Vector Fitting: The pole relocation process barely converged to tolerance. '
   496                                                                             'It took the max. number of iterations (N_max = {}). '
   497                                                                             'The results might not have converged properly. '.format(self.max_iterations)
   498                                                                             + msg_illcond, RuntimeWarning, stacklevel=2)
   499                                                           else:
   500         4        227.0     56.8      0.0                      warnings.warn('Vector Fitting: The pole relocation process stopped after reaching the '
   501                                                                             'maximum number of iterations (N_max = {}). '
   502         1          2.0      2.0      0.0                                    'The results did not converge properly. '.format(self.max_iterations)
   503         2          4.0      2.0      0.0                                    + msg_illcond, RuntimeWarning, stacklevel=2)
   504                                           
   505       100        185.0      1.9      0.0              if stop:
   506                                                           iterations = 0
   507                                           
   508                                                   # ITERATIONS DONE
   509         1          7.0      7.0      0.0          logging.info('Initial poles before relocation:')
   510         1          5.0      5.0      0.0          logging.info(self.initial_poles)
   511                                           
   512         1          4.0      4.0      0.0          logging.info('Final poles:')
   513         1         13.0     13.0      0.0          logging.info(poles * norm)
   514                                           
   515         1          4.0      4.0      0.0          logging.info('\n### Starting residues calculation process.\n')
   516                                           
   517                                                   # finally, solve for the residues with the previously calculated poles
   518         1          2.0      2.0      0.0          residues = []
   519         1          2.0      2.0      0.0          constant_coeff = []
   520         1          2.0      2.0      0.0          proportional_coeff = []
   521                                           
   522        17         47.0      2.8      0.0          for freq_response in freq_responses:
   523                                                       # calculate coefficients (row A_k in matrix) for each frequency sample s_k of the target response
   524                                                       # row will be appended to submatrix A_sub of complete coeff matrix A_matrix
   525                                                       # 2 rows per pole in result vector (1st for real part, 2nd for imaginary part)
   526                                                       # --> 2 columns per pole in coeff matrix
   527        16        665.0     41.6      0.0              n_cols = np.sum((poles.imag != 0) + 1)
   528                                           
   529        16         33.0      2.1      0.0              if fit_constant:
   530        16         44.0      2.8      0.0                  n_cols += 1
   531        16         32.0      2.0      0.0              if fit_proportional:
   532                                                           n_cols += 1
   533        16        116.0      7.2      0.0              A_matrix = np.empty((len(freqs_norm), n_cols), dtype=complex)
   534                                           
   535                                                       # Split up real and complex poles and store the correspondend column from A_sub
   536        16         93.0      5.8      0.0              real_mask = poles.imag == 0
   537        16        190.0     11.9      0.0              poles_real = poles[np.nonzero(real_mask)]
   538        16        153.0      9.6      0.0              poles_cplx = poles[np.nonzero(~real_mask)]
   539                                           
   540        16       6863.0    428.9      0.0              A_sub_real_idx = np.nonzero(self._get_real_pole_mask(poles))[0]
   541        16       6759.0    422.4      0.0              A_sub_cplx_idx = np.nonzero(~self._get_real_pole_mask(poles))[0][::2]
   542                                           
   543                                                       # add coefficients for a pair of complex conjugate poles
   544                                                       # part 1: first sum of rational functions (residue variable c)
   545                                           
   546        16        595.0     37.2      0.0              denom = poles_real.real ** 2 + omega[:, None] ** 2
   547        16       1337.0     83.6      0.0              A_matrix[:, A_sub_real_idx] = - (poles_real + 1j * omega[:, None]) / denom
   548                                           
   549                                                       # coefficient for real part of residue
   550        16       9061.0    566.3      0.0              denom1 = poles_cplx.real ** 2 + (omega[:, None] - poles_cplx.imag) ** 2
   551        16       8360.0    522.5      0.0              denom2 = poles_cplx.real ** 2 + (omega[:, None] + poles_cplx.imag) ** 2
   552        16      32837.0   2052.3      0.0              A_matrix[:, A_sub_cplx_idx] = (
   553        32      15134.0    472.9      0.0                  -1 * poles_cplx.real * (1 / denom1 + 1 / denom2)
   554        16      20668.0   1291.8      0.0                  + 1j * ((poles_cplx.imag - omega[:, None]) / denom1 - (poles_cplx.imag + omega[:, None]) / denom2)
   555                                                       )
   556                                                       
   557                                                       # coefficient for imaginary part of residue
   558        16      21890.0   1368.1      0.0              A_matrix[:, A_sub_cplx_idx + 1] = (
   559        32      21447.0    670.2      0.0                  (omega[:, None] - poles_cplx.imag) / denom1 - (poles_cplx.imag + omega[:, None]) / denom2
   560        16      12425.0    776.6      0.0                  + 1j * poles_cplx.real * (1 / denom2 - 1 / denom1)
   561                                                       )
   562                                           
   563        16        630.0     39.4      0.0              offset = np.sum((poles.imag != 0) + 1)
   564        16         36.0      2.2      0.0              if fit_constant:
   565        16        173.0     10.8      0.0                  A_matrix[:, offset] = 1
   566        16         51.0      3.2      0.0                  offset += 1
   567        16         32.0      2.0      0.0              if fit_proportional:
   568                                                           A_matrix[:, offset] = 1j * omega
   569                                           
   570        16     697922.0  43620.1      0.3              logging.info('A_matrix: condition number = {}'.format(np.linalg.cond(A_matrix)))
   571                                           
   572                                                       # solve least squares and obtain results as stack of real part vector and imaginary part vector
   573        16     702073.0  43879.6      0.3              x, residuals, rank, singular_vals = np.linalg.lstsq(np.vstack((A_matrix.real, A_matrix.imag)), np.hstack((freq_response.real, freq_response.imag)), rcond=None)
   574                                           
   575        16         54.0      3.4      0.0              i = 0
   576        16         35.0      2.2      0.0              residues_response = []
   577      3232       6638.0      2.1      0.0              for pole_im in poles.imag:
   578      3216       6763.0      2.1      0.0                  if pole_im == 0.0:
   579        32        146.0      4.6      0.0                      residues_response.append(x[i] + 0j)
   580        32         71.0      2.2      0.0                      i += 1
   581                                                           else:
   582      3184      15697.0      4.9      0.0                      residues_response.append(x[i] + 1j * x[i + 1])
   583      3184       6592.0      2.1      0.0                      i += 2
   584        16         33.0      2.1      0.0              residues.append(residues_response)
   585                                           
   586        16         33.0      2.1      0.0              if fit_constant and fit_proportional:
   587                                                           # both constant d and proportional e were fitted
   588                                                           constant_coeff.append(x[-2])
   589                                                           proportional_coeff.append(x[-1])
   590        16         31.0      1.9      0.0              elif fit_constant:
   591                                                           # only constant d was fitted
   592        16         38.0      2.4      0.0                  constant_coeff.append(x[-1])
   593        16         39.0      2.4      0.0                  proportional_coeff.append(0.0)
   594                                                       elif fit_proportional:
   595                                                           # only proportional e was fitted
   596                                                           constant_coeff.append(0.0)
   597                                                           proportional_coeff.append(x[-1])
   598                                                       else:
   599                                                           # neither constant d nor proportional e was fitted
   600                                                           constant_coeff.append(0.0)
   601                                                           proportional_coeff.append(0.0)
   602                                           
   603                                                   # save poles, residues, d, e in actual frequencies (un-normalized)
   604         1          9.0      9.0      0.0          self.poles = poles * norm
   605         1        191.0    191.0      0.0          self.residues = np.array(residues) * norm
   606         1          3.0      3.0      0.0          self.constant_coeff = np.array(constant_coeff)
   607         1         10.0     10.0      0.0          self.proportional_coeff = np.array(proportional_coeff) / norm
   608                                           
   609         1          3.0      3.0      0.0          timer_stop = timer()
   610         1          3.0      3.0      0.0          self.wall_clock_time = timer_stop - timer_start
   611                                           
   612         1         20.0     20.0      0.0          logging.info('\n### Vector fitting finished in {} seconds.\n'.format(self.wall_clock_time))
```

Vinc0110:
Interesting. Reducing the coefficient matrix with QR decomposition should give a big advancement for the subsequent least-squares solver. Regarding our previous conversation: I'm still not convinced the complex-valued matrix is not a problem. Maybe QR decomposition is such an example where it is much slower on complex-valued data? (Just an idea, I don't know). The docs don't seem to tell...

https://numpy.org/doc/stable/reference/generated/numpy.linalg.qr.html

FranzForstmayr:
The `linalg.qr` is called with a real only matrix. First line is real, second is imaginary, as before :)

skrf/vectorFitting.py
```diff
@@ -383,7 +383,7 @@ def vector_fit(self, n_poles_real: int = 2, n_poles_cmplx: int = 2, init_pole_sp
                 A_ri[::2] = A_sub.real
                 A_ri[1::2] = A_sub.imag
 
-                Q, R = np.linalg.qr(A_ri, 'reduced')
+                R = np.linalg.qr(A_ri, 'r')
```
Vinc0110, line 385: Good idea. Does it help much?

FranzForstmayr: About 40% reduction on my machine :)
Can solve the nasty port example in 2m 10s now.

Vinc0110:
> The `linalg.qr` is called with a real only matrix. First line is real, second is imaginary, as before :)

Oh, right.

By that way, why did you change the notebook? It seems like you embedded the plots as a png image. I would prefer to keep them, so we can use them as some sort of benchmark and testcase.

skrf/vectorFitting.py
Vinc0110: Nice!

FranzForstmayr:
> By that way, why did you change the notebook? It seems like you embedded the plots as a png image. I would prefer to keep them, so we can use them as some sort of benchmark and testcase.

Good catch, this was definitely not intended.

FranzForstmayr:
@Vinc0110 I'll try to split up the matrizes in re/im components again. 
I don't really understand, but the modification suggested [here](https://github.com/scikit-rf/scikit-rf/pull/605/files#r795988177) and similar ones slow down the code.

I'll need a few days more :)

Vinc0110:
I actually enjoy this kind of investigation, even though it can be annoying at times.

Are you saying this
```
denom1 = poles_cplx.real ** 2 + (omega[:, None] - poles_cplx.imag) ** 2
denom2 = poles_cplx.real ** 2 + (omega[:, None] + poles_cplx.imag) ** 2

coeff = (-1 * poles_cplx.real * (
1 / denom1 + 1 / denom2) + 
1j * ((poles_cplx.imag - omega[:, None]) / denom1 - (poles_cplx.imag + omega[:, None]) / denom2
))
```

is faster than this

`coeff = 1 / (1j * omega - poles_cplx) + 1 / (1j * omega - np.conj(poles_cplx))`

?

If so, those complex operations (conjugation, reciprocal, ...) might be more computationally heavy than you think (what I was getting at earlier in my other comments). I could still be wrong tough and the reason is something completely different...

FranzForstmayr:
Yes exactly, although it's not much, first guess is about 10s. However, the standard deviation is not small too, so will have to make a bunch of measurements, to find out if the difference is significant.

Sometimes it boils down to an optimization within numpy or the underlying math implementation. Maybe one specific optimization is unusable with complex numbers. I'll try to split up again :)

Vinc0110:
I made a few tests myself and I'm not sure what to think of the results:

Output of the script below:
```
recip_complex() took (7.48 +- 0.34) ms
recip_real1() took (18.29 +- 0.93) ms
recip_real2() took (18.39 +- 0.94) ms
coeff_complex() took (26.34 +- 1.38) ms
coeff_re_im() took (34.00 +- 1.94) ms
coeff_re() took (19.86 +- 1.44) ms
coeff_im() took (19.99 +- 1.36) ms
```
My conclusions:
- The direct calculation of `1/z` is faster than the analytic equation `1/z = (z' - jz'')/|z|²`
- The direct calculation of the complex coefficients in `coeff_complex()` is faster than the analytic complex equation in `coeff_re_im()` (26 ms versus 34 ms)
- However, the real-valued analytic equations for the individual real and imaginary parts are even faster with 20 ms each. Of course, they both need to be called (20 ms + 20 ms), but the total time probably depends on the implementation (serial versus parallel). I don't know what numpy is doing.

As you mentioned earlier, my results probably contain a lot of overhead for the function calls. So, I'm not sure they say anything about the behavior in the pole relocation algorithm.

Sorry for the large text, but github does not allow uploads of .py files...
```
import numpy as np
from timeit import default_timer as timer


def recip_complex(z):
    return 1 / z


def recip_real1(z):
    z_re = np.real(z)
    z_im = np.imag(z)
    denom = z_re ** 2 + z_im ** 2
    return (z_re + 1j * z_im) / denom

def recip_real2(z):
    denom = z.real ** 2 + z.imag ** 2
    return (z.real + 1j * z.imag) / denom


def coeff_complex(omega, pole):
    return 1 / (1j * omega - pole) + 1 / (1j * omega - np.conj(pole))


def coeff_re_im(omega, pole):
    denom1 = pole.real ** 2 + (omega - pole.imag) ** 2
    denom2 = pole.real ** 2 + (omega + pole.imag) ** 2
    return -1 * pole.real * (1 / denom1 + 1 / denom2) + \
            1j * ((pole.imag - omega) / denom1 - (pole.imag + omega) / denom2)


def coeff_re(omega, pole):
    denom1 = pole.real ** 2 + (omega - pole.imag) ** 2
    denom2 = pole.real ** 2 + (omega + pole.imag) ** 2
    return -1 * pole.real * (1 / denom1 + 1 / denom2)


def coeff_im(omega, pole):
    denom1 = pole.real ** 2 + (omega - pole.imag) ** 2
    denom2 = pole.real ** 2 + (omega + pole.imag) ** 2
    return (pole.imag - omega) / denom1 - (pole.imag + omega) / denom2


numbers_re = np.linspace(1, 1000, 1000)
numbers_im = np.linspace(-500, 500, 1000)

numbers = numbers_re[:, None] + 1j * numbers_im
omega = 2e9

n = 100
t_complex = np.empty(n)
t_real1 = np.empty(n)
t_real2 = np.empty(n)
t_coeffs_complex = np.empty(n)
t_coeffs_re_im = np.empty(n)
t_coeffs_re = np.empty(n)
t_coeffs_im = np.empty(n)
diffs = np.empty(n)
for i in range(n):
    t1 = timer()
    res_complex = recip_complex(numbers)
    t2 = timer()
    res_real1 = recip_real1(numbers)
    t3 = timer()
    res_real2 = recip_real2(numbers)
    t4 = timer()
    coeffs_complex = coeff_complex(omega, numbers)
    t5 = timer()
    coeffs_re_im = coeff_re_im(omega, numbers)
    t6 = timer()
    coeffs_re = coeff_re(omega, numbers)
    t7 = timer()
    coeffs_im = coeff_im(omega, numbers)
    t8 = timer()

    diffs[i] = np.average(np.abs(coeffs_re_im - coeffs_complex))

    t_complex[i] = t2 - t1
    t_real1[i] = t3 - t2
    t_real2[i] = t4 - t3
    t_coeffs_complex[i] = t5 - t4
    t_coeffs_re_im[i] = t6 - t5
    t_coeffs_re[i] = t7 - t6
    t_coeffs_im[i] = t8 - t7

print('recip_complex() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_complex) * 1e3, np.std(t_complex) * 1e3))
print('recip_real1() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_real1) * 1e3, np.std(t_real1) * 1e3))
print('recip_real2() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_real2) * 1e3, np.std(t_real2) * 1e3))
print('coeff_complex() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_coeffs_complex) * 1e3, np.std(t_coeffs_complex) * 1e3))
print('coeff_re_im() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_coeffs_re_im) * 1e3, np.std(t_coeffs_re_im) * 1e3))
#print('difference is {} +- {}'.format(np.average(diffs), np.std(diffs)))
print('coeff_re() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_coeffs_re) * 1e3, np.std(t_coeffs_re) * 1e3))
print('coeff_im() took ({:.2f} +- {:.2f}) ms'.format(np.average(t_coeffs_im) * 1e3, np.std(t_coeffs_im) * 1e3))
```

Vinc0110:
Oh and by the way, in the implementation for the pole relocation, the code in `coeff_re()` and `coeff_im()` would be not be separated and gets in the same run, so they can share the results in `denom1` and `denom2` (that's why I put them in separate variables). The results should not be much slower than 20 ms, I guess

FranzForstmayr:
I just benched the current commit adebae9b against the "Easier to read"-Implementation, and it seems there's not much difference.
I used this code:

```python
import numpy as np
import perfplot
import skrf as rf

import logging
import warnings
from timeit import default_timer as timer

class VectorFittingCplx(rf.VectorFitting):
    def vector_fit(self, n_poles_real: int = 2, n_poles_cmplx: int = 2, init_pole_spacing: str = 'lin',
                   parameter_type: str = 's', fit_constant: bool = True, fit_proportional: bool = False) -> None:
        """
        Main work routine performing the vector fit. The results will be stored in the class variables
        :attr:`poles`, :attr:`residues`, :attr:`proportional_coeff` and :attr:`constant_coeff`.

        Parameters
        ----------
        n_poles_real : int, optional
            Number of initial real poles. See notes.

        n_poles_cmplx : int, optional
            Number of initial complex conjugate poles. See notes.

        init_pole_spacing : str, optional
            Type of initial pole spacing across the frequency interval of the S-matrix. Either linear (lin) or
            logarithmic (log).

        parameter_type : str, optional
            Representation type of the frequency responses to be fitted. Either *scattering* (:attr:`s` or :attr:`S`),
            *impedance* (:attr:`z` or :attr:`Z`) or *admittance* (:attr:`y` or :attr:`Y`). As scikit-rf can currently
            only read S parameters from a Touchstone file, the fit should also be performed on the original S
            parameters. Otherwise, scikit-rf will convert the responses from S to Z or Y, which might work for the fit
            but can cause other issues.

        fit_constant : bool, optional
            Include a constant term **d** in the fit.

        fit_proportional : bool, optional
            Include a proportional term **e** in the fit.

        Returns
        -------
        None
            No return value.

        Notes
        -----
        The required number of real or complex conjugate starting poles depends on the behaviour of the frequency
        responses. To fit a smooth response such as a low-pass characteristic, 1-3 real poles and no complex conjugate
        poles is usually sufficient. If resonances or other types of peaks are present in some or all of the responses,
        a similar number of complex conjugate poles is required. Be careful not to use too many poles, as excessive
        poles will not only increase the computation workload during the fitting and the subsequent use of the model,
        but they can also introduce unwanted resonances at frequencies well outside the fit interval.
        """

        timer_start = timer()

        # create initial poles and space them across the frequencies in the provided Touchstone file
        # use normalized frequencies during the iterations (seems to be more stable during least-squares fit)
        norm = np.average(self.network.f)
        freqs_norm = np.array(self.network.f) / norm

        fmin = np.amin(freqs_norm)
        fmax = np.amax(freqs_norm)
        if init_pole_spacing == 'log':
            pole_freqs_real = np.geomspace(fmin, fmax, n_poles_real)
            pole_freqs_cmplx = np.geomspace(fmin, fmax, n_poles_cmplx)
        elif init_pole_spacing == 'lin':
            pole_freqs_real = np.linspace(fmin, fmax, n_poles_real)
            pole_freqs_cmplx = np.linspace(fmin, fmax, n_poles_cmplx)
        else:
            warnings.warn('Invalid choice of initial pole spacing; proceeding with linear spacing.', UserWarning,
                          stacklevel=2)
            pole_freqs_real = np.linspace(fmin, fmax, n_poles_real)
            pole_freqs_cmplx = np.linspace(fmin, fmax, n_poles_cmplx)

        # init poles array of correct length
        poles = np.zeros(n_poles_real + n_poles_cmplx, dtype=complex)

        # add real poles
        for i, f in enumerate(pole_freqs_real):
            omega = 2 * np.pi * f
            poles[i] = - omega

        # add complex-conjugate poles (store only positive imaginary parts)
        i_offset = len(pole_freqs_real)
        for i, f in enumerate(pole_freqs_cmplx):
            omega = 2 * np.pi * f
            poles[i_offset + i] = (-0.01 + 1j) * omega

        # save initial poles (un-normalize first)
        self.initial_poles = poles * norm
        max_singular = 1

        logging.info('### Starting pole relocation process.\n')

        # stack frequency responses as a single vector
        # stacking order (row-major):
        # s11, s12, s13, ..., s21, s22, s23, ...
        freq_responses = []
        for i in range(self.network.nports):
            for j in range(self.network.nports):
                if parameter_type.lower() == 's':
                    freq_responses.append(self.network.s[:, i, j])
                elif parameter_type.lower() == 'z':
                    freq_responses.append(self.network.z[:, i, j])
                elif parameter_type.lower() == 'y':
                    freq_responses.append(self.network.y[:, i, j])
                else:
                    warnings.warn('Invalid choice of matrix parameter type (S, Z, or Y); proceeding with scattering '
                                  'representation.', UserWarning, stacklevel=2)
                    freq_responses.append(self.network.s[:, i, j])
        freq_responses = np.array(freq_responses)

        # ITERATIVE FITTING OF POLES to the provided frequency responses
        # initial set of poles will be replaced with new poles after every iteration
        iterations = self.max_iterations
        self.d_res_history = []
        self.delta_max_history = []
        self.history_cond_A = []
        converged = False

        omega = 2 * np.pi * freqs_norm
        

        while iterations > 0:
            logging.info('Iteration {}'.format(self.max_iterations - iterations + 1))

            # count number of rows and columns in final coefficient matrix to solve for (c_res, d_res)
            # (ratio #real/#complex poles might change during iterations)

            n_cols_unused = np.sum((poles.imag != 0) + 1)

            n_cols_used = n_cols_unused
            n_cols_used += 1
            if fit_constant:
                n_cols_unused += 1
            if fit_proportional:
                n_cols_unused += 1
            n_rows_A = n_cols_used * len(freq_responses)

            # generate coefficients of approximation function for each target frequency response
            # responses will be reduced independently using QR decomposition
            # simplified coeff. matrices of all responses will be stacked in matrix A for least-squares solver
            A = np.empty((n_rows_A, n_cols_used))
            b = np.zeros(n_rows_A)

            for i_response, freq_response in enumerate(freq_responses):
                # calculate coefficients for each frequency response
                # A_sub which will be reduced first (QR decomposition) and then filled into the main
                # coefficient matrix A

                A_sub = np.empty((len(freqs_norm), n_cols_unused + n_cols_used), dtype=complex)
                A_row_extra = np.empty(n_cols_used)

                # responses will be weighted according to their norm;
                # alternative: equal weights with weight_response = 1.0
                # or anti-proportional weights with weight_response = 1 / np.linalg.norm(freq_response)
                weight_response = np.linalg.norm(freq_response)

                # Split up real and complex poles and store the correspondend column from A_sub
                real_mask = poles.imag == 0
                poles_real = poles[np.nonzero(real_mask)]
                poles_cplx = poles[np.nonzero(~real_mask)]

                A_sub_real_idx = np.nonzero(self._get_real_pole_mask(poles))[0]
                A_sub_cplx_idx = np.nonzero(~self._get_real_pole_mask(poles))[0][::2]


                # add coefficients for a pair of complex conjugate poles
                # part 1: first sum of rational functions (residue variable c)
                # merged with
                # part 3: second sum of rational functions (variable c_res)
                coeff = 1 / (1j * omega[:, None] - poles_real)

                # part 1: coeff = 1 / (s_k - p') = coeff_re + j coeff_im
                A_sub[:, A_sub_real_idx] = coeff

                # part 3: coeff = -1 * H(s_k) / (s_k - pole)
                # Re{coeff} = -1 * coeff_re * resp_re + coeff_im * resp_im
                # Im{coeff} = -1 * coeff_re * resp_im - coeff_im * resp_re
                A_sub[:, A_sub_real_idx + n_cols_unused] = - coeff * freq_response[:, None]

                # extra equation to avoid trivial solution:
                # coeff += Re(1 / (s_k - pole)) = coeff_re
                A_row_extra[A_sub_real_idx] = np.sum(coeff.real, axis=0)

                # coefficient for a complex pole of a conjugated pair: p = p' + jp''

                # row 1: add coefficient for real part of residue
                # part 1: coeff = 1 / (s_k - pole) + 1 / (s_k - conj(pole))
                coeff = 1 / (1j * omega[:, None] - poles_cplx) + 1 / (1j * omega[:, None] - np.conj(poles_cplx))
                A_sub[:, A_sub_cplx_idx] = coeff

                # extra equation to avoid trivial solution:
                # coeff += Re{1 / (s_k - pole) + 1 / (s_k - conj(pole))}

                A_row_extra[A_sub_cplx_idx] = np.sum(coeff.real, axis=0)
                # part 3: coeff = -1 * H(s_k) * [1 / (s_k - pole) + 1 / (s_k - conj(pole))]
                A_sub[:, A_sub_cplx_idx + n_cols_unused] = - coeff * freq_response[:, None]

                # part 1: coeff = 1j / (s_k - pole) - 1j / (s_k - conj(pole))
                coeff = 1j / (1j * omega[:, None] - poles_cplx) - 1j / (1j * omega[:, None] - np.conj(poles_cplx))
                A_sub[:, A_sub_cplx_idx + 1] = coeff

                # extra equation to avoid trivial solution:
                # coeff += Re(1j / (s_k - pole) - 1j / (s_k - conj(pole)))

                A_row_extra[A_sub_cplx_idx + 1] = np.sum(coeff.real, axis=0)
                # part 3: coeff = -1 * H(s_k) * [1j / (s_k - pole) - 1j / (s_k - conj(pole))]
                A_sub[:, A_sub_cplx_idx + 1 + n_cols_unused] = -coeff * freq_response[:, None]

                # part 4: constant (variable d_res)
                # coeff = -1 * H(s_k)
                A_sub[:,-1] = - freq_response

                # part 2: constant (variable d) and proportional term (variable e)
                offset = n_cols_unused - 1
                if fit_constant:
                    # coeff = 1 + j0
                    A_sub[:, offset] = 1
                    offset -=1
                    
                if fit_proportional:
                    # coeff = s_k = j omega_k
                    A_sub[:, offset] = 1j * omega
                    
                
                A_row_extra[-1] = len(freqs_norm)

                # QR decomposition
                A_ri = np.empty((2 * A_sub.shape[0], A_sub.shape[1]))
                A_ri[::2] = A_sub.real
                A_ri[1::2] = A_sub.imag

                R = np.linalg.qr(A_ri, 'r')

                # only R22 is required to solve for c_res and d_res
                R22 = R[n_cols_unused:, n_cols_unused:]

                # similarly, only right half of Q is required (not used here, because RHS is zero)
                # Q2 = Q[:, n_cols_unused:]

                # apply weight of this response and add coefficients to the system matrix
                A[i_response * n_cols_used:(i_response + 1) * n_cols_used, :] = np.sqrt(weight_response) * R22
                # multiplication of Q2 by rhs=0 omitted; right-hand side would also require weighting
                # b[i_response * n_cols_used:(i_response + 1) * n_cols_used] = np.matmul(np.transpose(Q2), rhs)

                # add extra equation to avoid trivial solution
                weight_extra = np.linalg.norm(weight_response * freq_response) / len(freq_response)
                A[(i_response + 1) * n_cols_used - 1, :] = np.sqrt(weight_extra) * A_row_extra
                b[(i_response + 1) * n_cols_used - 1] = np.sqrt(weight_extra) * len(freq_response)

            cond_A = np.linalg.cond(A)
            logging.info('Condition number of coeff. matrix A = {}'.format(cond_A))
            self.history_cond_A.append(cond_A)

            # solve least squares for real parts
            x, residuals, rank, singular_vals = np.linalg.lstsq(A, b, rcond=None)

            # assemble individual result vectors from single LS result x
            c_res = x[:-1]
            d_res = x[-1]

            # check if d_res is suited for zeros calculation
            tol_res = 1e-8
            if np.abs(d_res) < tol_res:
                # d_res is too small, discard solution and proceed the |d_res| = tol_res
                d_res = tol_res * (d_res / np.abs(d_res))
                warnings.warn('Replacing d_res solution as it was too small. This is not a good sign and probably '
                              'means that more starting poles are required', RuntimeWarning, stacklevel=2)

            self.d_res_history.append(d_res)
            logging.info('d_res = {}'.format(d_res))

            # build test matrix H, which will hold the new poles as eigenvalues
            H = np.zeros((len(c_res), len(c_res)))
            i = 0
            for i_pole in range(len(poles)):
                # fill diagonal with previous poles
                pole_re = poles.real[i_pole]
                pole_im = poles.imag[i_pole]
                if pole_im == 0.0:
                    # one row for a real pole
                    H[i, i] = pole_re
                    H[i] -= c_res / d_res
                    i += 1
                else:
                    # two rows for a complex pole of a conjugated pair
                    H[i, i] = pole_re
                    H[i, i + 1] = pole_im
                    H[i + 1, i] = -1 * pole_im
                    H[i + 1, i + 1] = pole_re
                    H[i] -= 2 * c_res / d_res
                    i += 2
            poles_new = np.linalg.eigvals(H)

            # replace poles for next iteration
            poles_lst = []

            for k, pole in enumerate(poles_new):
                if pole.imag >= 0.0:
                    # complex poles need to come in complex conjugate pairs; append only the positive part
                    poles_lst.append(pole)

            poles = np.array(poles_lst)
            # flip real part of unstable poles (real part needs to be negative for stability)
            poles.real = - np.abs(poles.real)


            # calculate relative changes in the singular values; stop iteration loop once poles have converged
            new_max_singular = np.amax(singular_vals)
            delta_max = np.abs(1 - new_max_singular / max_singular)
            self.delta_max_history.append(delta_max)
            logging.info('Max. relative change in residues = {}\n'.format(delta_max))
            max_singular = new_max_singular

            stop = False
            if delta_max < self.max_tol:
                if converged:
                    # is really converged, finish
                    logging.info('Pole relocation process converged after {} iterations.'.format(
                        self.max_iterations - iterations + 1))
                    stop = True
                else:
                    # might be converged, but do one last run to be sure
                    converged = True
            else:
                if converged:
                    # is not really converged, continue
                    converged = False

            iterations -= 1

            if iterations == 0:
                max_cond = np.amax(self.history_cond_A)
                if max_cond > 1e10:
                    msg_illcond = 'Hint: the linear system was ill-conditioned (max. condition number = {}). ' \
                                  'This often means that more poles are required.'.format(max_cond)
                else:
                    msg_illcond = ''
                if converged and stop is False:
                    warnings.warn('Vector Fitting: The pole relocation process barely converged to tolerance. '
                                  'It took the max. number of iterations (N_max = {}). '
                                  'The results might not have converged properly. '.format(self.max_iterations)
                                  + msg_illcond, RuntimeWarning, stacklevel=2)
                else:
                    warnings.warn('Vector Fitting: The pole relocation process stopped after reaching the '
                                  'maximum number of iterations (N_max = {}). '
                                  'The results did not converge properly. '.format(self.max_iterations)
                                  + msg_illcond, RuntimeWarning, stacklevel=2)

            if stop:
                iterations = 0

        # ITERATIONS DONE
        logging.info('Initial poles before relocation:')
        logging.info(self.initial_poles)

        logging.info('Final poles:')
        logging.info(poles * norm)

        logging.info('\n### Starting residues calculation process.\n')

        # finally, solve for the residues with the previously calculated poles
        residues = []
        constant_coeff = []
        proportional_coeff = []

        for freq_response in freq_responses:
            # calculate coefficients (row A_k in matrix) for each frequency sample s_k of the target response
            # row will be appended to submatrix A_sub of complete coeff matrix A_matrix
            # 2 rows per pole in result vector (1st for real part, 2nd for imaginary part)
            # --> 2 columns per pole in coeff matrix
            n_cols = np.sum((poles.imag != 0) + 1)

            if fit_constant:
                n_cols += 1
            if fit_proportional:
                n_cols += 1
            A_matrix = np.empty((len(freqs_norm), n_cols), dtype=complex)

            # Split up real and complex poles and store the correspondend column from A_sub
            real_mask = poles.imag == 0
            poles_real = poles[np.nonzero(real_mask)]
            poles_cplx = poles[np.nonzero(~real_mask)]

            A_sub_real_idx = np.nonzero(self._get_real_pole_mask(poles))[0]
            A_sub_cplx_idx = np.nonzero(~self._get_real_pole_mask(poles))[0][::2]

            # add coefficients for a pair of complex conjugate poles
            # part 1: first sum of rational functions (residue variable c)

            A_matrix[:, A_sub_real_idx] = 1 / (1j * omega[:, None] - poles_real)

            # coefficient for real part of residue
            A_matrix[:, A_sub_cplx_idx] = (1 / (1j * omega[:, None] - poles_cplx) + 
                1 / (1j * omega[:, None] - np.conj(poles_cplx)))

            # coefficient for imaginary part of residue
            A_matrix[:, A_sub_cplx_idx + 1] = (1j / (1j * omega[:, None] - poles_cplx) 
                - 1j / (1j * omega[:, None] - np.conj(poles_cplx)))

            offset = np.sum((poles.imag != 0) + 1)
            if fit_constant:
                A_matrix[:, offset] = 1
                offset += 1
            if fit_proportional:
                A_matrix[:, offset] = 1j * omega

            logging.info('A_matrix: condition number = {}'.format(np.linalg.cond(A_matrix)))

            # solve least squares and obtain results as stack of real part vector and imaginary part vector
            x, residuals, rank, singular_vals = np.linalg.lstsq(np.vstack((A_matrix.real, A_matrix.imag)), np.hstack((freq_response.real, freq_response.imag)), rcond=None)

            i = 0
            residues_response = []
            for pole_im in poles.imag:
                if pole_im == 0.0:
                    residues_response.append(x[i] + 0j)
                    i += 1
                else:
                    residues_response.append(x[i] + 1j * x[i + 1])
                    i += 2
            residues.append(residues_response)

            if fit_constant and fit_proportional:
                # both constant d and proportional e were fitted
                constant_coeff.append(x[-2])
                proportional_coeff.append(x[-1])
            elif fit_constant:
                # only constant d was fitted
                constant_coeff.append(x[-1])
                proportional_coeff.append(0.0)
            elif fit_proportional:
                # only proportional e was fitted
                constant_coeff.append(0.0)
                proportional_coeff.append(x[-1])
            else:
                # neither constant d nor proportional e was fitted
                constant_coeff.append(0.0)
                proportional_coeff.append(0.0)

        # save poles, residues, d, e in actual frequencies (un-normalized)
        self.poles = poles * norm
        self.residues = np.array(residues) * norm
        self.constant_coeff = np.array(constant_coeff)
        self.proportional_coeff = np.array(proportional_coeff) / norm

        timer_stop = timer()
        self.wall_clock_time = timer_stop - timer_start

        logging.info('\n### Vector fitting finished in {} seconds.\n'.format(self.wall_clock_time))




nw = rf.network.Network('default.s4p')
vf_opig = rf.VectorFitting(nw)
vf_cplx = VectorFittingCplx(nw)

out = perfplot.bench(
    setup=lambda n: n,
    kernels=[
        lambda n: vf_opig.vector_fit(n_poles_real = 0, n_poles_cmplx = n),
        lambda n: vf_cplx.vector_fit(n_poles_real = 0, n_poles_cmplx = n),
    ],
    labels=["vector_fit", "vector_fit_cplx"],
    n_range=np.arange(10,300,10),
    xlabel="Number of complex poles",
    # More optional arguments with their default values:
    # logx="auto",  # set to True or False to force scaling
    # logy="auto",
    equality_check=None,  # set to None to disable "correctness" assertion
    # show_progress=True,
    # target_time_per_measurement=1.0,
    # max_time=None,  # maximum time per measurement
    # time_unit="s",  # set to one of ("auto", "s", "ms", "us", or "ns") to force plot units
    # relative_to=1,  # plot the timings relative to one of the measurements
    # flops=lambda n: 3*n,  # FLOPS plots
)

out.show()
out.save("perf.png", transparent=True, bbox_inches="tight")
```

And this is the result. The implementation with the complex calculation tends to be a little bit slower, although the difference is really small.


perf

FranzForstmayr:
> * The direct calculation of `1/z` is faster than the analytic equation `1/z = (z' - jz'')/|z|²`
> * The direct calculation of the complex coefficients in `coeff_complex()` is faster than the analytic complex equation in `coeff_re_im()` (26 ms versus 34 ms)

This is probably due to some internal optimization within numpy or even deeper in hierarchy. Some experts implemented an efficient solution to calculate `1/z` if `z` is a complex number, which is faster than every _hand made_ algorithm in python.

Vinc0110:
> I just benched the current commit https://github.com/scikit-rf/scikit-rf/commit/adebae9b416f63827f19939bb81455e744906d23 against the "Easier to read"-Implementation, and it seems there's not much difference.

That's good. Then it's much better to use the simple expressions for the coefficients.

The plot also looks nice, but it seems the action really only starts after the plateau around the 200-poles mark. Maybe with much less than 200 poles the results are just garbage? At the end, the runtime alone is not saying much, the fit also has to be successful. Minor sidetrack: it would be very interesting to also record and plot the number of iterations needed. Do they all run into the limit at 100 iterations, even with 250 poles?

> This is probably due to some internal optimization within numpy or even deeper in hierarchy. Some experts implemented an efficient solution to calculate 1/z if z is a complex number, which is faster than every hand made algorithm in python.

And what about the test results with the real-valued equations in `coeff_re()` and `coeff_im()` for the real and imaginary parts of the coefficients? They are even faster. In general, most operations are much more intensive if the data is complex-valued, even when the routines are optimized. I think that's true for both numeric calculations and manual analysis, isn't it?

Just a few more ideas (maybe for another PR if you have enough of it):
- Compare the performance of a real-valued coefficients matrix with the results you got with the stacking of `real()` and `imag()` of the complex matrix.
- Would it make a difference to define `s = 1j * omega` instead of `omega`? Because **all** equations use `1j * omega[:, None]`, so maybe there is a little bit to gain here.
- You merged 2 of 3 loops; what about the last one? Can we also stack the frequency responses beforehand and run QR decomposition on the entire matrix instead of decomposing the individual sub-matrices and stacking them? The final matrix `A` should be the same in both cases, correct?

FranzForstmayr:
> The plot also looks nice, but it seems the action really only starts after the plateau around the 200-poles mark. Maybe with much less than 200 poles the results are just garbage? At the end, the runtime alone is not saying much, the fit also has to be successful. Minor sidetrack: it would be very interesting to also record and plot the number of iterations needed. Do they all run into the limit at 100 iterations, even with 250 poles?

The fit _should_ be as good as the current implementation, as there's no mathemathical difference. I'll check this of course before finalizing this PR, if the result is exactly the same.

> And what about the test results with the real-valued equations in `coeff_re()` and `coeff_im()` for the real and imaginary parts of the coefficients? They are even faster. In general, most operations are much more intensive if the data is complex-valued, even when the routines are optimized. I think that's true for both numeric calculations and manual analysis, isn't it?

Yeah, think so too. I just thought that the difference isn't that big.

> Would it make a difference to define s = 1j * omega instead of omega? Because all equations use 1j * omega[:, None], so maybe there is a little bit to gain here.

This is a great idea

> You merged 2 of 3 loops; what about the last one? Can we also stack the frequency responses beforehand and run QR decomposition on the entire matrix instead of decomposing the individual sub-matrices and stacking them? The final matrix A should be the same in both cases, correct?

Probably yes, I'll have to have a more detailed look.

FranzForstmayr:
I have to admit, I'm not sure how you would like to stack the responses before `qr`...

FranzForstmayr:
@Vinc0110 I think I'll stop for now, but I'm happy to help in future with possible optimizations. Most of the time is spent in `qr`, so there's probably only a small improvement, if we optimize some small array calculations. Thanks for your help!

skrf/vectorFitting.py
```diff
@@ -276,110 +288,101 @@ def vector_fit(self, n_poles_real: int = 2, n_poles_cmplx: int = 2, init_pole_sp
                 # calculate coefficients for each frequency response
                 # A_sub which will be reduced first (QR decomposition) and then filled into the main
                 # coefficient matrix A
-                # each frequency sample s_k of the target response generates two rows in the matrix
-                # (1st for real part, 2nd for imaginary part)
 
-                A_sub = np.empty((2 * len(freqs_norm), n_cols_unused + n_cols_used))
-                A_row_extra = np.zeros(n_cols_used)
+                A_sub = np.empty((len(freqs_norm), n_cols_unused + n_cols_used), dtype=complex)
+                A_row_extra = np.empty(n_cols_used)
```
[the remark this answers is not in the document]
FranzForstmayr: I filled with np.nan and it still works. The indexing scheme should make sure, that every column is assigned properly.

Vinc0110 (approved):
Thanks a lot for the many changes, tests, and your patience! The results speak for themselves. Let's leave it like this for now.

Update of my local tests with the old version (`master`), the initial state of this PR a93d169 and the current/final state 438162a:
- Example4 (1+26): `master`: 19.6 s; a93d169: 1.1 s; 438162a: 0.79 s
- Example4 (1+25): `master`: 14.9 s; a93d169: 1.0 s; 438162a: 0.6 s
- Example2 (4+4): `master`: 4.8 s; a93d169: 0.3 s; 438162a: 0.2 s
- Example2 (3+4): `master`: 2.2 s; a93d169: 0.16 s; 438162a: 0.11 s
- Nasty 4-port (0+200): `master`: 2151 s; a93d169: 249 s; 438162a: 176 s

jhillairet:
OK to merge then?

Vinc0110:
I think so.

Issue #611, opened 2022-02-27
Incorrect Type Hint for Network.f 


lstuemke:
Can we fix the type hint for the frequency getter "f" of the Network class?

![image](https://user-images.githubusercontent.com/49922901/155890316-78aeada1-8d5f-4e33-820a-828a6d0e4f88.png)

This leads to incorrect type warnings in IDEs like these:

![image](https://user-images.githubusercontent.com/49922901/155890379-1d36b0ee-7746-4094-a1d8-4cfc251aaf4b.png)

![image](https://user-images.githubusercontent.com/49922901/155890451-fd20c387-8090-4722-99f1-16d94bc8b6e5.png)

jhillairet:
Closed with #612. Thank you.

PR #612, opened 2022-02-28
Correct Network.f property return type-hint

lstuemke:
This is to address the issue I opened here: https://github.com/scikit-rf/scikit-rf/issues/611


jhillairet:
Thank you for your contribution and welcome as a contributor!

jhillairet (approved):

PR #620, opened 2022-03-19
Fix#618 scipy.ndimage.filters namespace deprecation warning

l-johnston:
- removed scipy.ndimage.convolve1d from skrf.network since it was not being used
- updated skrf.time to import from scipy.ndimage


jhillairet:
Thank you @l-johnston for fixing #620 and welcome to the scikit-rf community !

PR #621, opened 2022-03-30
Bug Fix for Parser.py yaml.load deprecated

xreymason:
yaml.load(stream) is deprecated due to it being unsafe, must use yaml.Loader to fix.

### DIFF
--- a/skrf/vi/scpi/parser.py
+++ b/skrf/vi/scpi/parser.py
@@ -298,7 +298,7 @@ def parse_yaml_file(driver_yaml_file):
 
     driver_template = None
     with open(driver_yaml_file, 'r', encoding='utf-8') as yaml_file:
-        driver_template = yaml.load(yaml_file)
+        driver_template = yaml.load(yaml_file, Loader=yaml.Loader)
 
     sets, queries = parse_branch(driver_template["COMMAND_TREE"])
 


jhillairet:
Thank you @xreymason for your contribution and welcome !

xreymason:
Thank you for taking a look. Love this package as a whole and I work with instrumentation automation for work, so this particular tool has come in handy since I've found it and made this modification.

jhillairet:
> Thank you for taking a look. Love this package as a whole and I work with instrumentation automation for work, so this particular tool has come in handy since I've found it and made this modification.

I would love to hear more about your work and how you automate things. Please email me!

skrf/vi/scpi/parser.py
```diff
@@ -298,7 +298,7 @@ def parse_yaml_file(driver_yaml_file):
 
     driver_template = None
     with open(driver_yaml_file, 'r', encoding='utf-8') as yaml_file:
-        driver_template = yaml.load(yaml_file)
+        driver_template = yaml.load(yaml_file, Loader=yaml.Loader)
```
jhillairet, line 301: would it give the same result using :
```suggestion
        driver_template = yaml.safe_load(yaml_file)
```
?
(https://pyyaml.docsforge.com/master/api/yaml/safe_load/)
If yes, this is maybe more explicit.

xreymason: That works but does create some potential issues with truncating zeros and possibly other background safeguards.

Question about this about the functionality you want users to have:
Should it be able to run python code when loading the .yaml file?
If not, then we might want to use the following:
`driver_template = yaml.load(yaml_file, Loader=yaml.BaseLoader)`
Otherwise, my original change using `yaml.Loader` will work best.

By default, the `yaml.load` methods Loader was `yaml.Loader` before this deprecation happened. `yaml.Loader` facilitates the execution of arbitrary Python code, `yaml.BaseLoader` doesn't allow for the execution of Python code. `yaml.SafeLoader` (which is used in `yaml.safe_load()`, doesn't allow code execution and has some other safeties I haven't looked into yet. The SafeLoader will also restrict your yaml tags to a degree which might limit your use cases.

jhillairet: > That works but does create some potential issues with truncating zeros and possibly other background safeguards.

OK, fine. Let's keep it like that then.

jhillairet:
Thank you !

xreymason:
No problem, glad I could help!

PR #624, opened 2022-04-06
Fixing SOLT description and add a SOLT calibration example 

jhillairet:
- the SOLT calibration example in the tutorial was in the uncorrect order (Thru should be the last element in the list of Networks)
- Added a SOLT example and some figures


arsenovic:
actually there is a conceptual error here . i found when i tried to add some  mismatch.  

see the lines below. there is a need for line2 to be `flipped` when embedding the right hand side.  it matters if the line2 is not symetric.   (note there is also an `embed` function for Calibration, but i like your example it is more transparent .)

```
line1 = media.line(d=20, unit='cm')**media.impedance_mismatch(1,2)
line2 = media.line(d=30, unit='cm')**media.impedance_mismatch(1,3)

# add some noise to make it more realistic
line1.add_noise_polar(.01, .1)
line2.add_noise_polar(.01, .1)

# fake the measured setup 
measured = line1 ** dut  ** line2

# fake the calibration measurements
open_measured = rf.two_port_reflect(line1 ** media.open(), line2.flipped() ** media.open())
short_measured = rf.two_port_reflect(line1 ** media.short(), line2.flipped() ** media.short())
load_measured = rf.two_port_reflect(line1 ** media.load(Gamma0=0), line2.flipped() ** media.load(Gamma0=0))```

Issue #521, opened 2021-09-16
Implementation of IEEE P370 De-embedding Method into De-embedding Class in Scikit-RF?


denzchoe:
As indicated at https://groups.google.com/g/scikit-rf/c/vwVKe8JaCtU

Hi all,

I have been observing this standard for quite some time from its early days. 
https://www.signalintegrityjournal.com/articles/1169-ieee-p370-a-ffixture-ddesign-and-ddata-qquality-mmetric-sstandard-for-iinterconnects-up-to-50-ghz?page=2

And only recently that it has been released as a Standard ( Published Date:2021-01-08  )
https://standards.ieee.org/standard/370-2020.html

There is an IEEE Committee of SI Engineers working in parallel like you guys over here to push their own version of the characterization & de-embedding Class for SI Community to implement using MATLAB. 

For those who doesn't own MATLAB, you can excess the .exe program over here (although it may not be the latest version):
https://www.amphenol-icc.com/software?__cf_chl_jschl_tk__=pmd_q5VhubKX2D37YIlaKn7dUKIgiBuzm0PVl3D4cWYdutA-1631516714-0-gqNtZGzNAhCjcnBszQ29

![image](https://user-images.githubusercontent.com/26157143/133537606-03c961ec-c863-48d5-a841-1c094b2c9224.png)

Fortunately the codes are open sourced and constantly updated.
https://opensource.ieee.org/elec-char/ieee-370/

Looking that now we have a De-Embedding Class inside scikit-rf (thanks to [Vikram](https://github.com/vik-s)); perhaps we can now consider adding the IEEE P370 de-embedding method into scikit-rf?
https://opensource.ieee.org/elec-char/ieee-370/-/tree/master/TG1
I have not tested it extensively enough yet (so far so good on just 2 samples until 70 GHz, no real world measurement yet) but looking at how some of the big Connectors, measurement & EDA companies are already member of this group, is hard to ignore its relevance.
https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9241568

We should also probably invite the Gitlab Code Owners over here to have a look at scikit-rf. I have had the opportunity to bump into Jason Ellison at Y2019 in a conference and spoke to him a little bit of scikit-rf, as a conference attendee after his presentation then. But.. I doubt my words carried any weight. However, some of you fellas here definitely have the Credentials to do so. 

[...]

PR #628, opened 2022-04-12
Implement IEEEP370 NZC and ZC deembedding methods

mhuser:
Implement IEEEP370 deembedding methods as discussed in issue #521.

- `Ieeep370nzc2xthru` does create left and right fixtures by bisecting a 2xthru without taking care if the impedance is not Z0 ([original ieee sources for matlab](https://opensource.ieee.org/elec-char/ieee-370/-/blob/master/TG1/IEEEP3702xThru.m))
- `Ieeep370zc2xthru` does create left and right fixtures by bisecting a 2xthru and use a fixture-dut-fixture to compensate for the actual impedance ([original ieee sources for matlab](https://opensource.ieee.org/elec-char/ieee-370/-/blob/master/TG1/IEEEP370Zc2xThru.m))

The sources were ported into python/scikit-rf and could be some more optimized (e.g. by using `media `and `lines` instead of recomputing lines mathematically in the class)


mhuser:
Still TODO:

- Add a comparison with fixture generated with IEEE tool in notebook
- Add a test case with the same data as notebook asserting consistency with IEEE tool generated fixtures

skrf/calibration/deembedding.py
```diff
@@ -908,4 +912,834 @@ def deembed(self, ntwk):
...
         caled.z = (h.z + h_.z) / 2
 
-        return caled
\ No newline at end of file
+        return caled
+
+
+class Ieeep370nzc2xthru(Deembedding):
+    """
+    IEEEP3702xThru.m creates error boxes from a test fixture 2x thru.
```
jhillairet, line 920:
```suggestion
    Create error boxes from a test fixture 2x thru.
```

skrf/calibration/deembedding.py
```diff
@@ -908,4 +912,834 @@ def deembed(self, ntwk):
...
+    
+    Based on https://opensource.ieee.org/elec-char/ieee-370/-/blob/master/TG1/IEEEP3702xThru.m
+    commit 49ddd78cf68ad5a7c0aaa57a73415075b5178aa6
+    See github issue https://github.com/scikit-rf/scikit-rf/issues/521
```
jhillairet, line 924: I'm not sure if it is useful to refer to a GitHub issue in the docstring.
```suggestion
```

skrf/calibration/deembedding.py
```diff
@@ -908,4 +912,834 @@ def deembed(self, ntwk):
...
+        receiver filter in COM defined by eq 93A-20
+        """ 
+        fdfr = f / fr
+        # eq 93A-20
+        return 1 / (1 - 3.414214 * fdfr**2 + fdfr**4 + 1j*2.613126*(fdfr - fdfr**3))
+    
+    
+    def makeStep(self, impulse):
+        """
+        mhuser : no need to call step function here, cumsum will be enough and efficient
```
jhillairet, line 1029: I'd rather put this comment into a comment in a code rather than in the docstring.

skrf/calibration/deembedding.py
```diff
@@ -908,4 +912,834 @@ def deembed(self, ntwk):
...
+        s_fixture_model_r2.renormalize(50)
+        s_side1 = s_fixture_model_r1
+        s_side2 = s_fixture_model_r2
+        
+        return (s_side1, s_side2)
+    
+class Ieeep370zc2xthru(Deembedding):
+    """
+    IEEEP370Zc2xThru.m creates error boxes from a test fixture 2x thru and the
+    fixture-dut-fixture S-parameters.
```
jhillairet, line 1210:
```suggestion
    Creates error boxes from a test fixture 2x thru and the fixture-dut-fixture S-parameters.

```

jhillairet: this is also the case for the previous function, isn't it? What is the difference then?

jhillairet (commented):
great work! I've made some minor comments to improve the docstring.

skrf/calibration/deembedding.py
mhuser: Done

skrf/calibration/deembedding.py
mhuser: Removed

skrf/calibration/deembedding.py
mhuser: Commented

skrf/calibration/deembedding.py
mhuser: - NZC take only fixture-fixture as input and bisect it.
- ZC take fixture-fixture and fixture-dut-fixture as inputs and accound for a possible impedance variation of the tru of boths.

mhuser:
> great work! I've made some minor comments to improve the docstring.

@jhillairet Thanks you a lot !
I have finished the example and applied your comments to docstrings. Only test case missing now.

mhuser:
> Out of curiosity, why was there a factor 2 in here the first place? I saw the similar thing in your earlier code (fork in your account before PR). Genuinely curious

I was testing against the old version of AICC tools that gave non-consistent results with NCZ results. I wondered if my cumsum method to build timesteps was 2 time shorter than convolution methods and later I finished to add factor 2 at many place but very wrongly seeing that the results were coherent with the old AICC tool.
After your comment I tested the latest AICC tool and the results were slightly differents but the factor 2 were everywhere and the harm was done.

denzchoe:
> > Out of curiosity, why was there a factor 2 in here the first place? I saw the similar thing in your earlier code (fork in your account before PR). Genuinely curious
> 
> I was testing against the old version of AICC tools that gave non-consistent results with NCZ results. I wondered if my cumsum method to build timesteps was 2 time shorter than convolution methods and later I finished to add factor 2 at many place but very wrongly seeing that the results were coherent with the old AICC tool. After your comment I tested the latest AICC tool and the results were slightly differents but the factor 2 were everywhere and the harm was done.

Thanks for the explanation!

denzchoe:
Hmmm.. From the perspective of readability, it's a little difficult to read. It's subjective to my opinion of course. 

Also if we're following actual IEEE standards (even though clone on Python), how close should our API names be? (perhaps for future cross-check)?

Hopefully others could chime in on their opinions too, like @jhillairet

mhuser:
> Hmmm.. From the perspective of readability, it's a little difficult to read. It's subjective to my opinion of course.
> 
> Also if we're following actual IEEE standards (even though clone on Python), how close should our API names be? (perhaps for future cross-check)?
> 
> Hopefully others could chime in on their opinions too, like @jhillairet

Indeed `Ieeep370mmnzc2xthru` is not very readable.
On the IEEE repo, there are these 3 names
- `IEEEP370Zc2xThru` for single-ended 2xtrhu, impedance correction.
- `IEEEP3702xThru` for single-ended 2xthru, no impedance correction. This one at my sense has reduced readability because the `370` and `2x` having no separation.
- `IEEEP370mmZc2xthru` for mixed-mode 2xthru, impedance correction.
- There is no function for mixed-mode 2x thru without impedance correction (but the AICC Tool and `scikit-rf` do it)

The files generated by the AICC Tool also have very meaningful a naming convention like for network `XXX`:

- `deembedded_SE_ZC_XXX.s2p`
- `deembedded_SE_NZC_XXX.s2p`
- `deembedded_MM_ZC_XXX.s4p`
- `deembedded_MM_NZC_XXX.s4p`

Already existing `Deembedding` classes use upper camel case like `class ImpedanceCancel(Deembedding)`

Hence it would be possible to get the better of the two with something in the flavour `IEEEP370_SeZc2xThru` or `IEEEP370_SE_ZC_2xThru` or keep compact like `IEEP370_2xthru` or even something else ?

denzchoe:
I also agree on your observation for below. I initially had trouble spotting the difference.
> * `IEEEP3702xThru` for single-ended 2xthru, no impedance correction. This one at my sense has reduced readability because the `370` and `2x` having no separation.
***
<br>

> Hence it would be possible to get the better of the two with something in the flavour `IEEEP370_SeZc2xThru` or `IEEEP370_SE_ZC_2xThru` or keep compact like `IEEP370_2xthru` or even something else ?

Ah I like this one, clear and concise. Makes the users aware there is a MM variant too: 
`IEEEP370_SE_ZC_2xThru`

jhillairet:
Don't feel obliged to follow strictly the.CamelCase convention, in
particular when dealing with acronyms such as IEEE. I prefer readability,
and IEEEP370_SE_ZC_2xThru is clear I think, in particular to make the
distinction between functions. And also.for future users that will come
from matlab...

mhuser:
- `IEEEP370_SE_NZC_2xthru`
- `IEEEP370_SE_ZC_2xthru`
- `IEEEP370_MM_NZC_2xthru`
- `IEEEP370_MM_ZC_2xthru`

Are now believed to be production ready 😄

jhillairet:
Great work and great addition, thank you @mhuser and @denzchoe. I'm merging it.

denzchoe:
@jhillairet, what is the timeline for the new skrf version on Conda-Forge? 
looking forward to the next stable release (especially the gmm2se bugfix and new deembedding classes)

jhillairet:
It's almost all green to make a release asap. 

Before I will merge the PR concerning nanoVNA and adjust the associated notebook in order to make the test pass.

shengfang2017:
I was using  IEEEP370_SE_ZC_2XThru function, realized s_side2 results are opposite when using makeErrorBox_v7 and makeErrorBox_v8. If using makeErrorBox_v7, s_side2 doesn't flip, I have to manually use s_side2_flip() to filp s_side2 to have deembed() function work correctly. But if using makeErrorBox_v8, s_side2 is correct, don't need to manually flip. When I check makeErrorBox_v7 and makeErrorBox_v8 code, I think they are correct. Not sure where is the problem in the v7 and v8.

shengfang2017:
Here are the screenshots of my result.
using makeErrorBox_v7
image

flip s_side2 manually with makeErrorBox_v7
image

makeErrorBox_v8 without flip s_side2
image

mhuser:
Hello @shengfang2017
Thank you for reporting this.
It should have been fixed in #1072

mhuser:
Have you update scikit-rf to version 1.1.0 ?

mhuser:
You can check with
```Python
import skrf 
print(skrf.__version__)
```

mhuser:
Eventually you could also consider using master branch to beneficiate of #1103 and use `leadin = 10` to capture the small bump of FIX-DUT-FIX before time 0 inside your fixture.

shengfang2017:
Hi @mhuser, thanks for the quick reply. After upgrading to 1.1.0, the problem is fixed. I was wondering which code did you modified to fix the issue. I checked makeErrorBox_v7, looks like same with old version.

mhuser:
Hi @shengfang2017 
You can see the changes here:
https://github.com/scikit-rf/scikit-rf/pull/1072/files
In the Matlab implementation, FIX-2 was flipped. For comparison's sake, I reverted to that convention and un-flipped it only in the end. Before that, I had some confusion with flip resulting in a mess.

PR #630, opened 2022-04-13
SOLT.ipynb example: suggests using nports=2.

biergaizi:
The current example for creating Short, Open, Load ideal components uses `two_port_reflect()`, which is cumbersome. Suggest using the more convenient `nports=2` argument in the tutorial.

### DIFF
--- a/doc/source/examples/metrology/SOLT.ipynb
+++ b/doc/source/examples/metrology/SOLT.ipynb
@@ -148,7 +148,7 @@
    "source": [
     "The ideal component Networks are obtained from your calibration kit manufacturers or from modelling.\n",
     "\n",
-    "In this example, we simulate ideal components from transmission line theory. We create a lossy and noisy transmission line (for the sake of the example) and the ideal components (Short, Open and Load):   "
+    "In this example, we simulate ideal components from transmission line theory. We create a lossy and noisy transmission line (for the sake of the example)."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


jhillairet:
Thanks for your contribution @biergaizi 

I totally agree that the example with `two_port_reflect` is cumbersome, but this is also the point : to be an example, and certainly not the optimized way of doing it. Saying that, I also agree that it is interesting to learn about `nports=2` in this example, but I think it would be more useful *after* having made the cumbersome part.

What I mean is that the "optimized" version:
```
    "# ideal 2-port Networks\n",
    "short_ideal_2p = media.short(nports=2)\n",
    "open_ideal_2p = media.open(nports=2)\n",
    "load_ideal_2p = media.match(nports=2)  # could also be: media.load(nports=2, Gamma0=0)\n",
```
should appear *after* this one:

```
    "short_ideal = media.short()\n",
    "open_ideal = media.open()\n",
    "load_ideal = media.match()\n",
    "short_ideal_2p = rf.two_port_reflect(short_ideal, short_ideal)\n",
    "open_ideal_2p = rf.two_port_reflect(open_ideal, open_ideal)\n",
    "load_ideal_2p = rf.two_port_reflect(load_ideal, load_ideal)"
```
With a comment explaining that there is a less cumbersome way of doing it (using the same variname). Especially also because we should be consistent with what have been written above in this example. Do see what I mean?

doc/source/examples/metrology/SOLT.ipynb
```diff
@@ -179,10 +175,19 @@
...
    "source": [
-    "# generate ideal 2-ports, as required by the scikit-rf SOLT calibration Class \n",
-    "short_ideal_2p = rf.two_port_reflect(short_ideal, short_ideal)\n",
-    "open_ideal_2p = rf.two_port_reflect(open_ideal, open_ideal)\n",
-    "load_ideal_2p = rf.two_port_reflect(load_ideal, load_ideal)"
+    "# ideal 2-port Networks\n",
+    "short_ideal_2p = media.short(nports=2)\n",
+    "open_ideal_2p = media.open(nports=2)\n",
+    "load_ideal_2p = media.match(nports=2)  # could also be: media.load(nports=2, Gamma0=0)\n",
+    "thru_ideal_2p = media.thru()\n",
```
jhillairet, line 182: pay attention that `thru_ideal_2p` is not used after, as the list of ideals uses `thru_ideal` instead. This leads to a failure of the notebook

biergaizi:
I just adjusted the wording and fixed the bug according to your suggestion. Please take a look and let me know if anything still needs to change.

jhillairet (approved):

jhillairet:
> I just adjusted the wording and fixed the bug according to your suggestion. Please take a look and let me know if anything still needs to change.

Thank you @biergaizi. I'm OK and will merge the PR.

PR #635, opened 2022-04-21
se2gmm z0 broadcasting, docstring and tests

Ttl:
* If se2gmm input network ports have unequal z0, calculate the mixed mode z0 from average of differential pair z0. This should make more sense than the previous code which didn't handle different port z0s too well. No change in the common case where all ports have the same z0.
* Allow broadcasting in z0 argument. Makes it possible to for example just give `z0_se=50` in `gmm2se` to make all ports 50 ohms instead of needing to make numpy array with the correct shape.
* Some slight docstring fixes. For complex z0 mention that pseudowave definition is used and add test that se2gmm renormalization matches the network class pseudowave renormalization.
* Add some more tests, no functional changes to the converstion mathematics.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3414,7 +3414,6 @@ class Network(object):
             `f x 2*p x 2*p` matrix of mixed mode impedances, optional
             if input is None, 2 * z0 Ohms differential and z0 / 2 Ohms common mode
             reference impedance is used.
-            Pseudowave definition is used.
 
             If None and differential pair z0 is not identical, average of pair
             z0 is used for mixed mode z0 calculation.
@@ -3495,7 +3494,6 @@ class Network(object):
         z0_se: Numpy array
             `f x 2*p x 2*p` matrix of single ended impedances, optional
             if input is None, extract the reference impedance from the differential network.
-            Pseudowave definition is used.
 
             if None, z0 is calculated as 0.5 * (0.5 * z0_diff + 2 * z0_comm) for each differential port.
         s_def : str -> s_def :  can be: 'power' or 'pseudo'
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3412,8 +3412,12 @@ class Network(object):
             number of differential ports
         z0_mm : Numpy array
             `f x n x n` matrix of mixed mode impedances, optional
-            if input is None, 2 * z_o Ohms differential and z_0 / 2 Ohms common mode 
-            reference impedance is used
+            if input is None, 2 * z0 Ohms differential and z0 / 2 Ohms common mode
+            reference impedance is used.
+            Pseudowave definition is used.
+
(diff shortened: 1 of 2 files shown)


Ttl:
The latest commit should fix #636. I changed the `se2gmm` and `gmm2se` z0 argument shape to only apply to the differential ports since the code did not renormalize the single-ended ports and it makes sense that the function would only affect the differential ports.

I also added a test `test_se2gmm_renorm` that checks that `se2gmm` to `gmm2se` works with various number of mixed mode and single-ended ports. Hopefully this was the last bug fix in this code and it works now.

jhillairet:
Thank you for this.


>     * Some slight docstring fixes. For complex z0 mention that pseudowave definition is used and add test that se2gmm renormalization matches the network class pseudowave renormalization.

Actually, `Network` uses power-waves definition by default, but pseudo-waves definition can be used if desired by changing the `s_def` parameter.

So I wonder if the `scaling` variable:

https://github.com/scikit-rf/scikit-rf/blob/4b1bce9d1425803f8823289be55fb52725752bf8/skrf/network.py#L3534

should follows the `self.s_def` definition, with something like:
```python
if self.s_def == 'power':
    scaling = 1/(2*npy.sqrt(z0.real)
elif self.s_def == 'pseudo':
    scaling = npy.sqrt(z0.real) / (2 * npy.abs(z0))
elif self.s_def == 'traveling':
    scaling = 1/(2*npy.sqrt(z0)
```
?

Ttl:
I added support for power waves and made it default to be consistent with the rest of the skrf. I'm not sure what's up with `traveling` definition? Is it even supposed to work with complex z0? How would S-parameter input and output waves be defined with it? @arsenovic

Ttl:
I added the `traveling` definition. Still not sure if it does anything useful with complex z0.

The default `s_def` should probably be read from `self.s_def` but I copied the way `renormalize` does it for now.

jhillairet:
> I added the `traveling` definition. Still not sure if it does anything useful with complex z0.

The `traveling` definition is in fact the implementation that was used initially in scikit-rf, before power-wave to be the new default. For sure, it does not work with complex z0. 

In fact, only pseudo-waves should work properly with complex z0 (according to Marks & Williams). The problem is that I never found a way to make proper tests with advanced/realistic geometries with the pseudo-waves definition, because none of the advanced/commercial solver use pseudo-waves (but all use power-waves). 

So the pseudo-waves definition exists, but I always wonder if it is interest for someone. The traveling wave definition exist to keep a trace of what was implemented before. 

> The default `s_def` should probably be read from `self.s_def` but I copied the way `renormalize` does it for now.

OK

jhillairet (approved):

jhillairet:
I've merged for now. Thank you @Ttl .

Issue #631, opened 2022-04-18
Suggestion to add a Q-factor fitting algorithm to SKRF


andrewg22:
I recently published a report that contains a previously-unpublished algorithm for fitting Q-factor and resonant frequency to swept-frequency complex S-parameter data. In the case of S21 data, the leakage vector (which has the effect of making resonant peaks on a magnitude display appear asymmetric) is also fitted.

[Report MAT 58](https://www.researchgate.net/profile/Andrew-Gregory-11/publication/357556968_Q-factor_Measurement_by_using_a_Vector_Network_Analyser/)

The algorithm works very well in automated experiments, and could be a useful addition to SKRF. Demonstration Python code can be downloaded. What do other people think about the possibiity of adding it to SKRF? Would this be this useful, or is Q-factor fitting too remote from SKRF's purpose? Comments on how it could be implemented would also be welcome. I have not studied the SKRF codebase in detail, so I may not be able to give an immediate response to detailed comments on coding.

The SKRF implementation would require six coefficient (normally used for S21 or S12 Q-circles) and seven coefficient (normally used for S11 or S22 Q-circles) algorithms to be implemented. These actually determine the indicated (loaded) Q-factor. The unloaded Q-factor, which is the Q-factor of the resonator uncoupled from the 50-Ohm VNA ports, can then be estimated. The best experimental method is to first trigger a sweep with a wide sweep to obtain provisional values of Q and resonant frequency, then repeat the measurement with a narrower sweep centred on the resonance.

Andrew.

jhillairet:
Dear Andrew,

Thank you for your proposal! Yes, I think this is definitely of great interest and fits very well the scikit-rf package purposes.

Concerning the implementation, I see few possibilities we can discuss here, depending on the complexity/functionalities  of your code:
- adding a new function into `skrf/utils.py` that would return Q for a given `skrf.Network` object
- adding a new dedicated class if more functionalities could be provided than just the Q value (is it the case?)

In both case, we can add a new method to `Network` to make the Q-value calculation. 

Any other opinion?

andrewg22:
There is some modest additional complexity to determine the "unloaded" Q from the fitted "loaded" Q. The formula used is different for transmission and reflection resonances. For calculating unloaded Q, there are a few options regarding the 'scaling" of the data if, as is usually the case, the VNA is uncalibrated.

[...]

PR #637, opened 2022-04-23
Q-factor(s) fitting

jhillairet:
This is an implementation of the Q-factor (loaded and unloaded) fitting methods proposed by @andrewg22  in #631

This first version is not yet optimized neither fully integrated to scikit-rf (it does not use `Network` yet)


jhillairet:
@andrewg22. Currently, the `Qfactor()` class does not do anything and the code behaves like in the examples, ie with following workflow:

```python
# preliminary: gets f, S and best guesses for Fseed and Qseed

# init Qfactor
Q = rf.Qfactor()

# step1: loaded Q
a_re, a_im, b_re, b_im, QL  = Q.initial_fit(f, S, N, Fseed, Qseed)

# Step 2: Optimised weighted fit
mv, weighting_ratio, number_iterations, RMS_Error = Q.optimise_fit8(f, D, N, Fseed, sv, loop_plan, Tol, quiet)

# unloaded Q-factor and some other useful quantities.
Qo, cal_diam, cal_gamma_V, cal_gamma_T  = Q.Q_unloaded(mv, scaling_factor_A, trmode, quiet)
```
There is room from improvements, and to make it object-oriented.

Any comments welcome.

arsenovic:
wow!

Vinc0110:
Very cool. I've not yet followed the code in much detail, but I will have a closer look later. Is the algorithm doing heavy calculations that require runtime optimizations? If so, I might be able to help with the tricks I learned during the vector fitting improvements.

One very minor thing I noticed: the source code does not contain the word "quality", not a single time. Of course, every rf engineer knows what a Q-factor is, but maybe it's still helpful to mention it once or twice in the docstrings. For example at the beginning:
https://github.com/scikit-rf/scikit-rf/blob/fd9b29ac16d1ad43114be106312deda400107ac0/skrf/qfactor.py#L5

jhillairet:
Yes, there is room for vectorization improvements, but I would say this is a second step ("premature optimization is..." ;)

What we should discuss first I think is the desired user workflow.

Currently, the fitting is performed by chaining some functions: an initial fit is required before making the improved fitting. Eventually these steps could be combined in a single, but it seems that the precision and the quality of the fit depends much on some user's know-how (like the initial seeds or loop plan)... At this least this is what I understood.  @andrewg22 could confirm or not.

andrewg22:
I've not yet looked at the new code, but it looks as if you are making great progress. I may be able to answer a few points.

The initial and optimised fits can definitely be combined.

I'm sure that some optimisations are possible. Most obviously it could use numpy arrays instead of Python arrays. It's not all that numerically demanding for a typical sweep of 201 points.

It is quite important to have a reasonably accurate initial value for the frequency (Fseed). This can be found from the point of maximum magnitude (transmission) or minimum (reflection or notch resonator). Fseed could be an input parameter with default value None - implying that it will be set from within the function. Qseed is a lot less critical, but it does influence the ability of the optimisation to converge if the VNA sweep is far wider than the resonance. The default value in the test scripts works well.

In some of my work I needed to measure very small shifts in Q-factor, e.g. 0.02 in 400. To allow me to control the iteration accuracy, I provided loop_plan and TOL. The calculation of weighting factors requires a fitted result - so if weightings are required the fitting process must take place at least twice.  Loop_plan allows this, but maybe a simpler scheme would be better for SKRF.

Vinc0110:
Your proposed API seems like the natural approach to me. It actually is very similar to the steps involved in `VectorFitting`, which might be nice for users doing all sorts of fitting with scikit-rf:
1) Init the fitting algorithm with a `Network`
2) Call the fitting method
3) Get the results

Only step 3 is different: `Qfactor.fit()` returns the results directly, whereas `VectorFitting.vector_fit()` does not return anything but stores the results in its attributes (`poles`, `residues`, ...). User can then read them individually or process them with other methods (exporting, plotting, ...). I suggest to do the same for `Qfactor`.

jhillairet:
Thanks @Vinc0110.

Another option I also like is for the `fit()` method to return the fitted version of the object (`return self`), a solution [adopted for example in scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans.fit) .

andrewg22:
I don't have any strong views on whether to obtain unloaded Q within fit() or by a separate function. It is not always necessary to find unloaded Q, so that would be an argument for making it a separate function .. so that would be a slight argument for making it a separate function.

jhillairet:
The proposed API is now:
```python
Q = rf.Qfactor(ntwk, res_type='transmission')

# Optimised weighted fit.
# Overwrite Q.Q_L and Q.f_L with fitted results.
res = Q.fit(method="NLQFIT6")

# Unloaded Q-factor and some other useful quantities.
Q0 = Q.Q_unloaded(res, scaling_factor_A)
cal_diam, cal_gamma_V, cal_gamma_T = Q.Q_circle(res, scaling_factor_A)
```

jhillairet:
@andrewg22 (but also other of course) , I've added a tutorial: 
https://github.com/scikit-rf/scikit-rf/blob/dae54b5e782c5b92f4cb14399c1b14f61960957d/doc/source/tutorials/Q-Factor.ipynb
can you have a look to?

I'm wondering about the API for the fitted trace. Any idea appreciated.

Will also add examples notebook.

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,435 @@
...
+    "\n",
+    "Then\n",
+    "\n",
+    "$$\n",
+    "BW = \\frac{1}{Q_L}\n",
+    "$$\n",
+    "\n",
+    "While the measurements of the loaded Quality factor is straightforward, obtaining uncertainty below 1% (which is considered to be low for Q-factor measurement) requires attention to several aspects of the experimental procedure.\n",
+    "\n",
+    "`scikit-rf` implements the methods described in [3] for determining loaded and unloaded Q-factors from frequency-domain S-parameters, that can be applied to measurements of transmission or reflection.\n",
```
Vinc0110, line 59: Isn't the code based on the method by @andrewg22 (issue #631)? Reference [3] is `David M. Pozar, "Microwave Engineering", 4th Edition, Éditeur Wiley, 2011`

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,435 @@
...
+    "\n",
+    "## Example with a Parallel RLC Circuit\n",
+    "To illustrate the usage of the `Qfactor` class of `scikit-rf`, a parallel RLC circuit as illustrated above is used as an canonical example for which analytical formulas are available for benchmarking.\n",
```
Vinc0110, line 62: That section renders incorrectly. The line break at the end is not working (at least in the Github preview).

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,435 @@
...
+   "execution_count": 5,
+   "id": "226c8348",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Fitted Resonant Frequency: f_L = [5.03291942] MHz\n",
+      "Fitted Loaded Q-factor: Q_L = [592.92718142]\n"
```
Vinc0110, line 200: Why is the fit of the loaded Q so different from the theoretical value above? ("Theoretical Loaded Q: Q_L = 474.34164902525686").
The other parameters match very well, so this seems wrong.

Vinc0110 (commented):
That notebook is really nice! I just added a few comments.

doc/source/tutorials/Q-Factor.ipynb
jhillairet: sure, unfortunate cut/paste, thanks for noticing!

doc/source/tutorials/Q-Factor.ipynb
Ttl: I think the theoretical loaded Q formula is not correct. `R` should be the total resistance in the circuit which is parallel combination of 30 ohm in the resonator and 50 ohm port impedance.

Vinc0110: Calculating the equation for loaded Q with *R = 50||30 Ohm = 18.75 Ohm* gives *Q_L = 296.46*. Interestingly, if you calculate the unloaded Q with *R = 18.75 Ohms*, it is *Q0 = 592.927*, which matches to fitted loaded Q. 

I think you need to use *R=30* for unloaded Q (with only internal losses) and *R=50||30* for loaded Q (with both internal and external losses), but both times using the same equation: `R * C / sqrt(L * C)`. By the way, the *2pi* cancel out elegantly.

18.75Ohm * 1uF / sqrt(1uF * 1nH) = 592.927 (loaded Q)
30Ohm * 1uF / sqrt(1uF * 1nH) = 948.683 (unloaded Q)

Now it matches the fitted results.

jhillairet: Pff, of course, shame on me. Thanks.

jhillairet:
OK, I think it's ready now. 

I've simplified the API. Now the results from the fitting are stored in the class and used per default.

```python
>>> Q = rf.Qfactor(rf.data.ring_slot_meas, res_type='reflection')
>>> opt_res = Q.fit()  # returns the optimized results if needed. Different optimizer can be selected if needed.

>>> print(Q)
Q-factor of Network ring slot measured. (fitted: f_L=85.968GHz, Q_L=3.685)

# resonant frequency and 3dB bandwidth
>>> print(Q.f_L)
[8.59683221e+10]
>>> print(Q.BW)
[2.33318004e+10]

# also exist in scaled to the frequency unit
>>> print(Q.f_L_scaled)  # here in GHz
[85.96832211]
>>> print(Q.BW_scaled)  # here in GHz
[23.33180037]

# passing opt_res is now optional for Unloaded Q, Q-circle, S-parameters and fitted network. 
# If not passed, it uses the one produced during the the last call to fit()
>>> print(Q.Q_unloaded())
[7.79553387]

>>> fitted_ntwk = Q.fitted_network()  

# comparing data and model
>>> rf.data.ring_slot_meas.plot_s_db()
>>> fitted_ntwk.plot_s_db(label='fitted resonant model')
# in polar
>>> rf.data.ring_slot_meas.plot_s_polar()
>>> fitted_ntwk.plot_s_polar(label='fitted resonant model')
```
image
image

jhillairet:
Rendered Doc overview for comments:

https://scikit-rf-jhillairet.readthedocs.io/en/latest/tutorials/Q-Factor.html
https://scikit-rf-jhillairet.readthedocs.io/en/latest/examples/qfactor/Finding_Dk_Df_from_resonance_fitting.html
(weird equation at the end)

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "def Q_RLC(R, L, C):\n",
+    "    return R * C / np.sqrt(L*C)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "99d749a3",
+   "metadata": {},
+   "source": [
+    "he unloaded Q-factor $Q_0$ is a characteristic of the resonator itself, in the absence of any loading effects caused by the external circuitry. In practice, of course, the resonator is connected and coupled to this circuitry which have the effect to lower the Q-factor of the loaded circuit $Q_L$. \n",
```
Vinc0110, line 136: Typo at the beginning: "he unloaded Q-factor ...."

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "\n",
+    "Hence, when the resonator is connected to the transmission line, it is loaded with an equivalent series resistance $Z0$ (assuming no loss in the transmission line), while it is not the case for the resonator alone:"
```
Vinc0110, line 138: Maybe to make it even more clear:
"... it is loaded with the port impedance $Z0$, which is connected in parallel to the intrinsic resonator resistor $R$. The equivalent resonator load is therefore $R||Z0 = (R*Z0) / (R+Z0)$.

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,464 @@
...
+   "source": [
+    "Q = rf.Qfactor(resonator, res_type='reflection')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6f246179",
+   "metadata": {},
+   "source": [
+    "Note that in case of multiple resonances, it is recommanded to also pass the estimated resonance frequency and eventually the estimated (order of magnitude) Q-factor. \n",
```
Vinc0110, line 176: How about linking to the other example notebook with multiple resonances here?

doc/source/tutorials/Q-Factor.ipynb
```diff
@@ -0,0 +1,464 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "240d7ea3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, ax = plt.subplots()\n",
+    "resonator.plot_s_db(label='Parallel RLC ', lw=2, ax=ax)\n",
+    "ax.axvspan(xmin=Q.f_L-Q.BW/2, xmax=Q.f_L+Q.BW/2, alpha=0.3, label='Bandwidth')\n",
```
Vinc0110, line 414: Numpy is not happy here, a warning is printed. I think the problem is that `Q.BW` is a list and should be a number. So maybe try `Q.BW[0]`?

conda.recipe/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "scikit-rf" %}
-{% set version = "0.21.0" %}
+{% set version = "0.22.0" %}
```
Vinc0110, line 2: This does not seem related. Did you mean to do these things in this PR? There is also another change in `doc/source/conf.py`.

Vinc0110 (approved):
Awesome! I also like the new example with multiple resonances. It makes it very clear why is is sometimes necessary to specify the frequency range of interest.

jhillairet:
I'll merge in the present state. One issue remains on the definition of the bandwidth for reflexion resonator. I've exchanged with a few colleagues on how they define it, but it seems there is not a clear definition. So currently the `.BW` property returns the value as defined for transmission resonator (resonance freq/Q_L), but I understand this is incorrect.

PR #649, opened 2022-04-30
More accurate Coaxial resistance model

Ttl:
Improved coaxial cable resistance calculation that works at DC.

S21 with 1 mm coax gives very similar results at high frequencies:
```python
coax1mm = Coaxial(freq, z0=50, Dint=0.44e-3, Dout=1.0e-3, sigma=1e7)
```
coax_s21

However S-parameters are not calculated correctly at DC. Example:

```python
import numpy as np
import skrf
from skrf.media import Coaxial

freq = skrf.Frequency(0, 0, 1)
coax1mm = Coaxial(freq, z0=50, Dint=0.44e-3, Dout=1.0e-3, sigma=1e6)
print('gamma', coax1mm.gamma)

line = coax1mm.line(1, 'm', embed=True)

r_dc = coax1mm.R[0]
print('DC R', r_dc)

# Series resistor with the same resistance
net_r = coax1mm.resistor(R=r_dc)

print()
print('Resistor S-parameters')
print(net_r.s)
print()
print('Line S-parameters')
print(line.s)
print()
```

```
skrf/tlineFunctions.py:160: RuntimeWarning: divide by zero encountered in true_divide
  return sqrt(rho/(pi*f*mu_r*mu_0))
gamma [0.+0.j]
skrf/media/media.py:340: RuntimeWarning: divide by zero encountered in double_scalars
  return 1.0*theta/npy.imag(gamma[int(gamma.size/2)])
skrf/media/media.py:340: RuntimeWarning: invalid value encountered in double_scalars
  return 1.0*theta/npy.imag(gamma[int(gamma.size/2)])
DC R 6.57664599284452

Resistor S-parameters
[[[0.06170813+0.j 0.93829187+0.j]
  [0.93829187+0.j 0.06170813+0.j]]]

Line S-parameters
[[[nan+nanj nan+nanj]
  [nan+nanj nan+nanj]]]

```

### DIFF
--- a/skrf/media/coaxial.py
+++ b/skrf/media/coaxial.py
@@ -14,8 +14,8 @@ A coaxial transmission line defined from its electrical or geometrical/physical 
 
 #from copy import deepcopy
 from scipy.constants import  epsilon_0, mu_0, pi, c
-from numpy import sqrt, log, real, imag, exp, size
-from ..tlineFunctions import surface_resistivity
+from numpy import sqrt, log, real, imag, exp, expm1, size
+from ..tlineFunctions import surface_resistivity, skin_depth
 from .distributedCircuit import DistributedCircuit
 from .media import Media, DefinedGammaZ0
 from ..constants import INF, NumberLike
@@ -227,7 +227,7 @@ class Coaxial(DistributedCircuit, Media):
     def b(self) -> NumberLike:
         """
         Outer radius of the coaxial line
-        
+
         Returns
         -------
         b : float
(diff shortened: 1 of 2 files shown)


jhillairet (approved):
OK, nice! 

Btw microwaves101 also published some year ago a loss formula with two different metals (ex: copper on inner, aluminium on outer conductor).

jhillairet:
@Ttl , is the PR ready for merging?

Ttl:
Yes.

The equations I used are from the same microwaves101 page that you mentioned and it's also referenced in the docstring. I set here both materials the same and the outer shield to be infinitely thick.

jhillairet:
OK. Eventually one could add later the possibility to use two different conductivity for the inner and the outer, if useful.

PR #655, opened 2022-05-06
Consistent Network handling with different s_def

### LINKED ISSUE #654: Connecting networks with complex port impedances gives incorrect results
By default `Network` S-parameters are power waves, but connecting of networks is always calculated following `traveling` definition. If we generate two random two-port networks with Z-parameters we can calculate their cascade easily with ABCD parameters. We can also convert the Z-parameters to S-parameters assigning random complex port impedances and calculate the cascade network. The two methods should give the same resulting network and identical S-parameters when renormalized to the same port impedances but this is not the case with `power` and `pseudo` S-parameter definitions.

```python
import skrf
import numpy as np
np.random.seed(123)

freq = skrf.F(1,1,1)

# Generate random Z-parameters for two networks
z1 = np.random.uniform(1, 100, size=(1, 2, 2)) + 1j*np.random.uniform(-100, 100, size=(1, 2, 2))
z2 = np.random.uniform(1, 100, size=(1, 2, 2)) + 1j*np.random.uniform(-100, 100, size=(1, 2, 2))

# Port impedances
if 1:
    z0_1 = np.random.uniform(1, 100, size=2) + 1j*np.random.uniform(-100, 100, size=2)
    z0_2 = np.random.uniform(1, 100, size=2) + 1j*np.random.uniform(-100, 100, size=2)
    z0_3 = np.random.uniform(1, 100, size=2) + 1j*np.random.uniform(-100, 100, size=2)
else:
    # All S-parameters should be identical with only real port impedances
    z0_1 = 50
    z0_2 = 50
    z0_3 = 50

# Cascade Z-parameters calculated with ABCD parameters
net3_z = skrf.a2z(skrf.z2a(z1) @ skrf.z2a(z2))

for s_def in skrf.S_DEFINITIONS:
    print('s_def:', s_def)
    print('Cascade S-parameters calculated with Z and ABCD parameters:')
    # Initializing with z doesn't work
    net3_ref = skrf.Network(s=[[0,0],[0,0]], frequency=freq, z0=z0_3, s_def=s_def)
    net3_ref.z = net3_z
    print(net3_ref.s)

    net1 = skrf.Network(s=[[0,0],[0,0]], frequency=freq, z0=z0_1, s_def=s_def)
    net2 = skrf.Network(s=[[0,0],[0,0]], frequency=freq, z0=z0_2, s_def=s_def)
    net1.z = z1
    net2.z = z2

    print('Cascade calculated with S-parameters:')
    net12 = net1 ** net2
    net12.renormalize(z0_3, s_def=s_def)
    print(net12.s)
    print('Max error', np.max(np.abs(net12.s - net3_ref.s)))
    print()
```

```
s_def: power
Cascade S-parameters calculated with Z and ABCD parameters:
[[[-0.23391251+0.08191357j -0.11384683-0.17083815j]
  [ 0.24742266+0.06958025j  0.48164147+0.16554454j]]]
Cascade calculated with S-parameters:
[[[-0.15537994+0.18034563j -0.13194359-0.08436486j]
  [ 0.1941495 -0.02734986j  0.50599347+0.10007274j]]]
Max error 0.1259215466224455

s_def: pseudo
Cascade S-parameters calculated with Z and ABCD parameters:
[[[-0.20269054+0.55222866j -0.32239341-0.22958527j]
  [ 0.2010088 -0.18775523j  0.75449885+1.01992456j]]]
Cascade calculated with S-parameters:
[[[-0.20269054+0.55222866j -0.26726058-0.25330074j]
  [ 0.18583264-0.22994014j  0.75449885+1.01992456j]]]
Max error 0.060017105406647724

s_def: traveling
Cascade S-parameters calculated with Z and ABCD parameters:
[[[-0.20269054+0.55222866j -0.2827237 -0.08380469j]
  [ 0.33699066-0.15076128j  0.75449885+1.01992456j]]]
Cascade calculated with S-parameters:
[[[-0.20269054+0.55222866j -0.2827237 -0.08380469j]
  [ 0.33699066-0.15076128j  0.75449885+1.01992456j]]]
Max error 5.214799211354908e-16
```

For `pseudo` definition the bug can be fixed by changing the S-parameters in `impedance_mismatch` function (https://github.com/scikit-rf/scikit-rf/blob/master/skrf/network.py#L6576) to:
```python
n = npy.abs(z2/z1) * npy.sqrt(z1.real / z2.real)
result[:, 0, 0] = gamma
result[:, 1, 1] = -gamma
result[:, 1, 0] = 2 * z2 / (n * (z1 + z2))
result[:, 0, 1] = 2 * z1 * n / (z1 + z2)
```

For `power` I don't currently know how it should be fixed.

Ttl:
Fixes #654. Adds support for connecting networks with different `s_def` including support for connecting network with different `s_def`. Keeps the default `s_def` as `power` but now S-parameters generated for example by `media.line` are actually `power` instead of `traveling` previously. Use `Network.s_def` in operations that depend on it such as `renormalize` and update it if it has changed during it.

If we would like to keep the old behavior with complex port impedances then the default `s_def` should be changed to `traveling`.

Few test failures in Circuit since it assumes `traveling` definition. All other tests seem to pass without changes. New tests are added to check that networks with different `s_def` and complex port impedances are handled correctly. No changes for real port impedances.


jhillairet (approved):
OK

skrf/tests/test_network.py
```diff
@@ -274,6 +274,110 @@ def test_connect_no_frequency(self):
...
+        net3_z = rf.a2z(rf.z2a(z1) @ rf.z2a(z2))
+
+        for s_def in rf.S_DEFINITIONS:
+            net3_ref = rf.Network(s=[[0,0],[0,0]], f=1, z0=z0_3, s_def=s_def)
+            net3_ref.z = net3_z
+
+            net1 = rf.Network(s=[[0,0],[0,0]], f=1, z0=z0_1, s_def=s_def)
+            net2 = rf.Network(s=[[0,0],[0,0]], f=1, z0=z0_2, s_def=s_def)
+            net1.z = z1
+            net2.z = z2
```
jhillairet, line 306: you can use the class method `network.from_z` eventually

jhillairet (commented):
Good job and great tests also.

skrf/media/media.py
```diff
@@ -821,7 +822,9 @@ def line(self, d: NumberLike, unit: str = 'deg',
             z0 = parse_z0(z0)* self.z0
 
         kwargs.update({'z0':z0})
-        result = self.match(nports=2,**kwargs)
+        s_def = kwargs.pop('s_def', S_DEF_DEFAULT)
+        # Zref = Z0 so these S-parameters correspond to traveling waves.
+        result = self.match(nports=2, s_def='pseudo', **kwargs)
```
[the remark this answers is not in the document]
Ttl: I added this because `NISTMultilineTRLTest2` test failed without it. I investigated it a little bit more and it shouldn't be in theory necessary. The issue is that with the default `power` connecting to thru in embed causes numerical issues. The reason is that when connecting `power` networks they need to be converted to either other two definitions first. The conversion is done by converting to Z-parameters and back. Unfortunately thru Z-parameters have infinities which cause the calculation to be inaccurate when there are both huge and small numbers. I think before this can be merged s_def conversion to and from power should be done in better way.

skrf/tests/test_network.py
Ttl: I wrote some code for supporting initializing networks with all supported parameters and not just s. I'll make a separate PR for it later.

jhillairet: > I wrote some code for supporting initializing networks with all supported parameters and not just s. I'll make a separate PR for it later.

like `rf.Network(z=[[0,0],[0,0]], f=1, z0=z0_1, s_def=s_def)` ?

skrf/media/media.py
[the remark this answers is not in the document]
Ttl: It actually needs to not be power, because it doesn't match with the traveling wave definition since it has the conjugate in reflection coefficient definition.

skrf/tests/test_network.py
Ttl: yes

skrf/network.py
```diff
@@ -3885,6 +3914,28 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
...
+        ntwkB = ntwkB.copy()
+        ntwkB.renormalize(ntwkB.z0, ntwkA.s_def)
+
+    s_def_original = ntwkA.s_def
+    if ntwkA.s_def == 'power' and have_complex_ports:
+        # Connecting power wave is not supported, so renormalize to pseudo waves
+        ntwkA = ntwkA.copy()
+        ntwkA.renormalize(ntwkA.z0, 'pseudo')
+        ntwkB = ntwkB.copy()
+        ntwkB.renormalize(ntwkB.z0, 'pseudo')
```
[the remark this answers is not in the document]
Ttl: It would get potentially raised constantly. I'm also not sure if it's a definition change but a bug in the previous implementation.

Ttl:
This should be working now. I added `s2s` function that can change the `s_def` without going through Z-parameters since that is potentially inaccurate in some cases. `renormalization` could also be implemented without Z-parameters using the equations in this paper: https://ieeexplore.ieee.org/document/8430023.

skrf/network.py
```diff
@@ -6564,6 +6766,12 @@ def impedance_mismatch(z1: NumberLike, z2: NumberLike) -> npy.ndarray:
             complex impedance of port 1
     z2 : number or array-like
             complex impedance of port 2
+    s_def : str
+        Scattering parameter definition:
+        'pseudo' for pseudo-waves definition.
+        'traveling' corresponds to the initial implementation.
+        'power' is not supported.
```
jhillairet, line 6773:
```suggestion
        'power' is not supported.
```
well, it could in case of real-valued reference impedance.

Did you use a reference for the equations, to add it in the docstring?

skrf/network.py
```diff
@@ -3885,6 +4000,28 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
...
+        ntwkA = ntwkA.copy()
+        ntwkA.renormalize(ntwkA.z0, 'pseudo')
+        ntwkB = ntwkB.copy()
+        ntwkB.renormalize(ntwkB.z0, 'pseudo')
+
+    s_def = ntwkA.s_def
+    if s_def == 'power':
+        # Ports are real. Save some time by not copying the networks
+        # and use the definition that works with real ports
+        s_def = 'traveling'
```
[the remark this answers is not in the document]
Ttl: This only affects how the calculation is done. Since all definitions agree when ports are real we can just use traveling calculation for power nets with real ports. It doesn't change the network s_def. Even with complex ports the s_def of the output network is changed back to what input net had.

skrf/network.py
```diff
@@ -3885,6 +4000,28 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
     if (l + num - 1 > ntwkB.nports - 1):
         raise IndexError('Port `l` out of range')
 
+    have_complex_ports = (ntwkA.z0.imag != 0).any() or (ntwkB.z0.imag != 0).any()
+
+    # If definitions aren't identical and there are complex ports renormalize first
+    # Output will have ntwkA s_def if they are different.
+    if ntwkA.s_def != ntwkB.s_def and have_complex_ports:
```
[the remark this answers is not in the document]
Ttl: Makes sense. Maybe warning would be enough in this case as the operation is valid but might have unexpected results?

skrf/network.py
Ttl: It's not directly in this form but the T-matrix for pseudo impedance mismatch is from: R. B. Marks et D. F. Williams, "A general waveguide circuit theory". I will add it to docstring.

skrf/network.py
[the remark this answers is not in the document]
jhillairet: OK

jhillairet (approved):
OK to me.

Ttl:
Testing notebooks seemed to fail on my branch: https://github.com/Ttl/scikit-rf/runs/6341058544 although it passed here.

The issue with that notebook seems to be that it generates line at 0 Hz and it calculates characteristic impedance to be nan due to math breaking down at 0 Hz. `nan != nan` so in renormalize the `z2s(s2z(...))` part is called which raises Exception when trying to invert matrix with nans. Apparently previously everything else was silently okay with nans.

skrf/network.py
```diff
@@ -6564,19 +6774,40 @@ def impedance_mismatch(z1: NumberLike, z2: NumberLike) -> npy.ndarray:
...
+        result[:, 0, 1] = (1 - gamma) * npy.sqrt(1.0 * z2 / z1)
+    elif s_def == 'pseudo':
+        n = npy.abs(z2/z1) * npy.sqrt(z1.real / z2.real)
+        result[:, 0, 0] = gamma
+        result[:, 1, 1] = -gamma
+        result[:, 1, 0] = 2 * z2 / (n * (z1 + z2))
+        result[:, 0, 1] = 2 * z1 * n / (z1 + z2)
+    else:
+        raise ValueError('Unsupported s_def: {}'.format(s_def))
+
```
jhillairet, line 6810: Actually, it is possible to derive the S-param of a mismatch using the following definitions:
image

jhillairet: I don't understand your answer. The equations above correpond to the S-parameters for a mismatch for the power wave definition, which is currently not supported in the `impedance_mismatch`. This is related with the if sections in `connect`: 

https://github.com/scikit-rf/scikit-rf/blob/9f5a57dc2ccb99367b8d317e07772ac13ba46d0a/skrf/network.py#L4015-L4031

Ttl: I didn't try it, but I don't think it solves the problem that power waves with complex ports can't be composed because wave leaving the port is not identical to the wave coming into port that is connected to it, even if the port impedances match. The connection algorithm depends on this continuity.

jhillairet: Sure, I didn't pretend it solves the problem of cascading (it can't be in fact), but this was just to provide a solution to `impedance_mismatch` for power waves, even it is not used by the `connect` algorithm (just in case).

Ttl:
I looked at some other methods in media and there are some that don't work correctly with complex ports. For example short is defined to have -1 S11, but that is not the case with powerwave S-parameters with complex ports. Also there is `impedance_mismatch` in media class which could make use of the powerwave mismatch equations.

skrf/media/media.py
```diff
@@ -646,11 +649,9 @@ def impedance_mismatch(self, z1: NumberLike, z2: NumberLike, **kwargs) -> Networ
         resistor
         """
         result = self.match(nports=2, **kwargs)
+        s_def = kwargs.get('s_def', S_DEF_DEFAULT)
         gamma = tf.zl_2_Gamma0(z1,z2)
```
jhillairet, line 653:
```suggestion
        gamma = tf.zl_2_Gamma0(z1,z2)
```
gamma is no more useful

jhillairet:
We will have to document all this clearly, as it may surprise some users ;)

PR #657, opened 2022-05-08
Add support for initializing network with different parameters

Ttl:
Adds support for initializing network from all support parameters (s, z, y, a, h and t). Also `f` initializer can now take `Frequency` object.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1328,7 +1328,7 @@ class Network(object):
         return self.frequency.f
 
     @f.setter
-    def f(self, f: NumberLike) -> None:
+    def f(self, f: Union[NumberLike, Frequency]) -> None:
         if isinstance(f, Frequency):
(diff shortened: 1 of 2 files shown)


skrf/network.py
```diff
@@ -1322,8 +1329,11 @@ def f(self) -> npy.ndarray:
 
     @f.setter
     def f(self, f: NumberLike) -> None:
```
jhillairet, line 1331:
```suggestion
    def f(self, f: Union[NumberLike, Frequency]) -> None:
```

jhillairet (approved):
Wow, nicely done. Just a suggestion for the typing, which should be mergeable to your PR.

jhillairet:
@Ttl , is it OK to be merged?

Ttl:
Yes

PR #661, opened 2022-05-13
Improve s2z, z2s, s2y and y2s accuracy

### LINKED ISSUE #659: Cannot reproduce the results from the "De-embedding with Scikit-RF" section in the online documentation
Python and relevant librairies versions
--------------------------------------------------

* skrf: 0.22.1 (from conda-forge)
* python: 3.7.13 (from conda)
* matplotlib: 3.5.1 (from conda)

Description
---------------

I tried to reproduce the example section [“De-embedding with Scikit-RF”](https://scikit-rf.readthedocs.io/en/latest/tutorials/Deembedding.html#De-embedding-with-Scikit-RF) in the scikit-rf online documentation but the results that I get after de-embedding look weird and do not agree with with is shown in the documentation. Not sure if this is a bug or if I am simply doing something wrong.

I generate the (synthetic) required data as Touchstone files using Qucs (0.0.19) and following the schematics shown in the documentation. Here is the Qucs *.sch file that I used in practice: [deembedding_example.sch.txt](https://github.com/scikit-rf/scikit-rf/files/8672073/deembedding_example.sch.txt) (*remove the `.txt` extension before use*).

Screenshots of the simulated setups:

**DUT**
dut_measurement_setup

**open**
open_measurement_setup

**short**
short_measurement_setup

Here are the output Touchstone files just in case (*remove the `.txt` extension before use*):
* [ind.s2p.txt](https://github.com/scikit-rf/scikit-rf/files/8672103/ind.s2p.txt)
* [open.s2p.txt](https://github.com/scikit-rf/scikit-rf/files/8672106/open.s2p.txt)
* [short.s2p.txt](https://github.com/scikit-rf/scikit-rf/files/8672108/short.s2p.txt)

Once this is done, I simply use the code provided in the documentation (with only minor tweaks):
```python
# Copy-pasted (with only very minor changes mostly related to style) from:
# https://scikit-rf.readthedocs.io/en/latest/tutorials/Deembedding.html#De-embedding-with-Scikit-RF

# Look at the raw inductor measurement with parasitics included
# From S11/S22, it is clear that it is not a pure inductance.

import skrf as rf
import numpy as np

from skrf.calibration import OpenShort

plt.close("all")

fig0, ax0 = plt.subplots(1, 1, num="raw_inductance_s-parameters")
raw_ind = rf.Network("ind.s2p")
raw_ind.plot_s_smith(ax=ax0)

# Plot the inductance and q-factor of the raw inductor measurement.
Lraw_nH = 1e9 * np.imag(1/raw_ind.y[:,0,0]) / 2 / np.pi / raw_ind.f
Qraw = np.abs(np.imag(1 / raw_ind.y[:,0,0]) / np.real(1 / raw_ind.y[:,0,0]))

fig1, (ax1, ax2) = plt.subplots(1, 2, num="raw_inductance_measurements")
ax1.plot(raw_ind.f*1e-9, Lraw_nH)
ax1.grid(True)
ax1.set_ylabel("Inductance (nH)")
ax1.set_xlabel("Frequency (GHz)")
ax2.plot(raw_ind.f*1e-9, Qraw)
ax2.grid(True)
ax2.set_ylabel("Q-factor")
ax2.set_xlabel("Frequency (GHz)")
fig1.tight_layout()  # might rather rely on `plt.subplots(..., 'layout'='constrained')` since Matplotlib 3.5

# Load in short/open dummies as networks.
fig2, ax3 = plt.subplots(1, 1, num="actual_inductance_s-parameters")
open_nw = rf.Network("open.s2p")
short_nw = rf.Network("short.s2p")

dm = OpenShort(dummy_open=open_nw, dummy_short=short_nw, name='tutorial')

actual_ind = dm.deembed(raw_ind)
actual_ind.plot_s_smith(ax=ax3)

# Plot the inductance of the de-embedded measurement.
# We ignore plotting Q-factor here, because an ideal lossless inductor has infinite Q.

Lactual_nH = 1e9 * np.imag(1 / actual_ind.y[:,0,0]) / 2 / np.pi / actual_ind.f

fig3, ax4 = plt.subplots(1, 1, num="actual_inductance_measurements")
ax4.plot(actual_ind.f*1e-9, Lactual_nH)
ax4.grid(True)
#ax4.set_ylim(0.95, 1.1)
ax4.set_ylabel("Inductance (nH)")
ax4.set_xlabel("Frequency (GHz)")
fig3.tight_layout()

# Uncomment the 2 following lines if you want to export the figures

#for fig in (fig0, fig1, fig2, fig3):
#    fig.savefig(fig.get_label() + ".png", dpi=100)
```

The output figures are (in production order):

**fig0** (`raw_inductance_s-parameters`)
raw_inductance_s-parameters

**fig1** (`raw_inductance_measurements`)
raw_inductance_measurements

**fig2** (`actual_inductance_s-parameters`)
actual_inductance_s-parameters

**fig3** (`actual_inductance_measurements`)
actual_inductance_measurements

If both fig0 and fig1 seems to be OK, I would expect fig2 to look like
tutorials_Deembedding_9_0
and fig3 like
tutorials_Deembedding_10_0
while they clearly do not. Did I do something wrong or is there some kind of an issue with the `OpenShort.deembed` method (or in the documentation)?

Expected outcome
-------------------------

Getting all the four same figures as in the  [“De-embedding with Scikit-RF”](https://scikit-rf.readthedocs.io/en/latest/tutorials/Deembedding.html#De-embedding-with-Scikit-RF) section of the online documentation.

Ttl:
Slightly nudge the S-parameters if Z-parameters would be too close to infinity by changing any eigenvalues that are too small.
The previous method of adding a small value if S-parameter was one did not fix the issue except for in special cases (open and short).

Also improves the numerical accuracy in calculation by avoiding directly calculating the inverse of a matrix.

Fixes #659.

Example:
```python
net = skrf.Network(s=[[0.5, 0.5],[0.5, 0.5]])
print(net.z)
print()
print(skrf.z2s(skrf.s2z(net.s, net.z0), net.z0))

```

```
Previously:
numpy.linalg.LinAlgError: Singular matrix

After:
[[[4.99999984e+10+0.j 4.99999983e+10+0.j]
  [4.99999983e+10+0.j 4.99999984e+10+0.j]]]

[[[0.49999998-0.j 0.50000002-0.j]
  [0.49999998-0.j 0.50000002-0.j]]]
```

### DIFF
--- a/skrf/mathFunctions.py
+++ b/skrf/mathFunctions.py
@@ -1264,6 +1264,9 @@ def rsolve(A: npy.ndarray, B: npy.ndarray) -> npy.ndarray:
 
     Calls numpy.linalg.solve with transposed matrices.
 
+    Same as B @ npy.linalg.inv(A) but avoids calculating the inverse and
+    should be numerically slightly more accurate.
+
     Input should have dimension of similar to (nfreqs, nports, nports).
 
     Parameters
@@ -1280,7 +1283,7 @@ def rsolve(A: npy.ndarray, B: npy.ndarray) -> npy.ndarray:
 
 def nudge_eig(mat: npy.ndarray, cond: float = 1e-9, min_eig: float = 1e-12) -> npy.ndarray:
     r"""Nudge eigenvalues with absolute value smaller than
-    min(cond * max(eigenvalue), min_eig) to that value.
+    max(cond * max(eigenvalue), min_eig) to that value.
     Can be used to avoid singularities in solving matrix equations.
 
     Input should have dimension of similar to (nfreqs, nports, nports).
@@ -1304,6 +1307,10 @@ def nudge_eig(mat: npy.ndarray, cond: float = 1e-9, min_eig: float = 1e-12) -> n
     max_eig = npy.amax(npy.abs(eigw), axis=1)
     # Calculate mask for positions where problematic eigenvalues are
     mask = npy.logical_or(npy.abs(eigw) < cond * max_eig[:, None], npy.abs(eigw) < min_eig)
+    if not mask.any():
+        # Nothing to do. Return the original array.
+        return mat
+
     mask_cond = cond * npy.repeat(max_eig[:, None], mat.shape[-1], axis=-1)[mask]
     mask_min = min_eig * npy.ones(mask_cond.shape)
     # Correct the eigenvalues
--- a/skrf/mathFunctions.py
+++ b/skrf/mathFunctions.py
@@ -1258,3 +1258,58 @@ def is_positive_semidefinite(mat: npy.ndarray, tol: float = ALMOST_ZERO) -> bool
     except npy.linalg.LinAlgError:
         return False
     return npy.all(v > -tol)
+
+def rsolve(A: npy.ndarray, B: npy.ndarray) -> npy.ndarray:
+    r"""Solves x @ A = B.
+
+    Calls numpy.linalg.solve with transposed matrices.
+
+    Input should have dimension of similar to (nfreqs, nports, nports).
+
+    Parameters
+    ----------
+    A : npy.ndarray
+    B : npy.ndarray
+
+    Returns
+    -------
+    x : npy.ndarray
+    """
+    return npy.transpose(npy.linalg.solve(npy.transpose(A, (0, 2, 1)).conj(),
+            npy.transpose(B, (0, 2, 1)).conj()), (0, 2, 1)).conj()
+
+def nudge_eig(mat: npy.ndarray, cond: float = 1e-9, min_eig: float = 1e-12) -> npy.ndarray:
+    r"""Nudge eigenvalues with absolute value smaller than
+    min(cond * max(eigenvalue), min_eig) to that value.
+    Can be used to avoid singularities in solving matrix equations.
+
+    Input should have dimension of similar to (nfreqs, nports, nports).
+
+    Parameters
+    ----------
+    mat : npy.ndarray
+        Matrices to nudge
+    cond : float, optional
+        Minimum eigenvalue ratio compared to the maximum eigenvalue
+    min_eig : float, optional
+        Minimum eigenvalue
+    Returns
+    -------
+    res : npy.ndarray
+        Nudged matrices
+    """
+    # Eigenvalues and vectors
+    eigw, eigv = npy.linalg.eig(mat)
+    # Max eigenvalue for each frequency
+    max_eig = npy.amax(npy.abs(eigw), axis=1)
+    # Calculate mask for positions where problematic eigenvalues are
+    mask = npy.logical_or(npy.abs(eigw) < cond * max_eig[:, None], npy.abs(eigw) < min_eig)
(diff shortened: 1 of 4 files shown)


skrf/mathFunctions.py
```diff
@@ -1258,3 +1258,58 @@ def is_positive_semidefinite(mat: npy.ndarray, tol: float = ALMOST_ZERO) -> bool
...
+    ----------
+    A : npy.ndarray
+    B : npy.ndarray
+
+    Returns
+    -------
+    x : npy.ndarray
+    """
+    return npy.transpose(npy.linalg.solve(npy.transpose(A, (0, 2, 1)).conj(),
+            npy.transpose(B, (0, 2, 1)).conj()), (0, 2, 1)).conj()
```
jhillairet, line 1279: Actually, this should work as well (tested with NumPy 1.20)
```suggestion
    return B @ npy.linalg.inv(A)
```

skrf/mathFunctions.py
```diff
@@ -1258,3 +1258,58 @@ def is_positive_semidefinite(mat: npy.ndarray, tol: float = ALMOST_ZERO) -> bool
...
+
+def nudge_eig(mat: npy.ndarray, cond: float = 1e-9, min_eig: float = 1e-12) -> npy.ndarray:
```
jhillairet, line 1281: nice. Did you use a reference for that? I'd like to learn more on that.

jhillairet (approved):
Great!

skrf/mathFunctions.py
Ttl: The point of this function is to avoid calculating the inverse because calculating the inverse and then multiplying the matrices is numerically more inaccurate than directly solving the linear system. In Z-parameter calculations some of the numbers can be very big and we want to keep as much precision as possible.

skrf/mathFunctions.py
Ttl: I don't have any specific references for this method. I first tried adding a small constant times diagonal matrix if the condition number of the matrix was very big but that could fail in some cases. Since condition number is calculated from eigenvalues it makes sense to just change the problematic eigenvalues directly to improve the condition number.

skrf/mathFunctions.py
jhillairet: OK, good to know. Could please add such a comment in the function? In case somewhere in the future would think like me that it could be simplified.

skrf/mathFunctions.py
jhillairet: OK.

skrf/mathFunctions.py
jhillairet: ok, already done. Perfect.

jhillairet:
OK to me, you can merge it if you are ready to.

Ttl:
I think I might as well apply the same fix to `s2y` and `y2s` first. While I think it's more uncommon to run problems with it there are some inputs that cause similar issues with it.

Issue #610, opened 2022-02-27
Offset loss Keysight Calkit model


tuxgvf:
Hi guys,
It seems to me that the example for defining a Calkit from thé polynomial model, the offset loss are missing (the famous unit in Gohm/s). 
It can be a good complement to approach metrology measurement.
Thanks and best.

jhillairet:
Hello,
Which example are you talking about?

tuxgvf:
Hello Julien,

I can't remember exaclty where I found it but it's something like this:

import sys
import skrf
import matplotlib.pyplot as plt
import numpy as np
skrf.stylely()

def tline_input(zl, z0, t, f):
    c = 299792458
    w = c/f
    b = np.pi*2/w
    l = t*c
    return z0*(zl+1j*z0*np.tan(b*l))/(z0+1j*zl*np.tan(b*l))

def gamma(zl, z0):
    return (zl-z0)/(zl+z0)

def make_open(freqs, c0, c1, c2, c3, offset_t, offset_z0):
    c_coefs = [c0*1e-15, c1*1e-27, c2*1e-36, c3*1e-45]

    reactance = []
    for f in freqs:
        c = 0
        for i in xrange(len(c_coefs)):
            c += c_coefs[i]*f**i
        xc = (-1.0j/(2*np.pi*f*c))
        reactance.append(xc)
    reactance = np.array(reactance)
    sparam = gamma(tline_input(reactance, offset_z0, offset_t, freqs), 50)

    return skrf.Network(s=sparam, f=freqs, f_unit='Hz')

def make_short(freqs, l0, l1, l2, l3, offset_t, offset_z0):
    c_coefs = [l0*1e-12, l1*1e-24, l2*1e-34, l3*1e-42]

    reactance = []
    for f in freqs:
        c = 0
        for i in xrange(len(c_coefs)):
            c += c_coefs[i]*f**i
        xc = (1.0j*2*np.pi*f*c)
        reactance.append(xc)
    reactance = np.array(reactance)
    sparam = gamma(tline_input(reactance, offset_z0, offset_t, freqs), 50)

    return skrf.Network(s=sparam, f=freqs, f_unit='Hz')

def make_load(freqs, r0, l0, l1, l2, l3, offset_t, offset_z0):
    c_coefs = [l0*1e-12, l1*1e-24, l2*1e-34, l3*1e-42]

    reactance = []
    for f in freqs:
        c = 0
        for i in xrange(len(c_coefs)):
[...]

Issue #651, opened 2022-05-01
Creating a line with characteristic impedance of the media is confusing


Ttl:
At very low frequencies characteristic impedance of transmission line is complex. skrf doesn't seem to handle it correctly because mixing of port reference impedance `z0` and characteristic impedance of line `Z0` in calculating the S-parameters and use of powerwave S-parameter definition.

Below is example of coaxial line with 1 cm length and very low conductivity resulting in high loss. DC resistance is about 50 ohms and the frequency is 1 MHz so the line is electrically very short and could be modeled as series resistor. Requires #649 for resistance to be calculated correctly.

`media_z0` is `z0` given to the media. Currently this means setting the port impedance `z0` to this value and characteristic impedance `Z0` is left unchanged. None sets port impedance to characteristic impedance.

`line_z0` is the characteristic and port impedance to use when creating a line.

If `embed` is True then the port impedances are renormalized to `z0`.

Given the above definitions I would assume that setting `media_z0=50, line_z0=None, embed=True` should make a line with the complex characteristic impedance and renormalize the ports to 50 ohms matching the series resistor calculation. However this is not the case.

```python
import numpy as np
import skrf
from skrf.media import Coaxial, Freespace

freq = skrf.Frequency(1, 1, 1, unit='MHz')
line_l = 1e-2 # meters

for media_z0 in [50, None]:
    for line_z0 in [None, 50]:
        for line_embed in [False, True]:
            print(f'media_z0: {media_z0}, line_z0: {line_z0}, Embed: {line_embed}')
            media = Coaxial(freq, z0=media_z0, Dint=0.5e-3, Dout=1.0e-3, sigma=1e3)
            print(' z0: {:.2f}, Z0: {:.2f}, gamma: {:.2f}'.format(media.z0[0], media.Z0[0], media.gamma[0]))
            line = media.line(line_l, 'm', embed=line_embed, z0=line_z0)
            print(' line S11', line.s[0,0,0])
            print(' line S21', line.s[0,1,0])
            print(' line z0', line.z0[0,0])
            if line.z0[0,0] != 50:
                line.renormalize(50, s_def='traveling')
                print(' line renorm S11', line.s[0,0,0])
                print(' line renorm S21', line.s[0,1,0])
            print()

print('S-parameters calculated as series resistor')
r_dc = line_l*(1/media.sigma)/(np.pi*media.a**2)
print('dc R', r_dc)

if 0:
    # Manually assemble series resistor S-parameters
[...]

PR #663, opened 2022-05-14
New example: Defining SOLT calibration standards.

biergaizi:
This example shows the method for modeling SOLT calibration standards in scikit-rf using their calibration coefficients, including offset loss, offset delay, and distributed capacitance and inductance defined by a third-degree polynomial function. Both the HP-Agilent-Keysight format and the Rohde & Schwarz / Anritsu format are included.
    
This example implements the request for documentation by issue #610.

<del>This is the first draft, all the code and examples in the document are untested.</del> Please review my calculations before merging it.


jhillairet:
Great introduction and tutorial, thanks ! 

> "in most SOLT calibrations, Thru is first connected to the test cable, and all the Open, Short, Load, and Thru standards are attached to the other side of the Thru"
Interesting. Some have told me the opposite, that is, Open, Short, Load should be connected to the cables directly, not to the Thru.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,590 @@
...
+   "source": [
+    "resistor_open = []\n",
+    "for f in freq.f:\n",
+    "        r = 2.2 * 29.242 * np.sqrt(f / 1e9)\n",
+    "        resistor_open += [r]\n",
+    "loss_open = medium.resistor(np.array(resistor_open))\n",
+    "\n",
+    "resistor_short = []\n",
+    "for f in freq.f:\n",
+    "        r = 2.36 * 31.785 * np.sqrt(f / 1e9)\n",
```
Ttl, line 182: This doesn't look right. The resistance seems way too high and the units here are (Gohms/s) * ps, which doesn't give ohms.

I have understood that the loss is transmission line loss similar to the other definition below. Lossless delay and resistor does not give the same answer as a lossy transmission line. Is this lumped model correct?

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,590 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "capacitor_open = []\n",
+    "for f in freq.f:\n",
+    "        capacitor_open += [poly3(f, 49.43, -310.1, 23.17, -0.1597)]\n",
```
Ttl, line 229: You should be able to replace all these for loops over frequency by just passing `freq.f` as a parameter to the function. For example: `capacitor_open = poly3(freq.f, 49.43, -310.1, 23.17, -0.1597)`

Ttl (changes requested):

biergaizi:
> Interesting. Some have told me the opposite, that is, Open, Short, Load should be connected to the cables directly, not to the Thru.

Thanks for pointing it out. I was confused by the "precision port adapter" mentioned in the Keysight 85033E and thought it was the Thru standard and since it's always used I thought it's a flush Thru. On second thought, it was unrelated. The Thru is defined to be zero-length because there's no Thru standard in this cal kit to begin with.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
jhillairet: Eventually, NumPy also provides convenient polynomial fit functions:

```python
f = np.array([10, 20, 30])
c = [5, 4, 3, 2]  # polynomial’s coefficients in decreasing powers
p = np.poly1d(c)  
print(p(f)) # same as poly3(f, 2, 3, 4, 5)
```

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: @Ttl You're right. I rechecked, it cannot be modeled as a lumped resistor. The unit "Ohm" in offset loss comes from the characteristic impedance. I'll rewrite this section.

biergaizi:
I just force-pushed a new commit. The transmission line attenuation section has been rewritten, the loops and polynomials are also removed is favor of numpy's broadcasting and `poly1d`. Please review the new revision.

Again, it's still a draft, all the code is still untested and the text has not been proofreaded. I'll keep working on it tomorrow. I just found Maury Microwave provides standard definitions in both Keysight and R&S format.

biergaizi:
BTW, why does a force push cause the action bot to remove the existing tag `Calibration`? Is it a bug in the build bot script?

biergaizi:
Still working on it. The calculation so far is still incorrect due to some typos in the code. But the progress so far looks promising.

For verification, I'm now comparing my result with the [`SignalIntegrity` package ](https://github.com/TeledyneLeCroy/SignalIntegrity) (It's GPL'd so I'm obviously not copying any code, just using as a comparison). This is a package for high-speed signal integrity analysis developed by Peter J. Pupalaikis at LeCroy. He has written an entire book as its documentation for this package to explain the derivation behind the code, *S-Parameters for Signal Integrity*, highly recommended. It has a built-in `CalKit` class. The S11 agreement for the offset transmission line is now within 0.005.

Quite unexpectedly, I believe I have now found a bug in `SignalIntegrity` rather than my code. When the short standard is modeled only with an offset loss and delay, `SignalIntegrity`'s return loss is always 0 dB - which is by definition incorrect. Meanwhile the loss calculated by my code from Keysight coefficients now agrees with the numerical loss specified by R&S coefficients. Since the loss is very low (within 0.005), not a big deal, I'll keep fixing and testing my implementation.

biergaizi:
> Quite unexpectedly, I believe I have now found a bug in SignalIntegrity rather than my code. When the short standard is modeled only with an offset loss and delay, SignalIntegrity's return loss is always 0 dB

Found the reason - its documentation is wrong. The unit is actually in ohm/s but documented as Gohm/s. After correcting the order of magnitude, the loss calculated by `SignalIntegrity` makes sense now. The agreement is not exact, but I suspect it's just normal numerical errors. I'll try modeling it as a full transmission line again and compare the results with the delayed attenuator model...

Another mistake in the current version is the definition of loss. Apparently in the R&S, the loss is two-way loss, not one-way...

mhuser:
Alternatively with [`skrf.media.definedAEpTandZ0`](https://scikit-rf.readthedocs.io/en/latest/api/media/generated/skrf.media.definedAEpTandZ0.DefinedAEpTandZ0.html#skrf.media.definedAEpTandZ0.DefinedAEpTandZ0), you can set impedance and attenuation in dB/m/sqrt(Hz) straightforward.

biergaizi:
@mhuser
Thanks for this suggestion. This is already similar to what I'm doing in the unpublished draft, here, I was calculating γl = αl + jβl from Keysight's folmula, and then usisg that as the `gamma` for `Media.DefinedGammaZ0`, then a line with a unity length of one meter is created by `medium.line(d=1, unit='m')` to obtain the S-parameter of this offset transmission line. 

> You can use 1. as relative permittivity ep_reff or velocity factor VF and define the delay when creating the line from the media.

My understanding is that the Keysight approach is explicitly made to avoid any specific assumption of any particular line length, dielectric constant or velocity factor. They're implicitly included as part of the definition in offset delay and offset loss. Then, γl and lossy Z is directly calculated.

I'm now following the derivation in [Application Note 1287-11](https://people.ece.ubc.ca/robertor/Links_files/Files/AN-1287-11.pdf), see equation 2B.5.

mhuser:
This sounds good ! From the equations it seems the proper relative permittivity is already embedded into the coefficient.

biergaizi:
Almost done.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,744 @@
...
+    "transimission line. The medium itself is physically meaningless."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "line_open = medium_open.line(d=1, unit='m')\n",
```
Ttl, line 273: This creates a line with complex characteristic and port impedance. The complex port impedance can be a problem if the S-parameters are passed to the calibration routine which usually assume a fixed real impedance. Also if the S-parameters are exported as touchstone, the S-parameters are not renormalized to a fixed real impedance and the exported network is incorrect. Unfortunately creating a line with characteristic impedance of the medium and given port impedance is not very intuitive (Issue #651). You should set the port impedance `z0` in the media and then generate the line with `z0=medium_open.Z0, embed=True`. It probably doesn't cause that large error in practice since the imaginary part of the characteristic impedance should be quite close to zero.

If you check the `z0` property of the generated standards its complex because of the input port is from this transmission line.

The same issue applies to lines in the later example.

Ttl (changes requested):
Looks quite good. It would be nice to also have some plots of the generated standards. For example phase of open and short or Smith chart whichever looks nicest.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: > It probably doesn't cause that large error in practice since the imaginary part of the characteristic impedance should be quite close to zero.

Thanks for the comment. I'm doing some final tests, one of them is generating a bunch of random parameters and compare SignalIntegrity and my results. I noticed a small numerical disagreement up to 0.006 in the offset line model (I thought it has been solved but apparently not really). This is probably the reason.

biergaizi: Yes! Setting `embed=True` and now I see a 10x reduction in numerical disagreement.

biergaizi: Solved.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,744 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Modeling the Offset Transmission Line\n",
+    "\n",
+    "To correctly model the offset transmission line, one should use the\n",
+    "offset delay, offset loss, and offset $Z_0$ to derive the *propangation\n",
```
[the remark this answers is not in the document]
biergaizi: Fixed.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,744 @@
...
+    "constant* ($\\gamma$) and the complex *charactestic impedance* ($Z_c$)\n",
```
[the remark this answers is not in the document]
biergaizi: I've pun the entire text into a spellchecker, the typos are all fixed now.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,744 @@
...
+    "of the lossy line. Then, an actual transmission line is defined in\n",
+    "those terms.\n",
+    "\n",
+    "The relationship between the offset line parameters and the propangation\n",
+    "constant is given by the following equations:\n",
```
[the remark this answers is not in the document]
biergaizi: Inline citations added.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,744 @@
...
+    "measurements results), you can simply export the S-parameters to Touchstone\n",
+    "files for use in scikit-rf. Similarly, if you're already using a data-based\n",
+    "calibration standard, it should be possible to use its data directly. The\n",
+    "S-parameters may be stored in device-specific file formats, consult your\n",
+    "vendor on whether they can be export as a Touchstone file.\n",
+    "\n",
+    "Finally,\n",
+    "for some non-critical measurements below 1 GHz, sometimes one can assume\n",
+    "the calibration standards are ideal. In scikit-rf, ideal responses can be\n",
+    "conveniently created with three lines of code by defining an ideal\n",
```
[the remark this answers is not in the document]
biergaizi: Solved. I wrote an additional section in the text below on generating ideal responses, and here I added to anchor link to that place.

biergaizi:
Good idea. I'll add some plots soon.

biergaizi:
Plots added.

biergaizi:
Tests performed:

1. The generated SOLT standards by the step-by-step tutorial and code snippet are identical, confirming that there's no typo in the guide.
2. The generated SOLT standard by R&S coefficient is identical to the same physical standard, Maury Microwave 8050CK10, specified in Keysight coefficients, ensuring that the R&S code is correct.
3. The offset transmission line model has been compared with the [SignalIntegrity package](https://github.com/TeledyneLeCroy/SignalIntegrity), package for high-speed signal integrity analysis developed by Peter J. Pupalaikis at LeCroy. When offset delay is within (1 ps, 1000 ps), offset loss is within (1 Gohm/s, 25 Gohm/s), frequency is 100 MHz to 60 GHz, the generated S-parameters agrees to better than 0.001 for both their real and imaginary parts (in setting `calcZc = True` mode, because by default `SignalIntegrity` uses a lossless impedance).

However, when an extremely large (and unpractical) offset delay or loss is specified, e.g. 100 Gohm/s, the difference can be as high as 0.02. I believe this is due to the difference of the physical model used. `SignalIntegrity` calculates $\gamma$ through an intermediate RLGC transmission line step, meanwhile, my implementation calculates $\gamma$ directly. I believe my implementation is "more correct", since it's the formulas given by Keysight and it's also the formula used by Federal Institute of Metrology (METAS), Switzerland. This difference is worth a discussion with the developers of SignalIntegrity. But, in any case, for practical purposes the results are the same.

Thus, I believe this guide is thoroughly checked and it's suitable for merging in its current state.

biergaizi:
The guide is ready for merging. Please give it a final review.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,1116 @@
...
+    "shunt_short = ideal_medium.inductor(inductor_list) ** ideal_medium.short()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `medium.open()` termination is important, a line terminated solely by a\n",
+    "`shunt_capacitor()` without an open produces invalid S-parameters due to\n",
+    "implementation-related reasons."
```
Ttl, line 397: open termination is needed because shunt capacitor creates a two-port and the other port needs to be open. Series capacitor terminated with short could have also been used to get the same result.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,1116 @@
...
+      "text/plain": [
+       "<matplotlib.legend.Legend at 0x72a2f675b4f0>"
+      ]
+     },
+     "execution_count": 17,
+     "metadata": {},
+     "output_type": "execute_result"
+    },
+    {
+     "data": {
```
Ttl, line 889: Results should be stripped from the notebook: https://scikit-rf.readthedocs.io/en/latest/contributing/index.html#contributing-to-the-documentation

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,1116 @@
...
+    "The definition of shunt impedance is identical to the Keysight format.\n",
+    "\n",
+    "But, beware of the units used for the capacitance and inductance! In the\n",
+    "given table, the capacitances are given in $10^{−15} \\text{ F}$, $10^{−15} \\text{ F/Hz}$,\n",
+    "$10^{−15} \\text{ F/Hz}^2$, and\n",
+    "$10^{−15} \\text{ F/Hz}^3$. For Keysight and Anritsu VNAs, they're given in $10^{-15} \\text{ F}$,\n",
+    "$10^{-27} \\text{ F/Hz}$, $10^{-36} \\text{ F/Hz}^2$ and $10^{-45} \\text{ F/Hz}^3$. Inductance\n",
+    "units have the same differences. Always double-check the units before start modeling. To\n",
+    "convert the units from the first to the second format, multiply $x_1$, $x_2$ and $x_3$\n",
+    "by 1000 (don't change the constant term $x_0$). For consistency, we'll use the second format\n",
```
Ttl, line 770: This doesn't calculate correctly. I think the units should be F/GHz? The table also has the same 1e-15*F/Hz units which are way too high when compared to the Keysight coefficients. F/GHz units should be consistent with the current calculations.

Ttl (commented):
Looks very good, only few minor comments.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: You're right, the unit is per GHz.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: Just added your clarification.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: Fixed.

biergaizi:
Updated.

Vinc0110 (approved):
That's a nice guide. Great job!
I'm also looking forward to the implementation of your code snippets as a proper CalKit class (https://github.com/scikit-rf/scikit-rf/issues/610#issuecomment-1126680392)

Ttl (approved):

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,1004 @@
...
+    "$$\n",
+    "\\begin{gather}\n",
+    "\\alpha l = \\frac{\\text{offset loss} \\cdot \\text{offset delay}}{2 \\cdot \\text{ offset }Z_0} \\sqrt{\\frac{f}{10^9}} \\\\\n",
+    "\\beta l = 2 \\pi f \\cdot \\text{offset delay} + \\alpha l \\\\\n",
+    "\\gamma l = \\alpha l + j\\beta l\\\\\n",
+    "Z_c = \\text{offset }Z_0 + (1 - j) \\frac{\\text{offset loss}}{2 \\cdot 2 \\pi f} \\sqrt{\\frac{f}{10^9}}\n",
+    "\\end{gather}\n",
+    "$$\n",
+    "\n",
+    "where $\\alpha$ is the attenuation constant of the line, in nepers per meter, $\\beta$ is the phase constant of the line, in radians per meter, $\\gamma$ is the propagation constant of the line, $l$ is the length of the line, $Z_c$ is the complex characteristic impedance of the lossy line.\n",
```
jhillairet, line 230:
```suggestion
    "where $\\alpha$ is the attenuation constant of the line, in nepers per meter, $\\beta$ is the phase constant of the line, in radians per meter, $\\gamma=\\alpha + j \\beta$ is the propagation constant of the line, $l$ is the length of the line, $Z_c$ is the complex characteristic impedance of the lossy line.\n",
```

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
```diff
@@ -0,0 +1,1004 @@
...
+    "a calibration standard.\n",
+    "\n",
+    "The first part is an \"offset\" lossy transmission\n",
+    "line, defined using three parameters: (1) a real characteristic impedance\n",
+    "of a lossless line - usually the system impedance, but for waveguide\n",
+    "standard, a special normalized value of `1` is used, (2) a delay -\n",
+    "represents its electrical length, given in picoseconds, and (3) a loss.\n",
+    "The loss is given in a somewhat unusual unit - gigaohms per second. With\n",
+    "the three parameters, one can calculate the *propagation constant*\n",
+    "($\\gamma$) and the complex *characteristic impedance* ($Z_c$) of the\n",
```
jhillairet, line 110: I understand that Z_c is distinct from Z_0, in particular because of the discussion below. So I would have added a note saying the difference in notation is intended, such as:
```suggestion
    "($\\gamma$) and the complex *characteristic impedance* ($Z_c$) of the (note that $Z_c$ may be different from the port impedance $Z_0$)\n",
```

jhillairet (approved):
Very nice tutorial. I appreciate the call for making additional documentation contributions ;) It also assumes that readers are not Python/Numpy experts, which is a good thing in my opinion (that we sometimes forgot).

A minor thing: I suggest shortening the title of the notebook, which is a little long as this is the title which will be used to generate the Table of Content on the left panel of the documentation:
image

`SOLT Calibration Standards` should directly appear after `SOLT`, but `Coaxial SOLT Calibration Standards` or `Creating SOLT Calibration Standards` are nice too. Up to you.

The "confusing" part on the creation of the transmission line length may change in the future (see discussion in #654 and other)


> "A series medium.capacitor() terminated by a medium.short() "

I would have used this solution first, and eventually say a word about shunt_capa terminated with open, to make it similar to the inductor case.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: I added an slightly longer explanation in the next paragraph.

doc/source/examples/metrology/Creating Network Definitions for SOLT Calibration Standards.ipynb
biergaizi: Taken.

biergaizi:
Guide has been updated with your suggested text changes.

I changed the title to "SOLT Calibration Standards Creation", so it shows up next to SOLT.

jhillairet (approved):

jhillairet:
Is OK to be merged?

biergaizi:
> Is OK to be merged?

Yes, I consider it finished.

jhillairet:
The generated documentation is here: https://scikit-rf.readthedocs.io/en/latest/examples/metrology/SOLT%20Calibration%20Standards%20Creation.html

Thank you @biergaizi

Issue #615, opened 2022-03-09
Support for old Python releases


Vinc0110:
The proposed changes in PR #614 require the latest major release of NumPy (>= v1.22.0), in which they dropped support for Python < 3.8. 
Scikit-rf is currently still supporting Python 3.6 and 3.7, but that would stop if we decide to depend on NumPy v1.22.

Python 3.6 is already at its end-of-life, but Python 3.7 is still a thing:

https://devguide.python.org/#status-of-python-branches

![image](https://user-images.githubusercontent.com/63644854/157493850-57148b3b-b7aa-46ab-990f-efb3bbd60cd9.png)

For PR #614, I can probably find a workaround and live with the oldish numpy, but is it really necessary?

What is the general opinion about keeping the support of Python 3.6 and 3.7? Does anyone care?

keikawa:
How about including a question about the version of Python in the next survey?, but It's still March now...

jhillairet:
Two years ago, in the 2020's survey, around 70% of the user were using Python 3.7. Since I didn't ask for the python version in the 2021's version of the survey, I can't compare...

However, as a lot (most?) of the users are occasional Python users, I'm not sure if they are using the last version of Python. In my work, for example, the default Python version, (which was upgraded last year), is still 3.7 (and I guess it will remain like this for a certain time...).

JAnderson419:
I definitely have some virtual environments on 3.6/3.7 with old versions of skrf that are "good enough" for opening and looking at s-parameters, maybe doing a few calculations, and some plotting. That being said, I think there is some advantage to not requiring support for 5 year old versions of Python, especially in the scientific domain when trying to provide a good (or good enough) alternative to commercial software that often has yearly release cycles.

Instead of supporting every Python version that isn't EOL, an intermediate approach would be to follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) since it seems to be a common one among many of the scientific libraries (including skrf dependencies). This requires the greater of 42 month or two minor version support for Python and 24 month or 3 minor version support for numpy. Currently, this would mean Python 3.8+ and numpy 1.19+ (we are requiring 1.19.5 currently), with that changing to numpy 1.20 in June. 

Adopting NEP 29 still wouldn't help the VF improvements that rely on numpy 1.22. That could be solved by creating nightly/rc branches for future releases that advanced users could install, but that might be more work than it is worth unless there are a large number of corporate users who require support of old versions for a long time...

[...]

PR #668, opened 2022-05-18
Flake8 Debugging Scratchpad, Please Dont Merge

jjoonathan:
Python 3.10 removes the collections.Iterable alias to collections.abc.Iterable.


jjoonathan:
Let's see if adding "dataclasses" to requirements.txt suffices for 3.6 support. It's a small dependency and it can go away once 3.6 support isn't needed.

jjoonathan:
Ah, probably needs to be in both requirements.txt and requirements_dev.txt

jjoonathan:
Hmm I'm out of obvious ideas on where to add the dependency so that the tests pick up. I'll take another look tonight, but if you know off the top of your head I'd appreciate a hint :)

jhillairet:
> Hmm I'm out of obvious ideas on where to add the dependency so that the tests pick up. I'll take another look tonight, but if you know off the top of your head I'd appreciate a hint :)

Hum, strange indeed. The package is correctly installed and the linter flake8 does produce a SyntaxError where it should not... 

Anyway, I think we could safely drop Python3.6 support now (related #615).

jhillairet:
by the way, the name of the PR does not reflect the main purpose of your work, isn't? (adding SweepPlan etc )

jjoonathan:
Yes, I originally intended these to be two different PRs but I am out of practice with github and forgot that the SweepPlan commit would be auto added to the PR if it lived on the same branch. Next time I'll use different branches for preparing different PRs. In any case, the python 3.10 support is trivial compared to the Sweep Planning, so it makes sense to change the name (done).

I was able to reproduce the issue locally, but I never figured out exactly why flake8 wasn't seeing the dataclasses package even though python3.6 REPL could. Yeah, dropping 3.6 support sounds reasonable and would certainly make this easier :)

Vinc0110:
It seems `dataclasses` was added with Python 3.7, so it won't work in Python 3.6: https://docs.python.org/3/library/dataclasses.html

I'm in favor of dropping Python 3.6, but if you need to get your code working for 3.6 (for whatever reason), you could also implement the classes the usual way without the automatic generation by `dataclasses`, right?

Vinc0110:
In the link I posted above, they import and use `dataclass` like this:

```
from dataclasses import dataclass

@dataclass
class InventoryItem:
    """Class for keeping track of an item in inventory."""
    name: str
```

Does that maybe solve the error?

jjoonathan:
Yes, commits 3,4,5 in this PR were attempts to backport support for python3.6 by using the `dataclasses` package. Unfortunately, flake8 doesn't seem to pick up the package even if it is present. I have managed to reproduce the problem but not fix it yet. Here is the docker setup I use:
```
sudo docker run --rm -it ubuntu:focal

cat <<END_DOC >/tmp/instl
set -ex
export DEBIAN_FRONTEND=noninteractive
export TZ=Etc/UTC
apt update
apt install -y software-properties-common build-essential
add-apt-repository -y 'ppa:deadsnakes/ppa'
apt update
apt install -y python3.6 python3.6-dev python3.6-venv
python3.6 -m ensurepip --default-pip
END_DOC
bash /tmp/instl
```

and here is a demonstration of the mismatch. The python3.6 REPL sees `dataclasses` but flake8 doesn't
```
root@21db1252a854:/scikit-rf# flake8 --version
4.0.1 (mccabe: 0.6.1, pycodestyle: 2.8.0, pyflakes: 2.4.0) CPython 3.6.15 on
Linux



root@21db1252a854:/scikit-rf# flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
./skrf/vi/vna/hp8510c_sweep_plan.py:1:14: E999 SyntaxError: invalid syntax
import dataclasses
             ^
1     E999 SyntaxError: invalid syntax
1



root@21db1252a854:/scikit-rf# python3.6
Python 3.6.15 (default, Apr 25 2022, 01:55:53) 
[GCC 9.4.0] on linux
>>> import dataclasses
>>>
```

jjoonathan:
Commits 6,7,8,9 revert the backport attempt and instead drop the 3.6 test. It looks like flake8 still doesn't like the dataclasses import, though, even in python3.7, which has dataclasses by default. I'll try using the `from dataclasses import dataclass` form and see if that helps. On the REPL, `import dataclasses` works just fine in python 3.6 (with dataclasses package), 3.7 (natively), and 3.10 (natively)-- this seems to be a flake8 issue, but I don't know enough about flake8 configuration to have immediate ideas on how to fix it. I'm looking into it.

jjoonathan:
Huh, flake8 doesn't like dataclasses even in python3.7 where it is "batteries included."
`sudo docker run --rm -it ubuntu:focal`
```
# 3.7
cat <<END_DOC >/tmp/instl
set -ex
export DEBIAN_FRONTEND=noninteractive
export TZ=Etc/UTC
apt update
apt install -y software-properties-common build-essential
add-apt-repository -y 'ppa:deadsnakes/ppa'
apt update
apt install -y python3.7 python3.7-dev python3.7-venv
python3.7 -m ensurepip --default-pip
END_DOC
bash /tmp/instl
```
```
root@538b103f3b65:/# python3.7 -c 'import dataclasses; print("ok")'
ok
root@538b103f3b65:/# python3.7 -c 'from dataclasses import dataclass; print("ok")'
ok
``

**Comment by Vinc0110** [comment]
Strange! The bug tracker does not seem to include any related issue: https://github.com/PyCQA/flake8/issues

I'd probably get rid of it and implement the classes the old fashioned way...

By the way: Maybe we should not drop support of Python 3.6 silently in this PR and rather create a new one for that. @jhillairet what do you think?

**Comment by jjoonathan** [comment]
Yeah, this PR is 2 different ideas + messy flake8 debugging. It makes sense to separate things out.

**Comment by jjoonathan** [comment]
Also, I don't think dropping python3.6 support is indicated anymore now that we have seen the same issue on python 3.7. Even though python 3.7 natively includes dataclasses, flake8 still flips out.

**Comment by jjoonathan** [comment]
Huh, the minimal reproduction seems compatible with python 3.7. I am beginning to suspect flake8 objects to something else and just flags the first line of code.

**Comment by jjoonathan** [comment]
Please don't merge! I am using this branch as a scratchpad to debug CI.

**Comment by jjoonathan** [comment]
Ok, now that the minimal repro passes on 3.7 let's see if it passes on 3.6

**Comment by jjoonathan** [comment]
Yes, 3.6 "minimal repro" works. This strongly points to flake8 objecting to something other than what it claims. Commencing binary search to figure it out.

**Comment by jjoonathan** [comment]
Ok, flake8's real problem is in the first half of SweepPlan.

**Comment by jjoonathan** [comment]
First quarter SweepPlan passes.

**Comment by jjoonathan** [comment]
Ok, the true objection is somewhere in matches_f_list

**Comment by jjoonathan** [comment]
Ok, true objection is in first half of SweepPlan.matches_f_list

File: skrf/vi/vna/hp8510c_sweep_plan.py [hunk 1]
```diff
@@ -72,4 +72,4 @@ def matches_f_list(self, hz):
         True if the frequencies sweepned by self equal those in the list hz.
         """
         plan_hz = np.array(sorted(self.get_hz()))
-        return good
+        return True
```
Vinc0110, line 75: Love it :)

jjoonathan: haha, yeah, the binary search ax chopped a bit too hard that time

jjoonathan:
Ok, I think I might have found the real issue. It's really unfortunate that flake8 sent us on the wild goose chase about dataclasses, because I'm pretty sure the actual problem was just that I used the {expression=} fstring syntax on line 70, and that's a python3.8+ thing.

`print(f'{len(plan_hz)=} {len(hz)=}')`

jjoonathan:
Yes, `print(f'{len(plan_hz)=} {len(hz)=}')`  was the real problem. Sigh.

jjoonathan:
I'll confirm the fix here and then open up a new PR.

Vinc0110:
Didn't even know things like `print(f'{len(plan_hz)=} {len(hz)=}')` exist. 
Great job catching the problem!

jhillairet:
> Yes, `print(f'{len(plan_hz)=} {len(hz)=}')` was the real problem. Sigh.

ho. I run into recently also in https://github.com/scikit-rf/scikit-rf/pull/637/commits/86b92e38cb0611db28c73599b91b7939ed02d1e7, but it was much easier to find out than here! Well done.

jjoonathan:
Yeah, it's a convenient little bit of syntax! Fortunately '=}' is quite greppable so next time should be much easier. In any case, this thread has served its purpose so time to close.

PR #669, opened 2022-05-20
Fixed NanoVNA backend configuration for PyVISA-py

Vinc0110:
The current initialization for the NanoVNA V2 does not work on machines with multiple VISA backends installed (IVI, PyVISA-py, ...). 

For the serial communication with the NanoVNA V2, the PyVISA-py backend is required, but the identifier was wrong (`py` instead of `@py`). This causes an error if PyVISA-py is not the default backend.

I might add another feature to the instrument class, so I'll create the PR as a draft for now.


Vinc0110:
Ok, done :)

jhillairet (approved):

jhillairet:
OK for merging

PR #671, opened 2022-05-21
Mark Circuit Ports with attribute instead of name

### LINKED ISSUE #670: Circuits should not rely only on the network name to detect a port
To define a external port in circuits Circuit.Port() can be used and the docu states `Passing the frequency and name is mandatory. Port name must include the word ‘port’ inside. (ex: ‘Port1’ or ‘port_3’)`. Unfortunately all functions rely only on this string to determine if a network is a port or not, e.g. port_indexes():
```
if 'port' in str.lower(ntw.name):
                port_indexes.append(idx_cnx)
```
I use the circuit class to work with S-parameters from touchstone files and recently I loaded a file like `module123_sim_export.s123p` which led to a network.name containing 'port' and therefore the resulting circuit was completely wrong.

I think, using the network name as an identifier for ports is probably not a good idea at all, but I understand that it is a simple solution, because no additional parameter or similar is needed. If you want to keep it like this, maybe add some more metrics/restriction, e.g. only one-ports can be ports or the ports name has to start with `Port_`instead just containing `port`somewhere in the name.

Ttl:
Fixes #670

### DIFF
--- a/doc/source/tutorials/Circuit.ipynb
+++ b/doc/source/tutorials/Circuit.ipynb
@@ -125,7 +125,6 @@
     "delay_load = tline_media.delay_load(rf.zl_2_Gamma0(Z_0, Z_L), theta, unit='deg', name='delay_load')\n",
     "\n",
     "# the input port of the circuit is defined with the Circuit.Port method\n",
-    "# In order for Circuit() to recognize the Network as a \"port\", its name must contains the word 'port':\n",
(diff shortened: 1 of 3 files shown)


jhillairet (approved):
OK.

As only the number associated to the portname is useful now, I wonder if changing the API for `Port(frequency, port_index)` would make more sense. TBD, eventually in a future PR.

jhillairet:
OK for merging

PR #672, opened 2022-05-21
Fix write_touchstone with complex or non-equal z0

Ttl:
Use network port impedance as the reference impedance if possible.
Add `r_ref` argument that can be used to specify reference impedance to be used for renormalization.

Also fixes `write_z0=True` not being able to be read back.


skrf/io/touchstone.py
```diff
@@ -507,18 +507,18 @@ def get_noise_data(self):
...
         status : boolean
-            True if the Touchstone file has been produced by HFSS
+            True if the Touchstone file might have complex port impedances for each frequency.
             False otherwise
         """
         if self.comments is None:
             return False
 
-        if 'exported from hfss' in str.lower(self.comments):
+        if 'hfss' in str.lower(self.comments) or 'not renormalized' in str.lower(self.comments):
```
jhillairet, line 521: note that you could catch a comment mentioning hfss, but not necessarily a file having (that was true before of course, but a bit less probable)

skrf/network.py
```diff
@@ -1977,6 +1977,8 @@ def read_touchstone(self, filename: Union[str, TextIO],
         # set z0 before s so that y and z can be computed
         if touchstoneFile.is_from_hfss():
             self.gamma, self.z0 = touchstoneFile.get_gamma_z0()
+            # This is the s_def that HFSS uses
+            self.s_def = 'traveling'
```
jhillairet, line 1981: not necessarily, I think. There is an option in HFSS, as well as in ANSYS Circuit, to switch from a travelling to power wave definition (matching on conjugate match). I'll check that.

skrf/io/touchstone.py
```diff
@@ -507,18 +507,18 @@ def get_noise_data(self):
 
     def is_from_hfss(self):
         """
-        Check if the Touchstone file has been produced by HFSS.
+        Check if the Touchstone might have complex port impedances for each frequency.
```
jhillairet, line 510: Exporting the characteristic impedances (which might be complex valued) is an option in HFSS, which can be unticked. ,

jhillairet (commented):
the self -> ntwk makes sense. I wonder what it wasn't like that before..

skrf/io/touchstone.py
Ttl: It shouldn't matter since the function reading the port impedances will return the reference impedance if it doesn't find them. It would be better for this function to actually check for the impedances instead, since it's used to check if the file might have port impedances.

skrf/network.py
Ttl: I looked at HFSS options and didn't find a setting for it a while ago. It might be there somewhere though, but this should be the default.

skrf/io/touchstone.py
Ttl: I changed it since this function is also used to check for port impedances in files exported by skrf. The name doesn't really match anymore though.

skrf/io/touchstone.py
```diff
@@ -149,7 +152,7 @@ def __init__(self, file: typing.Union[str, typing.TextIO],
         self.gamma = []
         self.z0 = []
 
-        if self.is_from_hfss:
+        if self.hfss_port_impedances:
```
jhillairet, line 155: I think this name is a bit misleading for a Boolean. `has_hfss_port_impedances` ?

skrf/network.py
jhillairet: It's in the Circuit setting, but I never know if it affects Circuit only, or both Circuit and HFSS
image

jhillairet (approved):

PR #674, opened 2022-05-21
Fix compatibility with python 3.10

jjoonathan:
Cherrypicking this into a separate, clean PR

### DIFF
--- a/qtapps/skrf_qtwidgets/util.py
+++ b/qtapps/skrf_qtwidgets/util.py
@@ -1,6 +1,6 @@
 import re
 import copy
-from collections import Iterable
+from collections.abc import Iterable
 import zipfile
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


jjoonathan:
Python 3.10 removes the collections.Iterable alias to collections.abc.Iterable. This patch switches from the alias, which is no longer present in 3.10, to the full path, which does work in 3.10.

Vinc0110 (approved):
Looks good. Thanks for splitting the PR!

jhillairet:
`Iterable` is also used in networkSet.py, so it may be affected as well ?

jjoonathan:
I think the one in networkSet.py is ok because it already imports from abc, it doesn't use the alias that went away. While I was grepping around I did find another use of the alias, though. Here's the patch.

PR #675, opened 2022-05-21
HP8510C Sweep Plan

jjoonathan:
HP8510 sweeps now go through a planning phase to allow for some cleverness. The cleverness is required to handle sweeps with multiple ranges/spacings in a reasonable time.


skrf/vi/vna/hp8510c_sweep_plan.py
```diff
@@ -0,0 +1,186 @@
...
+import numpy as np
+from typing import List,Union
+
+class SweepSection:
+    def get_hz(self):
+        raise NotImplementedError()
+    def apply_8510(self, hp8510c):
+        raise NotImplementedError()
+    def mask_8510(self, network):
+        return network
```
jhillairet, line 11: If the method `SweepSection` should not be instantiated directly but only be inherited, I guess it should be defined as an abstract class, no? Something like that?

```suggestion
from typing import List,Union
from abc import ABC, abstractmethod

class SweepSection(ABC):
    @abstractmethod
    def get_hz(self):
        pass
    @abstractmethod
    def apply_8510(self, hp8510c):
        pass
    def mask_8510(self, network):
        return network
```

skrf/vi/vna/hp8510c_sweep_plan.py
```diff
@@ -0,0 +1,186 @@
...
+# test_sp_1 = SweepPlan.from_hz(test_hz_1)
+# assert test_sp_1.matches_f_list(test_hz_1)
+# 
+# test_hz_2 = np.concatenate(([1,2,3], np.linspace(100,1000,1001), [9999]))
+# test_sp_2 = SweepPlan.from_hz(test_hz_2)
+# assert test_sp_2.matches_f_list(test_hz_2)
+# 
+# test_hz_3 = np.concatenate(([1,2,3], np.linspace(100,1000,1001), [9999]))
+# test_sp_3 = SweepPlan.from_hz(test_hz_3)
+# assert test_sp_3.matches_f_list(test_hz_3)
```
jhillairet, line 186: to be removed or better to be moved into a proper test

skrf/vi/vna/keysight_fieldfox.py
```diff
@@ -1,5 +1,6 @@
 import warnings
-from collections import OrderedDict, Iterable
+from collections import OrderedDict
+from collections.abc import Iterable
```
jhillairet, line 3: does it lead to an error if you remove these imports?

skrf/vi/vna/keysight_pna.py
```diff
@@ -1,5 +1,6 @@
 import warnings
-from collections import OrderedDict, Iterable
+from collections import OrderedDict
+from collections.abc import Iterable
```
jhillairet, line 3: same comment, and for rs_zva.py

jhillairet (approved):

Issue #677, opened 2022-05-22
Passivity enforcement not working correctly


Vinc0110:
Related to #524 

Passivity enforcement of vector fitted models with `VectorFitting.passivity_enforce()` does not always converge or breaks after 2 steps. Even worse, it totally destroys the fit.

Example:
```
import skrf
import numpy as np

nw = skrf.Network('./skrf/tests/Agilent_E5071B.s4p')
print(nw.is_passive()) # returns True

vf = skrf.VectorFitting(nw)
vf.vector_fit(2, 26)
freqs = np.linspace(0, 5e9, 101)

vf.is_passive() # returns False
vf.passivity_test() # returns array([[2.91439290e+08, 4.01223463e+08]])
vf.plot_s_db(freqs=freqs) # see Fig. 1
vf.plot_s_singular(freqs=freqs) # see Fig. 2

vf.passivity_enforce()
vf.plot_passivation() # see Fig. 3

vf.is_passive() # still returns False
vf.passivity_test() # returns huge array
vf.plot_s_db(freqs=freqs) # see Fig. 4
vf.plot_s_singular(freqs=freqs) # see Fig. 5
```

![fit_before](https://user-images.githubusercontent.com/63644854/169692415-208fcc9b-a491-404e-a7c7-09c56f02dc6d.png)
**Fig. 1**: Fit looks alright

![singular_before](https://user-images.githubusercontent.com/63644854/169692466-3db1e081-124c-4a55-8772-77bae31b4a82.png)
**Fig. 2**: Sigma1 is clearly greater than 1 at low frequencies

![passivation](https://user-images.githubusercontent.com/63644854/169692478-24654f81-e168-4f23-a61a-f0d6c6684ec0.png)
**Fig. 3**: Passivation seems to be done after only 2 steps (max singular values smaller than 1)

![fit_after](https://user-images.githubusercontent.com/63644854/169692489-0a150d06-f761-44c6-a72a-813434d71783.png)
**Fig. 4**: Disaster

![singular_after](https://user-images.githubusercontent.com/63644854/169692501-cd2aa936-3018-4be7-97c4-ab77b2e87df7.png)
**Fig. 5**: Sigma 1 is all over the place

PR #678, opened 2022-05-22
Vector Fitting: Bug fix for #677 and #702

Vinc0110:
Fixes issue #677 (and #702). 
Also a warning is printed in case the passivity enforcement did not succeed (when `n_samples` was too small).

The problem was an incorrect choice of the evaluation frequencies for the passivity evaluation.

Quote from the paper [1]:
`First, a dense set of frequencies
 is determined from
dc up to about 20% above the highest relevant frequency.
This highest relevant frequency is the maximum of the highest
crossing from a nonpassive to a passive region on one hand and
the maximum frequency of interest on the other hand.`

The second condition was missing in the code.

[1] T. Dhaene, D. Deschrijver and N. Stevens, "Efficient Algorithm for Passivity Enforcement of S-Parameter-
            Based Macromodels," in IEEE Transactions on Microwave Theory and Techniques, vol. 57, no. 2, pp. 415-420,
            Feb. 2009, DOI: 10.1109/TMTT.2008.2011201.


Vinc0110:
So I fixed one part, but I'm not sure if this really fixes all the problems. It still seems a bit strange for certain setups. I need to check it some more.

Vinc0110:
This PR ended up as a major revision of the passivity enforcement routine. It now works much better and the routine offers a bit more control with the new parameters `n_samples` and `f_max`. Issue #677 will is solved for `n_samples=500`.

I'll leave it for now and will merge soon, if no one complains.

Vinc0110:
Passivity assessment and enforcement now works for most of my test cases. Enforcement still sometimes fails if the fit is terribly non-passive, which is probably a natural limitation. A massive number of samples would be required, which quickly becomes impractical. A UserWarning is issued, suggesting to re-fit the model with fewer poles and/or without the constants.

jhillairet (approved):

PR #687, opened 2022-06-05
fix #684 phase jumps in 2xthru deembedding

### LINKED ISSUE #684: 180° phase jumps with IEEEP370_SE_NZC_2xThru deembedding
While using IEEEP370_SE_NZC_2xThru deembedding to extract side1 and side2 of a 100mm CPWG line measurement, 180° phase jumps appeared in insertion loss, inducing corresponding phase error in deembedding results:
image

These are caused by the computation of `e01` and `e10` using `sqrt`. The results of `sqrt` could be either positive or negative and the correct solution shall be picked.

This is done here:
https://github.com/mhuser/scikit-rf/blob/b7d026cbf513924e07c97274aab311d1e54f454f/skrf/calibration/deembedding.py#L1152-L1161
and here:
https://github.com/mhuser/scikit-rf/blob/b7d026cbf513924e07c97274aab311d1e54f454f/skrf/calibration/deembedding.py#L1163-L1172

The test assume the sign invert at each positive slope change of the phase. This is correct, but calibration error and measurement noise can cause positive spikes of small amplitudes that trigger the sign inversion wrongly, this is what happens on above plot:
image

mhuser:
- Remove 180° phase jumps in IEEEP370_SE_NZC_2xThru in presence of phase noise thanks to @Ttl elegant solution
- add a corresponding test case
- add paper reference for deembedding algorithm

This fixes #684


Ttl (approved):
Ok.

Not related to just this PR but note that in tests `setUp` function is always run before each test. A long setup that loads a lot of Networks and does calculations that are not needed for each test can be quite slow.

mhuser:
> Networks and does calculations that are not needed for each test can be quite slow.

Indeed this is a good point and especially true for the ZC 2xthru method.
I will initialize the `dm` inside the test to speed up things

mhuser:
Ok to merge ?

jhillairet:
For me, yes.

PR #694, opened 2022-06-09
Minor changes in IEEE P370 doc

jhillairet:
- added some additional content in the introduction and some figures (@mhuser could you check it's OK?)
- use coherent terminology for labelling fixtures and DUT (following the one proposed in the standard)
- fix sphinx rendering of the lists

 rendered page: https://scikit-rf--694.org.readthedocs.build/en/694/examples/networktheory/IEEEP370%20Deembedding.html


mhuser (approved):
Great, the notebook is more clear and follow more closely the standard with this notation!

There is however a pitfall:

The svg are clear and neat and following the standard but  the Matlab code ported to `skrf`, h is flipping FIX-2 in such a way DUT = `FIX-1.inv ** FIX-DUT-FIX ** FIX-2.inv` which is not correct according to the standard labelling.

I wonder if we should make FIX-2 compliant to the standard and svg so the equation would become `FIX-1.inv ** FIX-DUT-FIX ** FIX-2.flipped().inv`

doc/source/examples/networktheory/IEEEP370 Deembedding.ipynb
```diff
@@ -33,12 +48,10 @@
    "source": [
     "import skrf as rf\n",
     "import matplotlib.pyplot as plt\n",
-    "from skrf.calibration import IEEEP370_SE_NZC_2xThru\n",
-    "from skrf.calibration import IEEEP370_MM_NZC_2xThru\n",
-    "from skrf.calibration import IEEEP370_SE_ZC_2xThru\n",
-    "from skrf.calibration import IEEEP370_MM_ZC_2xThru\n",
-    "from skrf.media import MLine\n",
     "import numpy as np\n",
+    "from skrf.calibration import (IEEEP370_SE_NZC_2xThru, IEEEP370_MM_NZC_2xThru,\n",
```
biergaizi, line 52: I suggest changing `skrf.calibration` to `skrf.deembedding`. In scikit-rf, we clearly distinguish between calibration (almost complete error correction at the VNA measurement plane) and deembedding (partial error correction to remove fixture effects).

IEEEP370 is a deembedding method, not a calibration method. Thus, it should be imported from `skrf.deembedding`, not `skrf.calibration`. The fact that `skrf.calibration` also works is only an implementation detail and goes against the official API documentation, which says it's located under `skrf.deembedding`, thus importing from `skrf.calibration` should be discouraged in the *official* documentation.

 (I already opened a PR #696 but I found it creates a conflict with this PR, so it's better to change it here instead).

jhillairet:
> I wonder if we should make FIX-2 compliant to the standard and svg so the equation would become `FIX-1.inv ** FIX-DUT-FIX ** FIX-2.flipped().inv`

I think it should, to match the standard labelling proposal.

mhuser (approved):

jhillairet:
@mhuser , shall I merge this one as you are working on the notebook too?

mhuser:
> shall I merge this one as you are working on the notebook too?

I think yes, I will be soon finishing the first proposal so hopefully, it will not last very long with the non-flipped numbering.

PR #696, opened 2022-06-10
IEEEP370 example: import from skrf.deembedding, not calibration.

biergaizi:
In scikit-rf, we clearly distinguish between calibration (almost complete error correction at the VNA measurement plane) and deembedding (partial error correction to remove fixture effects).

IEEEP370 is a deembedding method, not a calibration method. Thus, it should be imported from `skrf.deembedding`, not `skrf.calibration`. The fact that `skrf.calibration` also works is only an implementation detail and goes against the official API documentation, and it should be discouraged.

### DIFF
--- a/doc/source/examples/networktheory/IEEEP370 Deembedding.ipynb
+++ b/doc/source/examples/networktheory/IEEEP370 Deembedding.ipynb
@@ -33,10 +33,10 @@
    "source": [
     "import skrf as rf\n",
     "import matplotlib.pyplot as plt\n",
-    "from skrf.calibration import IEEEP370_SE_NZC_2xThru\n",
-    "from skrf.calibration import IEEEP370_MM_NZC_2xThru\n",
-    "from skrf.calibration import IEEEP370_SE_ZC_2xThru\n",
-    "from skrf.calibration import IEEEP370_MM_ZC_2xThru\n",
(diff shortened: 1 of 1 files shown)


biergaizi:
Oops, there's a merge conflict with #694. I'll just request a chance in @jhillairet's pull request instead. My PR is unnecessary.

biergaizi:
I was wrong. Both calibration and deembedding classes are under the skrf.calibration namespace, as `skrf.calibration.calibration` and `skrf.calibration.deembedding`. Meanwhile, `skrf.calibration.deembedding` is also accessible via `skrf.deembeeding` and both calibration and deembedding classes are also accessible directly as `skrf.calibration`.

So quite confusingly, there are at least three ways to access `IEEEP370_SE_NZC_2xThru()`.

1. `from skrf.calibration.deembedding import IEEEP370_SE_NZC_2xThru`
    Theoretically, the proper way per API documentation.

2. `from skrf.calibration import IEEEP370_SE_NZC_2xThru`
    Currently used in the documentation.

3. `import skrf; foo = skrf.deembedding.IEEEP370_SE_NZC_2xThru()`

Close PR as being incorrect.

jhillairet:
OK. Indeed you are right in the fact it could be confusing. 

We should agree on the "official" way to import it. The solution 2 seems fine to me, but of course this could be discussed.

PR #701, opened 2022-06-11
Attempt to simplify media line creation

### LINKED ISSUE #651: Creating a line with characteristic impedance of the media is confusing

mhuser:
Attempt to fix #651 

`Media.line` creation should follow these a, b, c, and d cases depending on the parameters `Z0` and `z0` of `Media` and `line`. If one of these parameters is not `None` in both `Media` and `line`, `line` has priority.

| | Media port impedance |  Line port impedance | Expected resulting network port impedance and normalization |
|---|---|---|---|
| a | `media.z0 == None` | `line.z0 == None` | `z0 = media.Z0`, no renormalization required |
| b | `media.z0 == None` | `line.z0 == z2` | `z0 = line.z0 = z2`, renormalization media.Z0 to z2 |
| c | `media.z0 == z1` | `line.z0 == None` | `z0 = media.z0 = z1`, renormalization media.Z0 to z1 |
| d | `media.z0 == z1` | `line.z0 == z2` | `z0 = line.z0 = z2`, renormalization media.Z0 to z2 |

This should keep functioning: `media.line(line_l, 'm', embed=True, z0=media.Z0)` , but `embed` parameter should  issue a `FutureWarning` to indicate future deprecation of this behaviour if set to `True`.

### DIFF
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -847,7 +847,14 @@ class Media(ABC):
         result.s = \
                 npy.array([[s11, s21],[s21,s11]]).transpose().reshape(-1,2,2)
 
+        # todo: some magic here to make embed emmit a warning but keep
+        # `media.line(line_l, 'm', embed=True, z0=media.Z0)` working for
+        # backward compatibility
         if embed:
+            warnings.warn('In a future version,`embed` will be deprecated.\n'
+                          'Port impedance of the line network will be media or line Z0 embedded into media or line z0.\n'
+                          'Line Z0 or z0 will have priority on media if not set to None.',
+              FutureWarning, stacklevel = 2)
             # Use the same s_def here as the line to avoid changing it during
             # cascade.
             result = self.thru(s_def='traveling')**result**self.thru(s_def='traveling')


jhillairet:
A notebook documenting since table would be great. Ideally with some figures (I love figures in doc)

skrf/media/media.py
```diff
@@ -808,9 +810,12 @@ def line(self, d: NumberLike, unit: str = 'deg',
         unit : ['deg','rad','m','cm','um','in','mil','s','us','ns','ps']
                 the units of d.  See :func:`to_meters`, for details
         z0 : number, string, or array-like or None
-            the characteristic impedance of the line, if different
+            the port impedance of the line, if different
```
Ttl, line 813: I think this change causes too many issues with the existing code, after this change old code gives different answers. Not every user is going to follow skrf development and is not going to be aware of this change. I really want to avoid breaking backwards compatibility. `z0` should still be the characteristic impedance and port impedance should have a different name.

skrf/media/media.py
```diff
@@ -833,24 +838,74 @@ def line(self, d: NumberLike, unit: str = 'deg',
...
+        
+        else:
+            # priority given to line impedances on media impedances.
+            z_port = self.z0 if z0 is None else z0
+            z_char = self.Z0 if Z0 is None else Z0
+            
+            # z shape has to be fixed for renormalize to works properly
+            # or it will trigger
+            # AttributeError: 'bool' object has no attribute 'any'
+            # because of `if (self.z0 != z_new).any():` on line 3033
```
Ttl, line 876: This should be fixed in `renormalize` by changing `.any()` to `npy.any(...)`.

skrf/media/media.py
```diff
@@ -833,24 +838,74 @@ def line(self, d: NumberLike, unit: str = 'deg',
...
+    
+            # transmission line
+            theta = self.electrical_length(self.to_meters(d=d, unit=unit))
+            s11 = npy.zeros(self.frequency.npoints, dtype=complex)
+            s21 = npy.exp(-1*theta)
+            result.s = \
+                npy.array([[s11, s21],[s21,s11]]).transpose().reshape(-1,2,2)
+            
+            if not (z_renormalize is None):
+                result.renormalize(z_renormalize, s_def='traveling')
```
Ttl, line 906: There is another renormalize just after this one for s_def change. It can be removed and `s_def` set here (and other if branch) to avoid renormalizing twice.

skrf/media/media.py
```diff
@@ -833,24 +838,74 @@ def line(self, d: NumberLike, unit: str = 'deg',
...
+                s11 = npy.zeros(self.frequency.npoints, dtype=complex)
+                s21 = npy.exp(-1*theta)
+                result.s = \
+                        npy.array([[s11, s21],[s21,s11]]).transpose().reshape(-1,2,2)
+                z_renormalize = fix_z0_shape(self.z0, self.npoints, 2)
+                result.renormalize(z_renormalize, s_def='traveling')
+            else:
+                raise ValueError('Missing z0 or Z0 impedance specification')
+        
+        else:
```
Ttl, line 868: There is a lot of duplicated code here. This should be simplified.

jhillairet:
While this is true that breaking backward compatibility should be avoided, the situation is becoming more and more confusing for users... 

I hear the suggestion made in https://github.com/scikit-rf/scikit-rf/issues/717#issuecomment-1159564970 which has some sense : the `Media` characteristic impedance should be the physical ones (and should not be set to something else) for physical medias (Coax, waveguides, etc) and defined for `DefinedGammaZ0`. Then if ones wants to renormalize to a different impedance, he can use later the renormalization methods. While maybe verbose, it is also clearer, I think, in the workflow. This could be the desired API for version 1.0, and before that release, create some warnings and compatibility fixes until reaching the v1.0.

Would this way works for all cases?

skrf/media/media.py
mhuser: Thanks for your feedback @Ttl , it is much appreciated!

This draft is intended as an exploratory work, to see how scikit-rf would be affected by the changes proposed in #651 , it will probably never be merged but enable others to chime in and provide feedback.

Here are first hands statements that I will later post on the issue:
A good example is found in test_calibration.py. A waveguide media is defined here with parameter `z0`.
https://github.com/scikit-rf/scikit-rf/blob/ac3aedfe0228615ef10da3e25212c7908059e6e0/skrf/calibration/tests/test_calibration.py#L25

Tough a dimension is specified, the physical characteristic impedance is never used (no renormalization or mismatch, the `Z0` is just like overwritten by parameter `z0`.

This media (or the losseless variant) is later used to define calibration standards like
https://github.com/scikit-rf/scikit-rf/blob/ac3aedfe0228615ef10da3e25212c7908059e6e0/skrf/calibration/tests/test_calibration.py#L389-L394

Here media `z0` is used for port and characteristic impedance. and the physical dimensions are only used to compute eventually gamma and losses. If we change the `line` behaviour, it will not change the `match`, `short`, `open`, etc. only `thru` or `attenuator` or `delay_something` would, which I think will increase confusion. `match` has even a `z_norm` booleean which is like `embed` for `line`.

Should we apply the table described in #651, we would have to specify `Z0=None` at `line`, `thru`, `attenuator` and `delay_something` which seems ugly.

mhuser:
This exploratory PR is closed to be replaced by a fresh new one based on latest development in #651 issue discussion.

PR #705, opened 2022-06-13
drop_non_monotonic_increasing fix for 2D z0 shape

### LINKED ISSUE #704: Drop non monotonic increasing does not work, if Network's z0 is accessed
As the access of network's `z0` silently writes the value `drop_non_monotonic_increasing` is not working any more.

So this is fine:
```python
import skrf as rf

netw = rf.Network('corrupt_file.s2p')
netw.drop_non_monotonic_increasing()
print(netw.z0.shape)
```
But this not:
```python
import skrf as rf

netw = rf.Network('corrupt_file.s2p')
print(netw.z0.shape)
netw.drop_non_monotonic_increasing()
print(netw.z0.shape)
```

I discovered this while writing #703, so I suggest to keep this issue open and to discuss in the discussion thread about the underlying problem of the `z0` storage.

Ttl:
fixes #704

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1980,16 +1980,16 @@ class Network(object):
         """
         npoints = self.frequency.npoints
         idx = self.frequency.drop_non_monotonic_increasing()
-        self.s = npy.delete(self.s, idx, axis=0)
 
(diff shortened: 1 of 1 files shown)


FranzForstmayr:
Thanks Henrik :)

PR #708, opened 2022-06-15
Add forced_z11x to IEEEP370_SE_NZC_2xThru and correct typo

mhuser:
`IEEEP370_SE_NZC_2xThru` use the time-domain impulse response of s21 to compute the x length of the 2x-Thru network in time samples. The impedance z11x is taken on midpoint x in a step response of Z11 with DC extrapolation. This works well in the case where the 2x-thru has an uniform impedance segment around the midpoint.

Most of the time, the 2x-Thru length is not an integer multiple of time-domain sample time. This can lead to a ± 1 sample error in midpoint determination. If the impedance is non-uniform at midpoint, it can change slightly with a shift of 1 point left or right. This PR enable advanced users to manually set the z11x impedance to a known value in this case. 

- [x] Fix indentation error causing ZL and ZR being re-affected at each loop instead at the end
- [x] Fix ZL being computed twice and ZR never
- [x] Add the ability to manually force the midpoint impedance z11x for IEEEP370_SE_NZC_2xThru

### DIFF
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -1422,7 +1422,7 @@ class IEEEP370_MM_NZC_2xThru(Deembedding):
             
         forced_z11x_dd:
             If a value is specified, manually force this value for the midpoint
-            impedance z11x for diffrential-mode.
+            impedance z11x for differential-mode.
             This is only usefull in the case where the midpoint impedance is
             non-uniform and the x length ± 1 sample error caused by the delay
             not being an integer multiple of sampling time make the z11x choice
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -973,7 +973,7 @@ class IEEEP370_SE_NZC_2xThru(Deembedding):
     """
     def __init__(self, dummy_2xthru, name=None,
                  z0 = 50, use_z_instead_ifft = False, verbose = False,
-                 *args, **kwargs):
+                 forced_z11x = None, *args, **kwargs):
         """
         IEEEP370_SE_NZC_2xThru De-embedding Initializer
 
@@ -996,6 +996,14 @@ class IEEEP370_SE_NZC_2xThru(Deembedding):
             not enough point in time domain to determine length and impedance
             of midpoint. Parameter `verbose` could be used for diagnostic in
             ifft mode (default: False)
+            
+        forced_z11x:
+            If a value is specified, manually force this value for the midpoint
+            impedance z11x.
+            This is only usefull in the case where the midpoint impedance is
+            non-uniform and the x length ± 1 sample error caused by the delay
+            not being an integer multiple of sampling time make the z11x choice
+            incorrect. (Default: None)
         
         verbose :
             view the process (default: False)
@@ -1012,6 +1020,7 @@ class IEEEP370_SE_NZC_2xThru(Deembedding):
         self.z0 = z0
         dummies = [self.s2xthru]
         self.use_z_instead_ifft = use_z_instead_ifft
+        self.forced_z11x = forced_z11x
         self.verbose = verbose
 
         Deembedding.__init__(self, dummies, name, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


skrf/calibration/deembedding.py
```diff
@@ -996,6 +996,14 @@ def __init__(self, dummy_2xthru, name=None,
             not enough point in time domain to determine length and impedance
             of midpoint. Parameter `verbose` could be used for diagnostic in
             ifft mode (default: False)
+            
+        forced_z11x:
```
denzchoe, line 1000: As per the definition elsewhere in the skrf; should we reclassify the terminology from `forced_z11x` to `forced_z0_line`? 
https://github.com/scikit-rf/scikit-rf/blob/0d4b1a5bab8df3c0b6b89f7d639d98b7e1064bca/skrf/calibration/calibration.py#L2611-L2613

It seems to me their usage both have similar concepts / meaning.  Besides, z11x holds a lot of ambiguity in the context of rf users. It is open to discussion, but I was thinking that this could be a suggestion.

mhuser: This make sense.

mhuser: done.

skrf/calibration/deembedding.py
```diff
@@ -996,6 +996,14 @@ def __init__(self, dummy_2xthru, name=None,
             not enough point in time domain to determine length and impedance
             of midpoint. Parameter `verbose` could be used for diagnostic in
             ifft mode (default: False)
+            
+        forced_z0_line:
+            If a value is specified, manually force this value for the midpoint
+            impedance z0_line. The fixtures are renormalized with this value.
+            This is only usefull in the case where the midpoint impedance is
```
Ttl, line 1003: typo: useful

skrf/calibration/deembedding.py
```diff
@@ -996,6 +996,14 @@ def __init__(self, dummy_2xthru, name=None,
             not enough point in time domain to determine length and impedance
             of midpoint. Parameter `verbose` could be used for diagnostic in
             ifft mode (default: False)
+            
+        forced_z0_line:
+            If a value is specified, manually force this value for the midpoint
+            impedance z0_line. The fixtures are renormalized with this value.
+            This is only usefull in the case where the midpoint impedance is
+            non-uniform and the x length ± 1 sample error caused by the delay
```
Ttl, line 1004: what is "x length"? Should this be 2xthru length? Also I don't know what is meant by non-uniform midpoint impedance.

I understood that the method first uses s21 impulse response to determine the thru length, then it uses s11 step response at the half length to get the impedance at the split plane. Since each point in the time response corresponds to reflections from some finite length of line the method requires that the thru length is few time points long so that only the line reflections are in the mid time point. I don't think there is any requirement for length to be a multiple of time resolution. If that is a requirement, then fft should be zero padded or the output interpolated to get the length more accurately.

For very short thrus the time domain impedance determination doesn't work since there is no time point that has only the line reflections. In that case impedance could be given manually, but since e001 and e002 requires time domain transform at the midpoint the split thru includes some amount of launch.

Ttl (changes requested):

skrf/calibration/deembedding.py
denzchoe: Hi @Ttl,


There's a history of discussion during my testing here with @mhuser prior to this pull request. 
Perhaps taking a look at this https://github.com/mhuser/scikit-rf/issues/2#issuecomment-1149356021 may help in setting the documentation wording better.

Ttl: Yeah, so it is meant for very short thrus where there aren't enough points in the time domain to determine the split plane impedance. Any choice of time point for the impedance determination includes also reflections from the launches. It should be noted that even if the impedance is given manually, some amount of reflections from opposite side launch is still included in the split thrus since there isn't enough time resolution to separate them. It does improve the result though and I'm not saying there is anything wrong with the feature, just that the documentation should be improved.

skrf/calibration/deembedding.py
mhuser: done.

skrf/calibration/deembedding.py
mhuser: I am rewriting the docstring in this sense.

Ttl (approved):

jhillairet (approved):

jhillairet:
Seems OK for merging.

PR #709, opened 2022-06-17
Documentation updates

jhillairet:
- add recommentation for Q-factor fitting
- update doc of Network tuto on creating Network from other network parameters


doc/source/examples/qfactor/Finding_Dk_Df_from_resonance_fitting.ipynb
```diff
@@ -212,6 +212,75 @@
     "fitted_ntwk_144_4GHz.plot_s_polar(ax=ax, label=\"Fitted Model ~ 4GHz\", lw=2, ls=':', color='C3')"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "9f653dfd",
+   "metadata": {},
+   "source": [
+    "If you want to get the best accuracy in the calculation of the Q-factors, it’s recommanded to repeat the fitting using a S-parameters data having the resonance peak at the centre frequency and the span reduced to $\\pm f_L/Q_L$ (cf. [1] page 21). This is justified in two ways:\n",
```
Ttl, line 220: typo: recommended.

If this refitting with lower bandwidth is better should it be part of the fitting routine?

Ttl (approved):

doc/source/examples/qfactor/Finding_Dk_Df_from_resonance_fitting.ipynb
jhillairet: > If this refitting with lower bandwidth is better should it be part of the fitting routine?

Well, indeed, but this is very case-dependent, and I doubt that an automatic routine would solve all the cases... To me, this is more a recommendation for users requiring a high accuracy. For all the others, I suppose that a Q factor of 74.2 or 73.9 is the same. Maybe @andrewg22 can comment?

andrewg22: The fitting routine incorporates weighting factors (MAT 58 eqn. 28). Therefore, for data that has a broad span, re-fitting to a subset of the data with reduced span around the peak will not usually bring much benefit. To get the best accuracy, what's needed is to re-trigger the VNA with reduced span, and repeat the fit so that there are more points near the peak (see Fig. 6, MAT 58 page 21). The advantage is greatest if the resonance has poor shape, or if there are other modes nearby. I agree that not all users will need this .. it really depends on the application. For my work on measuring the dielectric loss of very low-loss materials (MAT 58 ref. [69]) I needed to measure Q changes of 1 in 10^4, so it was necessary to follow this approach.

My suggestion is to put some pseudo code for automation of the measurement in the documentation somewhere:
   # Set sweep parameters so the resonance is on the VNA screen
   TriggerVNA()
   ReadTrace()  # Complex S-parameters
   FitQ()
   SetCentreFreq(FL) # Using the fitted data
   SetSpan(2*FL/QL)
   TriggerVNA()
   ReadTrace()
   FitQ()
   CalculateUnloadedQ()

jhillairet: OK, Added.

Issue #629, opened 2022-04-13
"Calibration With Three Receivers" Needs to Explain the Correct SOLT Calibration Procedure for 1.5-port VNAs


biergaizi:
In the example *[Calibration With Three Receivers](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Calibration%20With%20Three%20Receivers.html)*, the Two Port One Path calibration algorithm is demonstrated, allowing full error correction on a VNA that can only measures `a1`, `b1` and `b2`. In other words, only S11 and S21 can be measured each time. The other half, S22 and S12 are unknown. Thus, the DUT is required to be measured twice in the forward and reverse direction.

However, in the subsequent demonstration, the example still implicitly assumes that all four S-parameters are available. The document says,

> The measurements of the calibration standards and DUT’s were downloaded from the VNA by saving touchstone files of the raw s-parameter data to disk.
> ```
>    $ ls three_receiver_cal/data/
>    'attenuator (forward).s2p'       simulation.s2p
>    'attenuator (reverse).s2p'       thru.s2p
>     load.s2p                       'wr15 shim and swg (forward).s2p'
>    'quarter wave delay short.s2p'  'wr15 shim and swg (reverse).s2p'
>     short.s2p
> ```

Here, all the measurement data, including the calibration data for Open and Short, are all given in s2p data. For example, this is the content of `attenuator (forward).s2p`, which has all the four S-parameters available. 

    # GHz S RI R 50.0 
    !freq ReS11 ImS11 ReS21 ImS21 ReS12 ImS12 ReS22 ImS22
    60.0 -0.883511900902 -1.44533252716 -8.34876118461e-06 1.33288849611e-05 -0.705963575399 -0.684623175984 0.916409479195 -0.800569661745

And more confusingly, S11/S22 and S21/S12 are not equal, suggesting they really are separate measurements made on a real 2-port VNA instead of a 1.5-port VNA. 

Now, I was trying to apply the Two Port One Path calibration to my real measurement data from a 1.5 port VNA, the data has the form of:

    # GHz S RI R 50.0 
    !freq ReS11 ImS11 ReS21 ImS21 ReS12 ImS12 ReS22 ImS22
    60.0 re lm re lm 0 0 0 0

Since S12 and S22 are unknown, they're represented as `0` in the data files.

This is confusing and raises several questions. These need to be addressed in the tutorial.

[...]

PR #711, opened 2022-06-18
Three Receivers Calibration: clarify SOLT procedures and source port.

biergaizi:
This update clarifies the Correct SOLT Calibration Procedure for 1.5-port VNAs. First, I added a reference to the SOLT example. I also explained the difference between a standard SOLT calibration and a three-receiver calibration, including the fact that SOL at port 2 is not needed (unless there's need to calibrate isolation), and that the results are still saved as 2-port network, but S22 and S12 will contain no meaningful data.

This PR clarifies the question reported in Issue #629.

### DIFF
--- a/doc/source/examples/metrology/Calibration With Three Receivers.ipynb
+++ b/doc/source/examples/metrology/Calibration With Three Receivers.ipynb
@@ -77,6 +77,21 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### Measure the Calibration Standands and DUT\n",
+    "\n",
+    "First, one measures the S-parameters of the Open, Short, and Load standards on the VNA according to the standard SOLT calibration procedures. We assume the reader is familiar with this task in scikit-rf. If not, it's fully described in the [SOLT example](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/SOLT.html).\n",
+    "\n",
+    "When calibrating a three-receiver VNA, the process is almost identical. But note that there are a few minor differences. First, the Open, Short, and Load standards are only measured at port 1, not port 2. The VNA cannot measure in the reverse orientation, so there's no reason to perform OSL calibration at port 2 (unless one needs the optional isolation calibration). \n",
+    "\n",
+    "Nevetheless, scikit-rf still expects two-port networks as input, we therefore still save all results as two-port networks. In this case, only $S_{11}$ and $S_{21}$ are the actual measurements, $S_{12}$ and $S_{22}$ are unused, not meaningful, and can contain arbitrary data. One can use all-zero values as placeholders. Internally, scikit-rf will set $S_{22} = S_{11}$ and $S_{12} = S_{21}$ for the purpose of calculations.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


biergaizi:
Just pushed another commit on `source_port`

Pointing out the fact that TwoPortOnePath assumes port 1 is the active port and S11/S21 is the valid data, but this assumption can be changed to S12/S22 by setting "source_port=2".

Ttl (approved):

jhillairet (approved):

doc/source/examples/metrology/Calibration With Three Receivers.ipynb
```diff
@@ -70,7 +70,20 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Measure the Calibration Standands and DUT\n",
+    "\n",
+    "First, one measures the S-parameters of the Open, Short, and Load standards on the VNA according to the standard SOLT calibration procedures. We assume the reader is familiar with this task in scikit-rf. If not, it's fully described in the [SOLT example](./SOLT.ipynb).\n",
+    "\n",
+    "When calibrating a three-receiver VNA, the process is almost identical. But note that there are a few minor differences. First, the Open, Short, and Load standards are only measured at port 1, not port 2. The VNA cannot measure in the reverse orientation, so there's no reason to perform OSL calibration at port 2 (unless one needs the optional isolation calibration). Nevetheless, scikit-rf still expects two-port networks as input, we therefore still save all results as two-port networks - only $S_{11}$ and $S_{21}$ are the actual measurements - $S_{12}$ and $S_{22}$ contain no meaningful data, but it's okay.\n",
```
Vinc0110, line 84: To clarify even more: S12 and S22 will be overridden in the `TwoPortOnePath` constructor to make the incomplete 2-port compatible for the `TwelveTerm` routine (by making it symmetrical with S22=S11, S12=S21). https://github.com/scikit-rf/scikit-rf/blob/7ad561755afbb6c939baa4c68a0d49b96d4c5684/skrf/calibration/calibration.py#L1825-L1829

So, the user can put any data they want for S12 and S22, most people will probably chose S12=S22=0, or they can directly put S12=S21, S22=S11. It does not matter.

Vinc0110 (approved):
Great. I only have one minor suggestion.

doc/source/examples/metrology/Calibration With Three Receivers.ipynb
biergaizi: I just changed the wording to:

> we therefore still save all results as two-port networks. In this case, only $S_{11}$ and $S_{21}$ are the actual measurements, $S_{12}$ and $S_{22}$ are unused, not meaningful, and can contain arbitrary data. One can use all-zero values as placeholders. Internally, scikit-rf will set $S_{22} = S_{11}$ and $S_{12} = S_{21}$ for the purpose of calculations.

jhillairet:
@biergaizi is it OK for merging?

biergaizi:
Yes.

PR #713, opened 2022-06-18
Convert All Documentation Links to Relative Path Links

biergaizi:
This PR converts all documentation links to relative path links. This way they also work in local documentation. A few broken links were also found and fixed.

P.S: I also noticed that there is a direct reference to https://scikit-rf.readthedocs.io/ in contributing/index.rst as the project's documentation site, which was *not* mindlessly converted, it's preserved as-is as it should.

### DIFF
--- a/doc/source/contributing/index.rst
+++ b/doc/source/contributing/index.rst
@@ -35,7 +35,7 @@ Basically, you will:
 3. Then send a `pull request (PR) <https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests>`_ for your additions to be merged into the main scikit-rf repository. Your proposal will be reviewed or discussed and you may receive some comments which only aim to make your contribution as great as possible!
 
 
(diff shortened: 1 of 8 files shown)


jhillairet (approved):
nice and useful fixes!

Vinc0110 (approved):
Cool! Did you build the docs locally and checked if there are any warnings related to these changes? (There are all sorts of warnings, unfortunately, but maybe you can have a quick glance.)

biergaizi:
I did. After the documents are changed, I tried building the docs repeatedly while watching for any warning from the console. I also tried clicking all the links in my local browser. It took me quite some minutes to figure out the correct syntax for all the links.

Now the only warning seems to be (this is from the CI but it's consistent with what I've seen):

> /home/docs/checkouts/readthedocs.org/user_builds/scikit-rf/checkouts/713/doc/source/tutorials/NetworkSet.ipynb:9: WARNING: Explicit markup ends without a blank line; unexpected unindent.
> /home/docs/checkouts/readthedocs.org/user_builds/scikit-rf/checkouts/713/doc/source/tutorials/NetworkSet.ipynb:10: WARNING: Line block ends without a blank line.

I don't understand what's going on. But I believe this warning already existed before the link change and it's unrelated. I don't notice other warnings otherwise. Still, it's all manual changes. So it's possible that I've missed something.

PR #720, opened 2022-06-23
Simplify z0 handing

FranzForstmayr:
This PR tries to simplify `z0` handling and stores `z0` in a numpy array regardless the simplicity. 
If `z0` is assigned with an scalar, the value is broadcasted automatically to `s`' shape. This allows the getter to be much simpler than before.

### DIFF
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -405,9 +405,10 @@ class Media(ABC):
         result.frequency = self.frequency
         result.s = npy.zeros((self.frequency.npoints, nports, nports), dtype=complex)
         if z0 is None:
-            z0 = self.z0
+            z0 = npy.zeros(result.s.shape[:2])
+            z0[:] = self.z0[:, None]
         elif isinstance(z0, str):
-            z0 = parse_z0(z0)*self.z0
+            z0 = npy.ones(result.s.shape[:2]) * parse_z0(z0)
 
         if z0_norm:
             z0 = z0*self.z0
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -405,8 +405,6 @@ class Media(ABC):
         result.frequency = self.frequency
         result.s = npy.zeros((self.frequency.npoints, nports, nports), dtype=complex)
         if z0 is None:
-            # z0 = npy.zeros(result.s.shape[:2])
-            # z0[:] = self.z0[:, None]
             z0 = self.z0
         elif isinstance(z0, str):
             z0 = npy.ones(result.s.shape[:2]) * parse_z0(z0)
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -405,8 +405,9 @@ class Media(ABC):
         result.frequency = self.frequency
         result.s = npy.zeros((self.frequency.npoints, nports, nports), dtype=complex)
         if z0 is None:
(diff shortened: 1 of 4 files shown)


FranzForstmayr:
Open for feedback :)

skrf/media/media.py
```diff
@@ -405,9 +405,11 @@ def match(self, nports: int = 1, z0: Union[NumberLike, None] = None,
         result.frequency = self.frequency
         result.s = npy.zeros((self.frequency.npoints, nports, nports), dtype=complex)
         if z0 is None:
+            # z0 = npy.zeros(result.s.shape[:2])
+            # z0[:] = self.z0[:, None]
             z0 = self.z0
```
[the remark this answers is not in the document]
FranzForstmayr: I wanted to, but the commented code failed lot of tests, I'll dig into this issue later.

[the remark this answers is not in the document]
mhuser: I think it could be left as is (because it is not in `Network`) and will be solved when refactoring `Media`.
`media.z0` can be `None` and so far I have also noticed depending the way `z0` of media is defined, it can raise an exception in `Network.renormalize(media.z0)` because of this line:
https://github.com/scikit-rf/scikit-rf/blob/ba3b0a7f391838dabaa6c122db1a8e28067e9494/skrf/network.py#L3036
In some conditions (with list instead numpy arrays) the result is a bool and has no method `any()`.

I do not know of course if it is the issue you encountered, but @Ttl suggested the line could be written like below so I mention it just in case.
```Python
if npy.any(self.z0 != z_new):
```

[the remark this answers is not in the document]
mhuser: Great PR anyway!

skrf/network.py
```diff
@@ -1981,13 +1962,13 @@ def drop_non_monotonic_increasing(self) -> None:
...
 
         """
-        npoints = self.frequency.npoints
         idx = self.frequency.drop_non_monotonic_increasing()
 
         # z0 getter and setter depend on s.shape matching z0.shape.
         # Call z0 getter and setter only when s and z0 shapes match.
         z0_new = npy.delete(self.z0, idx, axis=0)
         self.s = npy.delete(self.s, idx, axis=0)
+        print("23", self.s.shape, z0_new.shape, self.z0.shape)
```
jhillairet, line 1971:
```suggestion

```

skrf/network.py
```diff
@@ -958,6 +961,9 @@ def s(self, s: npy.ndarray) -> None:
         self._s = fix_param_shape(s)
         self.__generate_secondary_properties()
         self.__generate_subnetworks()
+        
+        if self.z0.ndim == 0:
+            self.z0 = self.z0
```
jhillairet, line 966: this is strange. Why is it needed here?

jhillairet (approved):

skrf/network.py
FranzForstmayr: Consider the following example:

```python
net = rf.Network()
net.z0 = 50 # _z0 is a scalar
net.s = np.random.rand(101,2,2) # _z0 is set again from _z0
# But because _s is specifed, the _z0 shape will be broadcasted to (101,2)

```

jhillairet: OK...

jhillairet (approved):

Issue #652, opened 2022-05-02
Writing touchstone with complex port impedances writes incorrect reference impedance


Ttl:
`write_touchstone` writes absolute value of `z0` of the first port first frequency point as the reference impedance, but doesn't seem to normalize the written S-parameters if other ports or frequency points are different.

Example:
```python
import numpy as np
import skrf
from skrf.media import DefinedAEpTandZ0

freq = skrf.Frequency(100, 200, 3, unit='MHz')
line_l = 0.1

media = DefinedAEpTandZ0(freq, A=0.1, ep_r=1, Z0=[50-50j, 50-25j, 50-15j], z0=50)
line_z0port = media.line(line_l, 'm', z0=media.Z0)

line_z0port.write_touchstone('line.s2p')
line_z0port_s2p = skrf.Network('line.s2p')

print('Written z0', line_z0port.z0)
print('Read z0', line_z0port_s2p.z0)

assert (line_z0port.s == line_z0port_s2p.s).all()
```

```
Written z0 [[50.-50.j 50.-50.j]
 [50.-25.j 50.-25.j]
 [50.-15.j 50.-15.j]]
Read z0 [[70.71067812+0.j 70.71067812+0.j]
 [70.71067812+0.j 70.71067812+0.j]
 [70.71067812+0.j 70.71067812+0.j]]
```

There's a question what should happen in this case? For one it seems that officially touchstone format doesn't support complex reference (page 7 in specification: http://ibis.org/touchstone_ver2.0/touchstone_ver2_0.pdf) although skrf can read them. Should reference always be 50 unless specified otherwise? There is `write_z0` option, but that's not standard.

mhuser:
Intuitively, if z0 in touchstone is not the same as network and `write_z0 == False`, shouldn't the output be renormalized to the real-valued reference impedance of touchstone file, which can be the lowest frequency absolute magnitude of first port impedance?
Obviously it is not he case here.
In network.py line 2158:
```Python
            # [HZ/KHZ/MHZ/GHZ] [S/Y/Z/G/H] [MA/DB/RI] [R n]
            output.write('# {} S {} R {} \n'.format(self.frequency.unit, form, str(abs(self.z0[0, 0]))))`
```
But there seems to be nothing done if any element of z0 is different of abs(self.z0[0, 0]), which lead to a slightly different network.

jhillairet:
What should we do about  this issue?

Ttl:
fixed in #672

PR #725, opened 2022-06-29
write_touchstone: don't write z_ref in complex format if it's real, fix #724

### LINKED ISSUE #724: write_touchstone() writes impedance in complex format when it's real and breaks interoperability
Today I've noticed that S-parameters generated by scikit-rf v0.23.0 via `write_touchstone()` stopped working in Qucs. I inspected the files and found the port reference impedance is given as a complex number even when it's real.

    !Created with skrf (http://scikit-rf.org).
    # MHz S RI R (50+0j)

This is a regression introduced by #672 when adding complex impedance support (#652). Before this change, the reference impedance was always given as a real number. Since complex impedance is not officially supported by Touchstone v1.1 or v2. It means this change <del>is breaking compatibility with a large number of RF applications.</del> causes scikit-rf to output an invalid file and it's breaking compatibility with nearly all applications.

<del>Also, the complex number, when given, is written as `(a+bj)` - but all I know is that this format is the native format for a complex number in Python - it's not clear to me whether some other tools are also using this format (HFSS?), a pointer/citation is appreciated.</del>

biergaizi:
<del>This commit checks if the imaginary part of the impedance is zero, if so, It also checks if the real impedance is an integer.</del> If so, it writes the reference impedance as an integer or real number, instead of a complex number. This fixes the issue #724.

This ensures that, for the most common use-case of 50 ohms, the impedance line is given in the familiar form:
<del>
    # MHz S RI R 50
</del>

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2282,7 +2282,14 @@ class Network(object):
                 output.write('!Data is not renormalized\n')
                 output.write('# {} S {} R\n'.format(ntwk.frequency.unit, form))
             else:
-                output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref))
+                # Write "r_ref.real" instead of "r_ref", so we get a real number "a" instead
+                # of a complex number "(a+0j)", which is unsupported by the standard Touchstone
+                # format (non-HFSS). We already checked in the beginning that "r_ref" must be
+                # real in this case (write_z0 == False).
+                assert r_ref.imag == 0, "Complex reference impedance is encountered when " \
+                                        "generating a standard Touchstone (non-HFSS), this " \
+                                        "should never happen in scikit-rf."
+                output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref.real))
 
             # write ports
             try:


Vinc0110 (approved):

skrf/network.py
```diff
@@ -2282,7 +2282,15 @@ def get_buffer() -> io.StringIO:
...
-                output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref))
+                # Write real reference impedance "a" if possible, complex impedance "(a+0j)"
+                # is not supported by most tools. Also, prefer integers if possible.
+                if r_ref.imag == 0 and int(r_ref.real) == r_ref.real:
+                    output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, int(r_ref.real)))
+                elif r_ref.imag == 0:
+                    output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref.real))
+                else:
+                    # Write complex impedance "(a+bj)".
+                    output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref))
```
mhuser, line 2293: Should eventually complex case emit a warning? Can also be left as is.

mhuser (approved):
Nicely done!

skrf/network.py
biergaizi: @mhuser Is there any tool in existence (other than scikit-rf) capable of reading a complex impedance in this form? If not, I'll add a warning message here.

mhuser: @biergaizi 
If there is, I know no examples of such tool, but I have not used them all. It is probably non-standard enough to justify the warning. 😄

biergaizi: I think I found the problem. I missed the forest while looking at a tree. 

Complex impedance is not supported to begin with (unless `write_z0` is used, which generates a HFSS-style Touchstone file), and in the beginning of `write_touchstone()`, it actually actively checks whether the imaginary part of the impedance is zero.

    if r_ref is None and write_z0 == False:
        if not equal_z0:
            raise ValueError("Network has unequal port impedances but reference impedance for renormalization 'r_ref' is not specified.")
        if have_complex_ports:
            raise ValueError("Network port impedances are complex but reference impedance for renormalization 'r_ref' is not specified.")
        r_ref = ntwk.z0[0, 0]
    elif r_ref is not None:
        if not npy.isscalar(r_ref):
            raise ValueError('r_ref must be scalar')
        if r_ref.imag != 0:
            raise ValueError('r_ref must be real')
        ntwk.renormalize(r_ref)

In other words, `r_ref` must always be a real number (`assert r_ref.imag == 0`) when we execute:

    output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref))

The output format complex number is purely a string formatting bug, which should not happen. So, instead of a UserWarning, the complex codepath should never be executed (unless there's a bug in scikit-rf). What we need here is not a warning but as assertion failure.

biergaizi:
It has been made clear in my previous discussion that there was no complex impedance support intended in the standard (non-HFSS) Touchstone codepath to begin with. In the beginning of `write_touchstone()`, it actually actively checks whether the imaginary part of the impedance is zero.

Thus, the complex number output format is purely a result of the string formatting bug, because the data type is `Complex`, but in this case, the imaginary part should always be `0`.

I just force-pushed a new fix with the following description:

> In scikit-rf v0.23.0, a regression was introduced in write_touchstone(),
> the port reference impedance is given as a complex number even when it's
> real.
>     
>     !Created with skrf (http://scikit-rf.org).
>     # MHz S RI R (50+0j)
>     
> This is a regression introduced by #672 after complex impedance support
> (#652) was added. Before this change, the reference impedance was always
> given as a real number. Since complex impedance is not officially
> supported by Touchstone v1.1 or v2. It means this change is breaking
> compatibility with a large number of RF applications.
>     
> In fact, a complex impedance should never appear here. In the beginning
> of write_touchstone(), we already checked and ensured that the imaginary
> part of the impedance must be zero when writing a standard (non-HFSS)
> Touchstone file.
>     
> This commit fixes the problem by writing "r_ref.real" instead of "r_ref",
> so we get a real number "a" instead of a complex number "(a+0j)". An
> assertion statement was also added to ensure the imaginary part of the
> complex impedance must be zero - this scenario should never happen.

I also removed the attempt to output an integer, since in previous versions of scikit-rf also used floating-point number, and a reread of the Touchstone spec shows floating-point is clearly supported. 

Please review the new changes.

mhuser (approved):

jhillairet (approved):
OK. Should we plan a bug fix release ?

skrf/network.py
```diff
@@ -2282,7 +2282,14 @@ def get_buffer() -> io.StringIO:
                 output.write('!Data is not renormalized\n')
                 output.write('# {} S {} R\n'.format(ntwk.frequency.unit, form))
             else:
-                output.write('# {} S {} R {} \n'.format(ntwk.frequency.unit, form, r_ref))
+                # Write "r_ref.real" instead of "r_ref", so we get a real number "a" instead
+                # of a complex number "(a+0j)", which is unsupported by the standard Touchstone
+                # format (non-HFSS). We already checked in the beginning that "r_ref" must be
+                # real in this case (write_z0 == False).
+                assert r_ref.imag == 0, "Complex reference impedance is encountered when " \
```
Vinc0110, line 2289: `assert` raises an `AssertionError`, which breaks the program, correct?
I think a `RuntimeWarning` would be better, see this example:
https://github.com/scikit-rf/scikit-rf/blob/21b3409966676d28ac5dfc29552b0eff6c408750/skrf/network.py#L560-L562

Vinc0110, line 2289: `assert` raises an `AssertionError`, which breaks the program, correct?
I think a `RuntimeWarning` would be better, see this example:
https://github.com/scikit-rf/scikit-rf/blob/21b3409966676d28ac5dfc29552b0eff6c408750/skrf/network.py#L560-L562

Vinc0110 (changes requested):
I just noticed one point to be discussed before merging.

Vinc0110:
> OK. Should we plan a bug fix release ?

Yes, let's do it. The IEEEP370 fixes should be included as well.

skrf/network.py
Vinc0110: Oh, the `stacklevel=2` argument is missing in the code above, but that's a good idea as well:
https://docs.python.org/3/library/warnings.html#warnings.warn

biergaizi: Yes, it breaks the program, but I believe it's the correct thing to do in this case.

This assertion `r_ref.imag == 0` is guaranteed to be **always true under all conditions**, unless there's another internal bug inside scikit-rf.

## Justification

Complex impedance is not supported to begin with (unless write_z0 is used, which generates a HFSS-style Touchstone file), and in the beginning of write_touchstone(), it actually actively checks whether the imaginary part of the impedance is zero.

```
if r_ref is None and write_z0 == False:
    if not equal_z0:
        raise ValueError("Network has unequal port impedances but reference impedance for renormalization 'r_ref' is not specified.")
    if have_complex_ports:
        raise ValueError("Network port impedances are complex but reference impedance for renormalization 'r_ref' is not specified.")
    r_ref = ntwk.z0[0, 0]
elif r_ref is not None:
    if not npy.isscalar(r_ref):
        raise ValueError('r_ref must be scalar')
    if r_ref.imag != 0:
        raise ValueError('r_ref must be real')
    ntwk.renormalize(r_ref)
```

Thus, r_ref must always be a real number (assert r_ref.imag == 0) when we execute. The complex number output format is purely a result of the string formatting bug as I reported. No complex number support was intended here.

## Summary

In other words, if `r_ref.imag != 0`, scikit-rf already raises a `ValueError` before it hits the "impossible" AssertionError.

If you still think an assertion is inappropriate, please continue the discussion. I don't rule out the possibility that I have missed something, but as far as I know, I don't see any reason not to raise an `AssertionError`.

biergaizi:
> > OK. Should we plan a bug fix release ?
> 
> Yes, let's do it. The IEEEP370 fixes should be included as well.

I noticed that in the current master branch, some non-bugfix feature updates have already been merged, including #719 and #720. Is it appropriate to include them in a bugfix-release? How was the same problem handled previously in scikit-rf? A common practice in many projects is cherry-picking only the bug fixes to a separate stable branch.

skrf/network.py
mhuser: I was ready to write that in `skrf` coding style, we mostly reserve `assert` for test cases scripts (`assert` are removed if python is executed with optimization `-o`, but raised exceptions remains, so assert are like `__debug__` stuffs useful for quickly locating bugs that should not be runtime errors). 

But actually, it appears `assert` is used at several places in the code outside test cases: 

https://github.com/scikit-rf/scikit-rf/blob/21b3409966676d28ac5dfc29552b0eff6c408750/skrf/network.py#L1674

https://github.com/scikit-rf/scikit-rf/blob/7ad561755afbb6c939baa4c68a0d49b96d4c5684/skrf/vi/vna/hp8510c_sweep_plan.py#L59

https://github.com/scikit-rf/scikit-rf/blob/6928886fbb99744e1e77f7843045a2497cc67c99/skrf/qfactor.py#L698

https://github.com/scikit-rf/scikit-rf/blob/1e998a967e523a7cd48cf5755640a0e5b6fe2f18/skrf/mathFunctions.py#L1026

https://github.com/scikit-rf/scikit-rf/blob/f9d0c6f1c5b8d271181ca40e3e892f73a4b599d4/skrf/calibration/calibration.py#L3880

biergaizi: Also pinging @Ttl for comment.

biergaizi: > I was ready to write that in skrf coding style, we mostly reserve assert for test cases scripts (assert are removed if python is executed with optimization -o, but raised exceptions remains

Good point. If it's the case, would it be better to repeat the same `raise ValueError('r_ref must be real')` at the beginning of the file instead of using an `assert`?

mhuser: In this specific situation (statement that would never be reached during runtime), it would require the function to be modified to trigger the assertion. This would probably happens during development e.g. within a test case and not in a user script.
Both approach seems acceptable to me but I let the other share their mind.

Ttl (approved):

skrf/network.py
Vinc0110: Thanks for the detailed comments. I have indeed missed the other previous checks of the realness.
In this case where the warning/error should never get executed, I don't have a strong opinion on how to do it anymore. Throwing another `ValueError` seems reasonable.

Ttl: assert should be fine since it is used to check for assumptions that should always hold.

biergaizi: The current consensus seems to be that the use of `assert` is acceptable, and since its only purpose is alerting future developers, it doesn't need to be replaced by a `ValueError`, 

Any additional comment? Is it okay to be merged?

jhillairet:
> I noticed that in the current master branch, some non-bugfix feature updates have already been merged, including #719 and #720. Is it appropriate to include them in a bugfix-release? How was the same problem handled previously in scikit-rf? A common practice in many projects is cherry-picking only the bug fixes to a separate stable branch.

Well, it's right that in some projects there is a `dev`/`stable` branch in which all PR are merged before being integrated into `master`... Up to now this is not the case for scikit-rf, but could be considered if the community of contributors continues to grow. So for now, I don't think cherry picking only the bug fixes worth the effort, and I'll produce a small release when this issue will be fixed.

mhuser:
This PR received wide approbation, so I think it is ok to merge.

mhuser:
@biergaizi thank you a lot for detecting the touchstone issue and fixing it!

PR #728, opened 2022-06-29
Fix IEEEP370 deembedding methods for dataset with existing dc point or interpolation required

mhuser:
Fix issues with IEEEP370 deembedding methods when there is a DC point and add test cases for the issue not to come back again.

### DIFF
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -1134,7 +1134,6 @@ class IEEEP370_SE_NZC_2xThru(Deembedding):
                     RuntimeWarning
                     )
                 flag_DC = True
-                fold = f
                 f = f[1:]
                 s = s[1:]
             else:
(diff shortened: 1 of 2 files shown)


mhuser:
Bonus: implement interpolation when the 2x-Thru and FIX-DUT-FIX have not an harmonic sweep frequency vector (a vector were each frequency is a multiple of f[0]).

It is better to measure with the harmonic sweep, but in case were the frequency is close to that (e.g. quasi-integer multiple of f[0], interpolation can be convenient and it is proposed by the IEEEP370 matlab implementation.

mhuser:
I will need the help of a courageous developer to double-check this! 😄 
@jhillairet can we add this to the forecoming fixes release ?

mhuser:
@jhillairet shall we wait for the release of #733 before merging this (still on hold for approbation)?

jhillairet:
Sure, ok.

jhillairet:
I think we can merge

mhuser:
Thanks!

PR #732, opened 2022-07-07
fix `media.line` and `media.shunt_xxx` does not keep their network names

### LINKED ISSUE #731: inconsistent behavior when passing a network name as kwarg
Defining a name of a network sometimes works and sometimes it does not.

Two examples where passing a name works **correctly**:

Looking into the code: Media.inductor (similar: capacitor, resistor and many others) starts its definition using a match, to which the kwargs are passed. Then the match is modified to yield the result, so the name is correctly in the result.

Media.attenuator explicitly sets the name.

Examples where passing a name **does not work**:

Media.line defines a result element which includes the name (passed in kwargs). 
`result = self.match(nports=2, s_def='traveling', **kwargs)`

Then, the result element is cascaded with other elements: line uses a thru at both "ends" of the line. 
`result = self.thru(s_def='traveling')**result**self.thru(s_def='traveling')`

Network.cascade calls Network.connect, which uses a copy of the **first** element to build the result, but the thru does not carry the intended name, so the name is lost. 

Media.shunt_capacitor and shunt_inductor define a circuit element: a capacitor connected to a short, which includes the name. But note that kwargs are not passed to shunt() itself.
`return self.shunt(self.capacitor(C=C, **kwargs) ** self.short())`

Inside shunt(), the shorted capacitor is connected with an unnamed tee (because shunt() never receives any kwargs from shunt_capacitor() ). Again, the first element in connect() defines the name of the result, so the name is lost.
`return connect(self.tee(**kwargs), 1, ntwk, 0)`

Maybe it would help to include the kwargs when calling shunt(), then they will be passed to tee() and further to splitter() and match().
`return self.shunt(self.capacitor(C=C, **kwargs) ** self.short(), **kwargs)`

But I cannot oversee if that would have any unwanted side effects.

mhuser:
This PR attempt to fix #731 

While creating a `media.line` with `embed == True`, the name of the network, if passed into `**kwargs` was not kept due to the `media.thru()` cascading on both side.

This does not fix the faulty `embed` behaviour regarding the characteristic impedances (see #651 for reference to this issue).

`media.shunt_xxx` methods suffer from the same issue because the `**kwargs` is not passed to the `shunt` method. This PR fixes that.

### DIFF
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -847,7 +847,7 @@ class Media(ABC):
         result.s = \
                 npy.array([[s11, s21],[s21,s11]]).transpose().reshape(-1,2,2)
 
-        if embed:
+        if embed and self.z0 is not None:
             # warns of future deprecation
             warnings.warn('In a future version,`embed` will be deprecated.\n'
                           'The line and media port impedance z0 and '
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -838,6 +838,8 @@ class Media(ABC):
         s_def = kwargs.pop('s_def', S_DEF_DEFAULT)
         # Need to use either traveling or pseudo definition here
         # for the network to match traveling waves.
+        # the definition of the line trough a match and a delay would not
+        # works for complex characteristic impedances otherwise.
         result = self.match(nports=2, s_def='traveling', **kwargs)
 
         theta = self.electrical_length(self.to_meters(d=d, unit=unit))
@@ -854,8 +856,6 @@ class Media(ABC):
                           'characteristic impedance Z0 will be used instead '
                           'to determine if the line has to be renormalized.',
               FutureWarning, stacklevel = 2)
-            # Use the same s_def here as the line to avoid changing it during
-            # cascade.
             result.renormalize(self.z0, s_def=s_def)
         else:
             result.renormalize(result.z0, s_def=s_def)
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -836,10 +836,12 @@ class Media(ABC):
 
         kwargs.update({'z0':z0})
         s_def = kwargs.pop('s_def', S_DEF_DEFAULT)
-        # Need to use either traveling or pseudo definition here
-        # for the network to match traveling waves.
-        # the definition of the line trough a match and a delay would not
-        # works for complex characteristic impedances otherwise.
+        # The use of either traveling or pseudo waves s-parameters definition
+        # is required here.
+        # A line with a complex characteristic impedance equal to port
+        # impedance does not have zero reflection coefficients with the power
+        # waves definition.
+        # Instead, reflection coefficients will have conjugation. 
         result = self.match(nports=2, s_def='traveling', **kwargs)
 
         theta = self.electrical_length(self.to_meters(d=d, unit=unit))
@@ -903,7 +905,8 @@ class Media(ABC):
         delay_short
         delay_open
         """
-        return self.line(d=d, unit=unit, **kwargs) ** self.load(Gamma0=Gamma0, **kwargs)
+        return self.line(d=d, unit=unit, **kwargs) ** self.load(Gamma0=Gamma0,
+                                                                **kwargs)
 
     def delay_short(self, d: Number, unit: str = 'deg', **kwargs) -> Network:
         r"""
@@ -1115,8 +1118,8 @@ class Media(ABC):
         shunt_delay_short
         shunt_inductor
         """
-        return self.shunt(self.capacitor(C=C, **kwargs) ** self.short(),
-                          **kwargs)
+        return self.shunt(self.capacitor(C=C, **kwargs) ** 
+                          self.short(**kwargs), **kwargs)
 
     def shunt_inductor(self, L: NumberLike, **kwargs) -> Network:
         r"""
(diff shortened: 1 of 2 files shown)


jhillairet (approved):

skrf/media/media.py
```diff
@@ -838,6 +838,8 @@ def line(self, d: NumberLike, unit: str = 'deg',
         s_def = kwargs.pop('s_def', S_DEF_DEFAULT)
         # Need to use either traveling or pseudo definition here
         # for the network to match traveling waves.
+        # the definition of the line trough a match and a delay would not
+        # works for complex characteristic impedances otherwise.
```
Ttl, line 842: typo: works.

A better explanation would be that power wave reflection coefficient has conjugation. A line with characteristic impedance equal to the port impedance does not have zero reflection coefficient even with a zero length line if port impedance is complex.

skrf/media/media.py
```diff
@@ -1114,7 +1115,8 @@ def shunt_capacitor(self, C: NumberLike, **kwargs) -> Network:
         shunt_delay_short
         shunt_inductor
         """
-        return self.shunt(self.capacitor(C=C, **kwargs) ** self.short())
+        return self.shunt(self.capacitor(C=C, **kwargs) ** self.short(),
```
Ttl, line 1118: I think `short` call also needs `**kwargs` or at least `s_def` to avoid warning of connecting different `s_def` networks when giving `s_def`.

skrf/media/media.py
```diff
@@ -1146,7 +1148,8 @@ def shunt_inductor(self, L: NumberLike, **kwargs) -> Network:
         shunt_delay_short
         shunt_capacitor
         """
-        return self.shunt(self.inductor(L=L, **kwargs) ** self.short())
+        return self.shunt(self.inductor(L=L, **kwargs) ** self.short(),
```
Ttl, line 1151: Same as above.

Ttl (changes requested):
Can you add tests?

skrf/media/media.py
mhuser: > I think `short` call also needs `**kwargs` or at least `s_def` to avoid warning of connecting different `s_def` networks when giving `s_def`.

Indeed. I will do that.

skrf/media/media.py
mhuser: I will rephrase in this sense.
While this is outside the scope of this PR: isn't the `media.match` method with `npy.zeros()` wrong for the power waves ?

mhuser: Done

skrf/media/media.py
mhuser: Done

skrf/media/media.py
mhuser: Done

mhuser:
> Can you add tests?

Yes, I have added test cases for the naming of networks generated by the various `media` methods right now.

jhillairet (approved):
Great work!

skrf/media/media.py
```diff
@@ -836,10 +836,12 @@ def line(self, d: NumberLike, unit: str = 'deg',
...
-        # Need to use either traveling or pseudo definition here
-        # for the network to match traveling waves.
-        # the definition of the line trough a match and a delay would not
-        # works for complex characteristic impedances otherwise.
+        # The use of either traveling or pseudo waves s-parameters definition
+        # is required here.
+        # A line with a complex characteristic impedance equal to port
+        # impedance does not have zero reflection coefficients with the power
+        # waves definition.
+        # Instead, reflection coefficients will have conjugation.
```
Ttl, line 844: This sounds misleading. The definition of the reflection coefficient for powerwaves has conjugation.

skrf/media/tests/test_media.py
```diff
@@ -24,46 +25,210 @@ def setUp(self):
...
+        Test the naming of the network. When circuit is used to connect a
+        topology of networks, they should have unique names.
+        """
+        name = 'attenuator,-10dB'
+        self.dummy_media.frequency = Frequency(1, 1, 1, 'GHz')
+        skrf_ntwk = self.dummy_media.attenuator(-10, d = 90, unit = 'deg',
+                                                name = name)
+        self.assertEqual(name, skrf_ntwk.name)
+        
+    def test_lossless_mismatchh(self):
```
Ttl, line 213: typo

Ttl (approved):

skrf/media/media.py
mhuser: Just to be sure of my understanding: if we have a line of characteristic impedance 50 + 1j ohm, we should have port impedances of 50 - 1j ohm to have S11 equal to 0, is this correct?

mhuser: Maybe shall we just replace the 4 last comment line by ”The definition of the reflection coefficient for power waves has conjugation." 😅

skrf/media/tests/test_media.py
mhuser: Done.

mhuser:
@jhillairet shall we wait for the release of #733 before merging this?

jhillairet:
> @jhillairet shall we wait for the release of #733 before merging this?

No, I think we can merge it to include it in the next release.

jhillairet (approved):

mhuser:
Done !

Issue #734, opened 2022-07-18
time_gate with multiplication in time domain


j-bau:
For time gate skrf currently only supports convolution in frequency domain. In some cases it is more useful to multiply the gate in time domain.

The modification would be placed in time.py within the function time_gate and could be accessed by a new variable to switch between fd and td.

Code as optional path starting from line 300 .. 312 in scikit-rf/skrf/time.py
```python
out = ntwk.copy()
s_td = np.fft.ifft(out.s[:,0,0])
s_td_g = s_td * np.fft.ifftshift(gate)
s_fd_g = np.fft.fft(s_td_g)
out.s[:,0,0] = s_fd_g
```

Vinc0110:
Maybe the `mode` parameter could be used to set the desired behavior. For example `mode='time'` to directly multiply the window in the time domain.

jhillairet:
For clarification, the code under discussion is:
https://github.com/scikit-rf/scikit-rf/blob/7f4eee82c19ce3fe2e5f79472a904bad15e86bcb/skrf/time.py#L302-L312

I assume the [convolve1d ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.convolve1d.html) function from Numpy is already optimized. @j-bau , did you run into cases where multiplication in time domain was faster than the convolution in frequency domain?

j-bau:
Multiplication in time domain is indeed about 25x faster. I tested it versus my code proposal.

- _FD=True is for convolution in frequency domain
- _FD=False is for multiplication in time domain
- Based on Touchstone file with 5001 points

```
In [1]: %timeit -n 50 -r 7 time_gate(ntw.s21, window=('kaiser', 6), center=4, span=2, boundary='wrap', _FD=True)
71.1 ms ± 160 µs per loop (mean ± std. dev. of 7 runs, 50 loops each)

In [2]: %timeit -n 50 -r 7 time_gate(ntw.s21, window=('kaiser', 6), center=4, span=2, boundary='wrap', _FD=False)
2.74 ms ± 33.1 µs per loop (mean ± std. dev. of 7 runs, 50 loops each)
```
But this wasn't my main point to rise here. In some cases convolution leads to other results than multiplication. In my case the time domain approach delivered a more clear signal to continue working with.

Vinc0110:
> In some cases convolution leads to other results than multiplication. In my case the time domain approach delivered a more clear signal to continue working with.

[...]

PR #735, opened 2022-07-22
Fixes and improves time-domain gating

Vinc0110:
Fix for #734.
As disussed in the issue thread, the new implementation uses time-domain multiplication instead of frequency-domain convolution. This gives a speed benefit of multiple orders of magnitude, depending on the number of samples.

### DIFF
--- a/skrf/time.py
+++ b/skrf/time.py
@@ -287,9 +287,9 @@ def time_gate(ntwk: 'Network', start: float = None, stop: float = None, center: 
     gate[start_idx:stop_idx] = window
 
     # time-domain gating
-    s_td = ifftshift(ifft(nw_gated.s[:, 0, 0]))
+    s_td = fftshift(ifft(nw_gated.s[:, 0, 0]))
     s_td_g = s_td * gate
-    nw_gated.s[:, 0, 0] = fft(fftshift(s_td_g))
+    nw_gated.s[:, 0, 0] = fft(ifftshift(s_td_g))
 
     if mode == 'bandstop':
         nw_gated = ntwk - nw_gated
--- a/skrf/time.py
+++ b/skrf/time.py
@@ -223,12 +223,12 @@ def time_gate(ntwk: 'Network', start: float = None, stop: float = None, center: 
     span : number, or None
         span of time gate, (ns).  If None span will be half of the
         distance to the second tallest peak
-    mode: ['bandpass','bandstop']
+    mode : ['bandpass', 'bandstop']
         mode of gate
-    boundary:  {'reflect', 'constant', 'nearest', 'mirror', 'wrap'},
-        passed to `scipy.ndimage.filters.convolve1d`
     window : string, float, or tuple
         passed to `window` arg of `scipy.signal.get_window()`
+    return_all : bool
+        whether or not the response of the normalized time gate is returned together with the gated Network as a Tuple.
 
     Note
     ----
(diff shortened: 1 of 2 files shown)


skrf/time.py
```diff
@@ -287,47 +279,29 @@ def time_gate(ntwk, start: float = None, stop: float = None, center: float = Non
...
-    # conditionally  un-delay ntwk
-    if center!=0:
-        out = out.delay(center*1e9, 'ns',port=0,media=media)
+    gate = npy.zeros_like(t)
+    gate[start_idx:stop_idx] = window
+
+    # time-domain gating
+    s_td = npy.fft.ifftshift(npy.fft.ifft(nw_gated.s[:, 0, 0]))
+    s_td_g = s_td * gate
+    nw_gated.s[:, 0, 0] = npy.fft.fft(npy.fft.fftshift(s_td_g))
```
mhuser, line 291: Suggestion: addition of `from numpy.fft import fft, ifft, fftshift, ifftshift` on top of the file to improve readability.
```Python
    s_td = ifftshift(ifft(nw_gated.s[:, 0, 0]))
    s_td_g = s_td * gate
    nw_gated.s[:, 0, 0] = fft(fftshift(s_td_g))
```

mhuser (approved):
Great work!
A small suggestion to improve readability, see comment.

skrf/time.py
Vinc0110: Thanks! Done.

skrf/time.py
```diff
@@ -287,47 +280,29 @@ def time_gate(ntwk, start: float = None, stop: float = None, center: float = Non
...
-        gate_ntwk.s = kernel
-        gate_ntwk= gate_ntwk.delay(center*1e9, 'ns', media=media)
-
-        return {'gated_ntwk':out,
-                'gate':gate_ntwk}
+        # compute and return frequency response of time gate
+        nw_gate = nw_gated.copy()
+        nw_gate.s = ifftshift(fft(fftshift(gate)))
+        nw_gate.name = 'gate'
+        return nw_gated, nw_gate
```
Vinc0110, line 306: I just remembered I changed the output for `return_all=True` because I did not like the `Dict` format. Should I change it back? 
I'm wondering if this part could be removed completely. The `return_all` parameter has not even been documented previously, so not sure if anyone cares... Is there a use case where you would need the frequency-domain response of the time gate delivered in a 1-port Network?

```suggestion
        return {'gated_ntwk': nw_gated, 'gate': nw_gate}
```

jhillairet: I would say to drop it as I do not like functions that does not return a fixed number of arguments (like in matlab), however I have no idea if this could be useful or not...

jhillairet (approved):

skrf/time.py
Vinc0110: I removed the parameter in 800966f, but now the builds of read-the-docs fail for some reason. The error output is very messy, so I don't know what's going on...

Vinc0110: @jhillairet Do you know how to debug this? I tried rerunning the checks, but it did not to rerun the build for read-the-docs. Building the docs locally works without a problem...

FranzForstmayr:
Hi @Vinc0110 

despite the fact I have a build issue on my local machine at the moment too, I just found out that the sphinx version rtd is using is quite old.
Actually we are using 1.8.4
https://docs.readthedocs.io/en/stable/build-default-versions.html#external-dependencies

Vinc0110:
Thanks for fixing it!

PR #738, opened 2022-08-01
Checking if port impedances are real

qmfrf:
When working with large matrices (>300 ports) the matrix multiplication used in s2s can very slow. However, if the port terminations z0 are completely real then the travelling and power wave definitions are the same, so no conversion is necessary.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -5467,12 +5467,13 @@ def s2s(s: NumberLike, z0: NumberLike, s_def_new: str, s_def_old: str):
         # Nothing to do.
         return s
         
-    if npy.isreal(z0).all():
-         # Nothing to do because all port impedances are real so the used definition (power or travelling) does not make a difference.
-        return s
-
     nfreqs, nports, nports = s.shape
     z0 = fix_z0_shape(z0, nfreqs, nports)
+
+    if npy.isreal(z0).all():
+        # Nothing to do because all port impedances are real so the used 
+        # definition (power or travelling) does not make a difference.
+        return s
 
     # Calculate port voltages and currents using the old s_def.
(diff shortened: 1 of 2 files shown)


FranzForstmayr:
LGTM

jhillairet (approved):
Thank you @qmfrf and welcome as a contributor to scikit-rf. It's OK to me.

skrf/network.py
```diff
@@ -5466,6 +5466,11 @@ def s2s(s: NumberLike, z0: NumberLike, s_def_new: str, s_def_old: str):
     if s_def_new == s_def_old:
         # Nothing to do.
         return s
+        
+    if npy.isreal(z0).all():
```
Ttl, line 5470: I don't think this works with scalar `z0`. Either `npy.all(npy.isreal(z0))` or move the check after `fix_z0_shape`.

Ttl (changes requested):

skrf/network.py
FranzForstmayr: `z0` should be a numpy array in any case. `z0` is set at the Network constructor.

FranzForstmayr: Sorry, my fault

jhillairet: Indeed, using `ntwk.z0` as the input of the function would work, but using directly a constant like `50` would fail.

 @qmfrf, as suggested by @Ttl, could you solve this issue by moving your code just below the lines:

```python
    nfreqs, nports, nports = s.shape
    z0 = fix_z0_shape(z0, nfreqs, nports)
```

And, eventually, adding a new test case for passing a scalar value in the `s2s` method, for example in `test_s_def_setters` in `test_network.py`.

jhillairet:
Also solve issue #764

PR #739, opened 2022-08-13
Fix OpenShort and ShortOpen de-embedding procedure

keikawa:
The current implementation of OpenShort and ShortOpen de-embedding differs slightly from the reference [1] presented in the API documentation. This pull request fixes the procedure of OpenShort and ShortOpen de-embedding to be the same as in reference [1]. Specifically, parasitic components in the Open and Short dummies are first removed prior to removing parallel or series parasitics from the DUT. This fix gives more accurate results than the previous implementation.

[1] M. C. A. M. Koolen, J. A. M. Geelen, and M. P. J. G. Versleijen, “An improved de-embedding technique for on-wafer high-frequency characterization,” in Proceedings of the 1991 Bipolar Circuits and Technology Meeting, Minneapolis, MN, USA, 1991, pp. 188–191. doi: 10.1109/BIPOL.1991.160985.


Vinc0110:
This sounds good, but I'm not an expert in this field. You say this fix provides more accurate results than the current implementation. Do you maybe have an example to show it, maybe even with real measurements instead of the ADS simulations? Reading the [tutorial](https://scikit-rf.readthedocs.io/en/latest/tutorials/Deembedding.html), it seems the 1 nH inductance is being deembedded without a problem.

What about the existing unit tests? They seem to be fine. Are the tests not realistic enough to cover real-world cases?

Don't take this the wrong way, I'm just wondering where the problem is and if this fix is really required.

keikawa:
Evaluating de-embedding methods using measurement results is not easy because the "true value" of the DUT can not known. Here, we test whether the parasitic components of the assumed equivalent circuit can be accurately removed. Implementation before the fix is not exact because it over-removes parallel or series parasitics. The de-embedding results for the 1-nH inductor shown in the tutorial are more accurate after the fix as follows:

Before fix: Lactual_nH[-1] = 1.000360021406259
After fix: Lactual_nH[-1] = 1.000000032913312

I changed the tolerance of test to be stricter from 1e-3 to 1e-6. The procedure before the fix does not pass the more stringent test, but the after fix does.
Another important reason to make the fix is that a difference between the reference and the code may confuse users. The code after the fix is exactly the same as equation (3) in reference [1].
Thank you ( ´ ▽ ` )ﾉ

Vinc0110 (approved):
Ok, that all sounds very reasonable.
I first thought that 1.00036 nH is close enough to 1 nH, but maybe we can to have higher expectations for such ideal calculations. Tightening the relative tolerance to 1e-6 is probably a good idea.
Thank you!

jhillairet (approved):
Great PR, thank you @keikawa

PR #740, opened 2022-08-15
Python upgrade

FranzForstmayr:
Let's see if the github runner is able to use the new python3.11 prerelease.


FranzForstmayr:
Ok, Github Actions is not working yet, it seems there are issues with scipy, as there is no installable wheel yet.
However, I don't expect issues for the upcoming python release, as all the tests work on my local machine.

Let's wait some time, maybe we get a new scipy release soon.

PR #746, opened 2022-08-31
Update general.py - introducing sort to read_all networks function

xafra:
The reading of files is not traceable or inconsistent, especially for linux machines. 
Filenames were chosen arbitrary and a simple filelist.sort() could fix it.

I already submitted somehow same request in 2020.

I would even recommend to set sorted even as standard. I do not know why this request has been removed again afterwards, probably just by error or there is a specific reason. Probably there's a reason for the backchange I'm not aware of or missed when skimming the code for the first time.

I did not set it the sorted as standard so far because of above mentioned reasons. 

Please consider this change, as I am really struggling with it. 
Thanks in advance.

### DIFF
--- a/skrf/io/general.py
+++ b/skrf/io/general.py
@@ -249,7 +249,7 @@ def write(file, obj, overwrite = True):
         pickle.dump(obj, fid, protocol=2)
         fid.close()
 
-def read_all(dir: str ='.', sorted = None, contains = None, f_unit = None, obj_type=None, files: list=None, recursive=False) -> dict:
+def read_all(dir: str ='.', sort = None, contains = None, f_unit = None, obj_type=None, files: list=None, recursive=False) -> dict:
     """
     Read all skrf objects in a directory.
 
@@ -319,8 +319,8 @@ def read_all(dir: str ='.', sorted = None, contains = None, f_unit = None, obj_t
     else:
         filelist.extend(files)
          
-    if sorted is not None:
-      filelist = sorted(filelist)
+    if sort is not None:
+      filelist.sort()
 
     for filename in filelist:
         if contains is not None and contains not in filename:
@@ -414,6 +414,7 @@ def write_all(dict_objs, dir='.', *args, **kwargs):
     """
     if not os.path.exists('.'):
         raise OSError('No such directory: %s'%dir)
+         
 
 
     for k in dict_objs:
@@ -498,7 +499,7 @@ def load_all_touchstones(dir = '.', contains=None, f_unit=None):
 
     Notes
     -------
-    Alternatively you can use the :func:`read_all` function.
+    Alternatively you can use the :func:`#` function.
 
     Parameters
     -----------
--- a/skrf/io/general.py
+++ b/skrf/io/general.py
@@ -249,7 +249,7 @@ def write(file, obj, overwrite = True):
         pickle.dump(obj, fid, protocol=2)
         fid.close()
 
-def read_all(dir: str ='.', sort = None, contains = None, f_unit = None, obj_type=None, files: list=None, recursive=False) -> dict:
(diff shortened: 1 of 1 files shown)


skrf/io/general.py
```diff
@@ -495,7 +499,7 @@ def load_all_touchstones(dir = '.', contains=None, f_unit=None):
 
     Notes
     -------
-    Alternatively you can use the :func:`read_all` function.
+    Alternatively you can use the :func:`#` function.
```
jhillairet, line 502: What is the `#` doing?

jhillairet (changes requested):
Thanks. I probably missed the previous PR... 

Please find a few comments.

xafra:
Thanks a lot for your highly appreciated comments and  outstanding response time. 

There is no need to apologize for not implementing PR in 2020, as it was done. 
The actual code changed a lot and I think it was just overlooked to implement it again.

To your comments:
- why not a boolean for sort instead? Are you thinking about different sorting options?
Yes, my plan was to keep it open for different sorting options. 

- Please also add the sort parameter in the docstring.
Yes, I will do this. 

- What is the # doing?
Sorry this is then probably a mistake. I will remove it. I do not know, why it exists.

xafra:
I hope I was able to implement the changes as expected. 
Struggling with github. Sorry for any inconveniences caused

skrf/io/general.py
```diff
@@ -249,7 +249,7 @@ def write(file, obj, overwrite = True):
         pickle.dump(obj, fid, protocol=2)
         fid.close()
 
-def read_all(dir: str ='.', contains = None, f_unit = None, obj_type=None, files: list=None, recursive=False) -> dict:
+def read_all(dir: str ='.', sort = True, contains = None, f_unit = None, obj_type=None, files: list=None, recursive=False) -> dict:
     """
```
jhillairet, line 253: to be decided of the default value is True.

By the way, letting sort=True does not allow different sorting method, as you proposed initially. So, to be decided as well if further sorting method should be employed.

Any advice ?

xafra: Thank you for your quick response.

Default: True
Of course it need to be decided, if the default value is True. No offense. Don't want to be rude.
Actually I thought a bit about it and could not find a drawback. 
For Linux users the filename import into filelist can be illogical, while for Windows users I do not expect a change.

Boolean: 
I picked up your initial idea with the boolean parameter,
I do not know other possible useful sorting methods for the time being which require an instant representation in skrf. 
I thought like this: 
A start has been made with introducing the idea of applying a sort algorithm.
If someone sees a need for a special sort implementation, they hopefully will implement or make a proposal to apply changes.

xafra:
I assume you have not been notified on the direct answer, so I want to make you aware, of my most recent comments. 
It would be nice, if you can consider pull request.

jhillairet:
> I assume you have not been notified on the direct answer, so I want to make you aware, of my most recent comments. It would be nice, if you can consider pull request.

Indeed, or I missed it. Sorry. Thanks for the ping.

skrf/io/general.py
jhillairet: > Default: True Of course it need to be decided, if the default value is True. No offense. Don't want to be rude. Actually I thought a bit about it and could not find a drawback. For Linux users the filename import into filelist can be illogical, while for Windows users I do not expect a change.

It is important to not brake previous code, so I'd prefer the `sort` parameter to be off by default. Thanks. 
 
> Boolean: I picked up your initial idea with the boolean parameter, I do not know other possible useful sorting methods for the time being which require an instant representation in skrf. I thought like this: A start has been made with introducing the idea of applying a sort algorithm. If someone sees a need for a special sort implementation, they hopefully will implement or make a proposal to apply changes.

Yes, eventually. So let's keep the boolean parameter, and if later someone need to sort the file by time, we will advise.

jhillairet (approved):

jhillairet:
Thanks @xafra for the PR.

PR #760, opened 2022-09-29
Make vector fitting Spice output compatible to LTSpice

FranzForstmayr:
LTSpice does not have the `mult` argument for the voltage controlled voltage source:
https://ltwiki.org/LTspiceHelp/LTspiceHelp/G_Voltage_Dependent_Current_Source.htm

This PR multiplies the `mult` factor with the gain directly.

### DIFF
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -1980,7 +1980,7 @@ class VectorFitting:
             f.write('L1 n_pos 1 {ind}\n')
             f.write('C1 1 2 {cap}\n')
             f.write('R1 2 n_neg {res}\n')
-            f.write('G1 n_pos n_neg 1 2 {gm} m={mult}\n')
+            f.write('G1 n_pos n_neg 1 2 {gm * mult}\n')
             f.write('.ENDS rcl_vccs_admittance\n')
 
             f.write('*\n')
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -1875,7 +1875,7 @@ class VectorFitting:
             for n in range(self.network.nports):
                 str_input_nodes += 'p{} '.format(n + 1)
 
-            f.write('.SUBCKT {} {}\n'.format(fitted_model_name, str_input_nodes))
+            f.write(f'.SUBCKT {fitted_model_name} {str_input_nodes}\n')
 
             for n in range(self.network.nports):
                 f.write('*\n')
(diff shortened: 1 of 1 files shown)


skrf/vectorFitting.py
```diff
@@ -1873,7 +1875,7 @@ def get_new_subckt_identifier():
             for n in range(self.network.nports):
                 str_input_nodes += 'p{} '.format(n + 1)
 
-            f.write('.SUBCKT s_equivalent {}\n'.format(str_input_nodes))
+            f.write('.SUBCKT {} {}\n'.format(fitted_model_name, str_input_nodes))
```
jhillairet, line 1878:
```suggestion
            f.write(f'.SUBCKT {fitted_model_name} {str_input_nodes}\n')
```
should also work (?)

jhillairet (approved):

skrf/vectorFitting.py
```diff
@@ -1980,7 +1982,7 @@ def get_new_subckt_identifier():
             f.write('L1 n_pos 1 {ind}\n')
             f.write('C1 1 2 {cap}\n')
             f.write('R1 2 n_neg {res}\n')
-            f.write('G1 n_pos n_neg 1 2 {gm} m={mult}\n')
+            f.write('G1 n_pos n_neg 1 2 {gm * mult}\n')
```
Vinc0110, line 1985: Good idea!
It seems the `mult` parameter only exists in ngspice. But does it always work to have a negative transductance (in other versions than LTspice)? Based on the docs, I'm not entirely sure. Did you test this? Otherwise, we can also flip the controlling nodes in case of `multi = -1`.

Berkeley SPICE: http://bwrcs.eecs.berkeley.edu/Classes/IcBook/SPICE/UserGuide/elements.html#462
ngspice (chapter 4.2.1): https://ngspice.sourceforge.io/docs/ngspice-37-manual.pdf

skrf/vectorFitting.py
FranzForstmayr: Yes, I always forget that f-strings work with python3.6 already

skrf/vectorFitting.py
FranzForstmayr: I tested with Qucs (ngspice backend) and this worked. However this was my first try with qucs, maybe you could double check. :)

jhillairet (approved):

skrf/vectorFitting.py
Vinc0110: Ok. Sorry for the delay. I'm running some tests now.

Vinc0110: Works well in both LTSpice and ngspice

PR #761, opened 2022-09-29
Improve efficiency of network_2_dataframe

stevehamblin:
Simple change to eliminate redundant calculations of the desired format ('s_db', 's_rad', etc.) when executing network_2_dataframe.

This is not terribly noteworthy for one-time execution, but provides some improvement when batch processing hundreds of networks.


jhillairet (approved):
Indeed, thanks.

Vinc0110 (approved):

PR #768, opened 2022-10-09
Vector Fitting: bugfix for data including dc samples

Vinc0110:
Vector Fitting crashes if the fitted data includes a dc sample.

The first starting pole $p_0$ is always positioned at $\omega_0=2\pi f[0]$, which is $0$ in such cases. The respective coefficient then becomes `inf` or `NaN` and causes the crash:
$\frac{1}{j \pi f_0 - p_0} = \frac{1}{0-0}$.

This fix places $p_0$ at $\omega_0=2\pi f[1] / 1000$, so the coefficient is less problematic:
$\frac{1}{j \pi f_0 - p_0} = \frac{1}{0-\omega_1 / 1000}$.

The factor 1/1000 is chosen arbitrarily and works well in my tests. There might be a better way to do it, but let's go with this for now.

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -46,7 +46,7 @@ class VectorFittingTestCase(unittest.TestCase):
 
     def test_dc(self):
         # perform the fit on data including a dc sample (0 Hz)
-        nw = skrf.Network('./cst_example_4ports.s4p')
+        nw = skrf.Network('./skrf/tests/cst_example_4ports.s4p')
         vf = skrf.VectorFitting(nw)
         vf.vector_fit(n_poles_real=3, n_poles_cmplx=0)
         # quality of the fit is not important in this test; it only needs to finish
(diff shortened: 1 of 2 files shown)


FranzForstmayr:
Could you add a testcase please? 
Great fix anyway :)

Vinc0110:
Done. 
Are the tests not running on the updated code of this PR? They don't fail when I run them locally...

skrf/tests/test_vectorfitting.py
```diff
@@ -44,6 +44,14 @@ def test_no_convergence(self):
             vf.vector_fit(n_poles_real=0, n_poles_cmplx=5, fit_proportional=False, fit_constant=True)
             self.assertEqual(warning[-1].category, RuntimeWarning)
 
+    def test_dc(self):
+        # perform the fit on data including a dc sample (0 Hz)
+        nw = skrf.Network('./cst_example_4ports.s4p')
```
FranzForstmayr, line 49: The s4p files are usually not in the root folder, maybe that's the issue?

Vinc0110: That was it. Thanks!

jhillairet (approved):

jhillairet:
ok for merging

Issue #774, opened 2022-10-19
https://scikit-rf.readthedocs.io site is broken 


vishak-venkatraman:
<img width="769" alt="image" src="https://user-images.githubusercontent.com/67716193/196723271-90e5c51e-8ed2-4aaa-90fe-e11fb1c603c7.png">

FranzForstmayr:
I'll check later

FranzForstmayr:
Hmm, I'm unable to reproduce locally...
And it's strange that no build error is shown:
https://readthedocs.org/api/v2/build/18384426.txt

FranzForstmayr:
Opened a support ticket at readthedocs.org, hope I'll get an answer.

FranzForstmayr:
Fixed with #775

arsenovic:
fyi the calls the rf.styley() are causing errors in the docs still...

FranzForstmayr:
Should be fixed with the next build too

PR #775, opened 2022-10-19
Try to fix rtd build

FranzForstmayr:


FranzForstmayr:
Fixes #774

jhillairet:
what was the real reason for the RTD bug ? The visa package requirement?

FranzForstmayr:
Nah, the packaging from pyproject.toml. Due to the development install it worked on local machines but not on rtd

PR #776, opened 2022-10-19
use pyupgrade for python3.7+

FranzForstmayr:
This is an autogenerated change from the pyupgrade tool (https://github.com/asottile/pyupgrade)
Due to the drop of python3.6 I thought we could update the codebase to a newer syntax standard. For example old string formatters are converted to f-strings and old python2 code is removed.


Vinc0110:
Cool. I did not even know I'm not supposed to use `'a={}'.format(a)` anymore...

skrf/vi/scpi/parser.py
```diff
@@ -3,7 +3,7 @@
 import re
 import sys
 
-kwarg_pattern = re.compile('<([a-zA-Z0-9_]+=?[a-zA-Z0-9_, \(\)\'\"]*)>')
+kwarg_pattern = re.compile('<([a-zA-Z0-9_]+=?[a-zA-Z0-9_, \\(\\)\'\"]*)>')
```
jhillairet, line 6: is the result the same?

jhillairet (commented):
While I totally agree that f-string is an improvement in readability, I'm not that a fan of replacing the `set` and `dict` function with just `{}`. I find it less clear when reading, especially to understand when it is a set and not a dict...

skrf/vi/scpi/parser.py
FranzForstmayr: Hmm, i have to check

FranzForstmayr: Yes it is:

```python
import re
re1 = re.compile('<([a-zA-Z0-9_]+=?[a-zA-Z0-9_, \(\)\'\"]*)>')
re2 = re.compile('<([a-zA-Z0-9_]+=?[a-zA-Z0-9_, \\(\\)\'\"]*)>')
print(re1)
print(re2)
print(re1 == re2)
```

```
re.compile('<([a-zA-Z0-9_]+=?[a-zA-Z0-9_, \\(\\)\'"]*)>')
re.compile('<([a-zA-Z0-9_]+=?[a-zA-Z0-9_, \\(\\)\'"]*)>')
True
```

FranzForstmayr:
@jhillairet should  I return the `set`/`dict` related changes?

jhillairet:
> @jhillairet should I return the `set`/`dict` related changes?

is it easy? If not, no big deal.

FranzForstmayr:
I could do it manually of course, however I like the `dict` comprehension and find it more readable than a list comprehension of tuples. 
Your decision...

jhillairet:
Ok leave it like the PR. I was more concerned with the set. Knowing it is a
set rather than a list is important..

FranzForstmayr:
Ok, I'll check the sets

PR #779, opened 2022-10-21
SOLT.ipynb: distinguish offset and port impedances, close #754.

### LINKED ISSUE #754: Possible mistake in the 'SOLT Calibration Standards Creation' example
I have a self-built cal kit and I've been working on fitting Keysight's model to measured S-parameters, so that I can plug the coefficients into my E5062A VNA's cal kit definitions.

I've been using the code from the SOLT Calibration Standards Creation example and it's been extremely helpful (thank you @biergaizi!), but I noticed that it wasn't matching the VNA's internal model when `offset_z0` was set to anything other than 50.

In the example, when creating the mediums `z0` is set to 50:

https://github.com/scikit-rf/scikit-rf/blob/dde299c2f9e8c38f90665db6d90a22a6b6d831f1/doc/source/examples/metrology/SOLT%20Calibration%20Standards%20Creation.ipynb?short_path=ad9db82#L321-L337

However, in the summary at the end `z0` is set to `offset_z0`:

https://github.com/scikit-rf/scikit-rf/blob/dde299c2f9e8c38f90665db6d90a22a6b6d831f1/doc/source/examples/metrology/SOLT%20Calibration%20Standards%20Creation.ipynb?short_path=ad9db82#L553

I'm still getting my head around all the theory, but I *think* the latter code is a mistake and `z0` should also be set to 50 here. If I make that change, then the model matches what I see from the VNA very well.

FWIW, my test procedure looks like this: I do a least_squares fit to find suitable coefficients to match the model to my measured S-parameters, program those into the VNA, and do a response cal on that particular standard. I leave the standard connected, take a measurement, and then compare it to my stored S-params.

biergaizi:
Currently, in example code snippets, if a specific offset impedance is used, the VNA port impedance is also set to this matching value. The original justification was that, this way, specifying a 75-ohm offset impedance will create S-parameters as measured by a 75-ohm VNA, which is a reasonable default.

However, sometimes an offset impedance value slightly different from the port impedance is used to model imperfections in the standard. For example, a 50-ohm VNA standard may have a 50.209-ohm or 49.992-ohm offset impedance. Setting port impedance to offset impedance in this case is misleading and incorrect.

This commit updates the guide to highlight this fact in the explanation. The example code snippets are also updated with an additional parameter, allowing users to explicitly specify different offset and port impedances, followed by 3 usage examples. Thus, it fixes Issue #754.

Also, a brief explanation of the deprecation warning is added.


jhillairet (approved):

jhillairet:
OKfor me. @biergaizi can we merge it ?

PR #781, opened 2022-10-25
IEEEP370 Deembedding add differences compared to matlab in docstring

mhuser:
This PR add a warning in the docstring to inform matlab users of the differences compared to matlab original implementation.
The differences are:

- FIX-2 is flipped (there is already a numbering diagram in docstring)
- Better root choice solution to avoid 180° phase jumps in certain circumstances (NZC methods only)

Fix 780


mhuser:
Hum, I did only change some docstrings and now in fail Code linting and testing / Tests (x.y) 😲

FranzForstmayr:
This is a caused by a new pytest release (7.2) from yesterday.
Strange error message, seems to be related with pytest-cov in some way.

mhuser:
> This is a caused by a new pytest release (7.2) from yesterday. Strange error message, seems to be related with pytest-cov in some way.

Strange, I have opened the separate issue #783 to keep track of that.

skrf/calibration/deembedding.py
```diff
@@ -973,6 +973,14 @@ class IEEEP370_SE_NZC_2xThru(Deembedding):
         -|1  2|---|1  2|---|2  1|-
          +----+   +----+   +----+
          
+
+    Warning
+    -------
+    There are two differences compared to the original matlab implementation:
```
jhillairet, line 979: is there a reference available for the Matlab implementation (a link?)

jhillairet (approved):

skrf/calibration/deembedding.py
mhuser: yes, it is in the sources below the warning, but I will add it in the comment

jhillairet (approved):

mhuser:
@jhillairet Dare I merge even if the tests are not passing?

FranzForstmayr:
@mhuser could you merge master in your branch? The tests should pass afterwards :)

skrf/calibration/deembedding.py
```diff
@@ -1263,40 +1272,43 @@ def split2xthru(self, s2xthru):
...
                 if i > 0:
                     if npy.abs(-e01[i] - e01[i-1]) < npy.abs(e01[i] - e01[i-1]):
-                        e01[i] *= -1
-                e10[i] = e10e01[i] / e01[i]
+                        e01[i] = - e01[i]
+            e10 = np.sqrt(s12r * (1 - e111 * e112))
+            for i in range(n):
+                if i > 0:
+                    if npy.abs(-e10[i] - e10[i-1]) < npy.abs(e10[i] - e10[i-1]):
+                        e10[i] = - e10[i]
```
FranzForstmayr, line 1286: Is this functionality the same as `np.unwrap`?
https://numpy.org/doc/stable/reference/generated/numpy.unwrap.html

If yes, function would be much faster probably

FranzForstmayr: Sorry, did not recognice that you merged already :)

mhuser: No problem :)
It looks like unwrap but is not strictly similar: instead of adding `2*k*pi` to the value, it changes the angle of the complex with the value `pi`. This is because the square root solution can be positive or negative, so it takes the value that avoids a phase jump.

PR #782, opened 2022-10-25
remove matplotlib dependency from deembedding

mhuser:
Remove matplotlib strict dependency from deembedding.py using a decorator as discussed and to fix #778


jhillairet (commented):
Couldn't the `check_plotting` decorator be moved in the plotting.py file, and imported in deembedding.py and vectorfitting.py instead of being duplicated?

mhuser:
> Couldn't the `check_plotting` decorator be moved in the plotting.py file, and imported in deembedding.py and vectorfitting.py instead of being duplicated?

Yes, sounds like what @FranzForstmayr proposed. I will have a try.

mhuser:
Currently it does not works because `NameError: name 'check_plotting' is not defined`

mhuser:
Seems to work better as of now. The wrapper was missing a return statement for my wrapped function to return 😄

skrf/vectorFitting.py
```diff
@@ -6,39 +6,19 @@
 if TYPE_CHECKING:
     from .network import Network
 
-from functools import wraps
 try:
     from . import plotting    # will perform the correct setup for matplotlib before it is called below
-    import matplotlib.pyplot as mplt
+    from .plotting import check_plotting
+    import matplotlib.pyplot as plt
```
Vinc0110, line 12: Why this change from `mplt` to `plt`? Can't we just leave it as is? 

As far as I know, it's only a matter of personal taste without a right or wrong. I'm pretty sure I will get errors when trying to write `mplt` in a future update of the vector fitting code. This also introduces many useless pseudo changes in the code.

mhuser: I understand this. It is not my intention to enforce `plt` or `mplt` or any other mtplotlib.pyplot alias.
The reason was that the alias appears in the `check_plotting` wrapper in plotting.py so I picked one of two.
I wonder if I can add a parameter to the wrapper to just fit the syntax of the file it is used.

mhuser: @Vinc0110 I reverted back to `mplt` in `vectorFitting.py` and find a more creative solution to accommodate different alias names with a single wrapper.

Vinc0110: Very good. And sorry for the trouble. Had I known there was a good reason for this change, I'd be absolutely fine with it. Anyway, your solution is even better, thanks!

skrf/vectorFitting.py
```diff
@@ -6,9 +6,9 @@
 if TYPE_CHECKING:
     from .network import Network
 
-from functools import wraps
 try:
     from . import plotting    # will perform the correct setup for matplotlib before it is called below
+    from .plotting import check_plotting
```
FranzForstmayr, line 11: Have you tested this line? This should not be possible, when no matplotlib is installed.

skrf/plotting.py
```diff
@@ -82,6 +82,29 @@
 from . util import now_string_2_dt
 import matplotlib.tri as tri
 
+from functools import wraps
+    
+def check_plotting(matplotlib_pyplot_alias):
```
FranzForstmayr, line 87: A function like this should probably in __ini__.py or utils.py, otherwise we would not be able to run it.

skrf/vectorFitting.py
```diff
@@ -1388,7 +1368,7 @@ def get_model_response(self, i: int, j: int, freqs: Any = None) -> np.ndarray:
                 resp += residues[i] / (s - pole) + np.conjugate(residues[i]) / (s - np.conjugate(pole))
         return resp
 
-    @check_plotting
+    @check_plotting(mplt)
```
FranzForstmayr, line 1371: We could extend the decorator, so that mplt is automatically passed as kwargs for the decorated function. We would not be required to pass the matplotlib instance every time.

FranzForstmayr (changes requested):

skrf/vectorFitting.py
mhuser: Moved to `util.py`

skrf/plotting.py
mhuser: Moved to `util.py`

mhuser:
Huuum seems my decorator with parameter does not pass the tests

mhuser:
the error
```
if ax is None:
  >           ax = mplt.gca()
  E           AttributeError: 'NoneType' object has no attribute 'gca'`
```
Is strange because `NoneType` imply that `mplt is None`, but the decorator does not raise the `RuntimeError`.
If  `plt = None` is hard-coded in `deembedding.py`, the `RuntimeError` is raised but if  `mplt = None` is hard-coded in `vectorFitting.py` it trigger `AttribueError` instead.

skrf/vectorFitting.py
mhuser: I am not sure how to do this.

mhuser:
Unfortunately, I will be busy on another sprint these next weeks. I convert this PR as a draft as of now to avoid stalling version 0.24.0.

FranzForstmayr:
I'll try to have a look

mhuser:
@FranzForstmayr 
Should the decorator test only if `matplotlib.pyplot` is loaded in `sys.modules`?
This could remove the need to know its actual name, whatever it is, hence remove the argument from decorator.
[how-can-i-check-if-a-module-has-been-imported](https://stackoverflow.com/questions/30483246/how-can-i-check-if-a-module-has-been-imported)
But I wonder how we could unload the module for the testcase.

mhuser:
Solved by #785

PR #784, opened 2022-11-01
update to pytest-cov ~=4.0 and drop the deprecation warning ignore

altendky:
pytest-cov v4 updated to use the modern pytest hooks and avoids the warnings so they need not be ignored.

https://github.com/pytest-dev/pytest-cov/issues/561#issuecomment-1297143745


jhillairet:
Thank you @altendky.

PR #785, opened 2022-11-01
matplotlib as optional dependency

FranzForstmayr:
This PR makes matplotlib an optional dependency again and adds a testsetup which tries to test the most basic network commands without matplotlib installed.

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -5,6 +5,7 @@ import skrf
 import numpy as np
 import tempfile
 import os
+import sys
 import warnings
 
 
@@ -59,6 +60,9 @@ class VectorFittingTestCase(unittest.TestCase):
         # quality of the fit is not important in this test; it only needs to finish
         self.assertLess(vf.get_rms_error(), 0.2)
 
+    @pytest.mark.skipif(
+        "matplotlib" not in sys.modules, 
+        reason="Spice subcircuit uses Engformatter which is not available without matplotlib.")
     def test_spice_subcircuit(self):
         # fit ring slot example network
         nw = skrf.data.ring_slot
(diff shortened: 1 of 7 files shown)


skrf/tests/test_vectorfitting.py
```diff
@@ -101,9 +105,9 @@ def test_read_write_npz(self):
         self.assertTrue(np.allclose(vf.proportional_coeff, vf2.proportional_coeff))
         self.assertTrue(np.allclose(vf.constant_coeff, vf2.constant_coeff))
 
+    @pytest.mark.skipif("matplotlib" in sys.modules, reason="Raise Error only if matplotlib is not installed.")
```
mhuser, line 108: The test is usually not skipped because matplotlib is no more in the dependencies, is it like that ?

mhuser (approved):
Excellent, you definitively are a decorator master !

skrf/tests/test_vectorfitting.py
FranzForstmayr: This test is skipped for the default test runs (py37 to py311) as matplotlib is installed and no exception is raised.
The test is executed in the minimal setup and asserts the RuntimeError

mhuser: Great!

skrf/vectorFitting.py
```diff
@@ -1388,9 +1369,9 @@ def get_model_response(self, i: int, j: int, freqs: Any = None) -> np.ndarray:
                 resp += residues[i] / (s - pole) + np.conjugate(residues[i]) / (s - np.conjugate(pole))
         return resp
 
-    @check_plotting
+    @axis_kwarg
     def plot(self, component: str, i: int = -1, j: int = -1, freqs: Any = None,
-             parameter: str = 's', ax: mplt.Axes = None) -> mplt.Axes:
+             parameter: str = 's', *,ax = None):
```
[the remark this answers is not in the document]
FranzForstmayr: One should mention, that with this change `ax` is a keyword-only property. I hope this is fine @Vinc0110?

mhuser (approved):

skrf/vectorFitting.py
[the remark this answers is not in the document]
Vinc0110: It should be fine. I did not know this ugly syntax, but it seems to be according to [PEP 3102](https://peps.python.org/pep-3102/).

jhillairet (approved):
Nice and clean. Eventually, a comment could be added in the tox.ini to remember why this minimal env exists.

PR #790, opened 2022-11-05
Make qtapps runable again

FranzForstmayr:
This PR creates a optional qt dependency and installs the qtapps in the current environment.

Changes were quite small, so I don't think it's necessary to move the complete qtapps into a separate repository.


FranzForstmayr:
Quick Google research seems to be not promising, see this issue.
https://github.com/conda/conda/issues/2984

jhillairet (approved):

README.md
```diff
@@ -33,6 +33,14 @@ If you have [installed conda](https://docs.conda.io/projects/conda/en/latest/use
 conda install -c conda-forge scikit-rf 
 ```
 
+**scikit-rf** includes standalone gui applications which can be installed and used like this
+```Shell
+python -m pip install scikit-rf[plot,qtapps,visa]
```
jhillairet, line 38:
```suggestion
# Install the additional dependencies for the gui applications
python -m pip install scikit-rf[plot,qtapps,visa]
```

Question: I'm discovering this kind of installation method with pip. Is the same kind of install command exist with conda as well?

jhillairet (approved):
Nice PR !

skrf_qtapps/widget_test.py
```diff
@@ -6,7 +6,7 @@
 
 class nwa_dummy():
     def __init__(self, NPORTS=2, NCHANNELS=False):
-        self.NPORTS = NPORTS
```
dvincentwest, line 9: why is nports being changed to lower case but not nchannels?

dvincentwest (commented):
For the packaging attributes I defer since I am not familiar with them.

For the removal of the sip import and sip.destroyonexit(False) in qt.py I also defer since I haven't looked at that or run these in a long time.

Everything else looks fine except I don't understand the nwa_dummy initializer having two variables, nports and nchannels but only 1 is upper case.

skrf_qtapps/widget_test.py
FranzForstmayr: Hmm, I think this was by accident, as something failed. I'll try this again.

FranzForstmayr:
> For the removal of the sip import and sip.destroyonexit(False) in qt.py I also defer since I haven't looked at that or run these in a long time.

I didn't get this running, so I removed the line, and found no difference. If more people use the qtapps, we'll probably find another bug. However, I found it important, that there's an easier start with the qtapps.

FranzForstmayr:
@dvincentwest 
nports was accessed in widgets.py.
https://github.com/scikit-rf/scikit-rf/blob/master/qtapps/skrf_qtwidgets/widgets.py#L739

So it seems, that widget_test.py is not runnable since quite a while ago. I don't know the exact purpose, maybe we should delete this file.

jhillairet:
Is this OK for being merged?

FranzForstmayr:
Should I delete widget_test, or do we still need it?

jhillairet:
> Should I delete widget_test, or do we still need it?

If it is not usable since long ago, I suggest to delete it.

Issue #718, opened 2022-06-19
Towards a version 1.0.0


jhillairet:
### Discussed in https://github.com/scikit-rf/scikit-rf/discussions/692

According to recommended [semantic versioning](https://semver.org/):

> How do I know when to release 1.0.0?
> 
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

scikit-rf should probably already be released as 1.0.0. In addition, I've met some potential users who said to me that looking at `scikit-rf`, they saw the version number as being `0.XY` and deduced that the code was highly unstable, then drop and didn't look it further.

So I think scikit-rf is (soon to be) ready for a release 1.0.0.

This issue aims to track the desired improvements we want before releasing 1.0.0, a list which is open to discussion:

- [x] Use SI units ('Hz', 's') instead of ('GHz', etc.) (done in #935 )
   - [x] but warn users before doing it!
- [x] simplify `Media` definition of characteristic and ports impedances (done in #727 )
- [x] remove obsolete parts of the code (done in #936):
  - [x] `/examples` dir
  - [x] `/apps` dir ? Is someone use it?
  - [x] `Network2` ? (despite its very nice features, it now departs heavily from `Network` and is not maintained)
- [x] fix pyvisa issues and improve user experience with hardware (#744)
- [x] fix QtApps issues (done in #790)
- [ ] fix most old issues
- [x] support Touchstone V2

FranzForstmayr:
Should we move the qtapps in a separate repository? Would be easier to maintain probably?

dvincentwest:
@FranzForstmayr I was just reviewing the commit history for qtapps, and it doesn't see a ton of action.  In addition I know there is much needed work underway to revamp the virtual instrument code, in part to support a different GUI framework that uses skrf.  Those changes will likely break some behavior in the qtapps, and without active maintenance, it probably doesn't make sense to keep qtapps around in the main repo.

jhillairet:
I guess we are ready now for a 1.0 release, isn't it?

cafeclimber:
I think the apps directory still needs to be removed or moved to another repo, Network2 removed, and touchstone 2 support finalized before then at the least

mhuser:
We sounds now very close.

PR #792, opened 2022-11-06
Warn users about future change of the default frequency unit

jhillairet:
Future versions of scikit-rf (cf dicussions in #718 and #692) will use SI units, so "Hz" a the default unit and not "GHz" as it is currently.

Before making the change, one or few intermediate versions should warn users of the future change to come, and recommend passing the unit explicitly to avoid future compatibility problems in their codes.


mhuser (approved):
Nicely done.
As of now, we will have warnings in some notebooks.

FranzForstmayr:
LGTM

PR #799, opened 2022-11-16
Time ns warning

FranzForstmayr:
Similar to #792 this PR prepares skrf to use seconds instead of nanoseconds by default.

### DIFF
--- a/doc/source/examples/networktheory/Time Domain.ipynb
+++ b/doc/source/examples/networktheory/Time Domain.ipynb
@@ -54,7 +54,7 @@
     "s11 = probe.s11 \n",
     "\n",
     "#  time-gate the first largest reflection\n",
-    "s11_gated = s11.time_gate(center=0, span=.2)\n",
+    "s11_gated = s11.time_gate(center=0, span=.2, t_unit='ns')\n",
     "s11_gated.name='gated probe'\n",
     "\n",
     "# plot frequency and time-domain s-parameters\n",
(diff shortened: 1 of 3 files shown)


FranzForstmayr:
Hmm `time.detect_span` still returns span in nanoseconds. I'm not sure, if this function is used from anyone outside `skrf`?

jhillairet (approved):

FranzForstmayr:
@jhillairet do you think we need a public `detect_span` function? Otherwise we could deprecate this public function which returns nanoseconds and use an internal `_detect_span`.
If it's need, I'll need to add a `t_unit` keyword argument.

FranzForstmayr:
Can we get this into #800 ?

jhillairet:
> @jhillairet do you think we need a public `detect_span` function? Otherwise we could deprecate this public function which returns nanoseconds and use an internal `_detect_span`. If it's need, I'll need to add a `t_unit` keyword argument.

I have no idea how much this function is used. 

I think raising a warning in this function saying it will return time in seconds in the future is sufficient for now, and it will be included in #800. 

Then, after, and towards v1.0, it raises the need to question the [`Frequency.t_ns`](https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.frequency.Frequency.t_ns.html) attribute itself (to keep or not), and adding adding a `unit`  param to this function.

jhillairet:
OK.There is print left over, but this could be corrected later

PR #802, opened 2022-11-21
Use fft_window when specified and not window

FranzForstmayr:
Currently we check if `fft_window` parameter is not None, but we use `window` parameter to get the window.
This PR uses the `fft_window` as argument for `get_window`.


Vinc0110 (approved):
Wow, that was not my brightest moment. Good catch!

By the way, I propose to set the default to `fft_window='cosine'` while you are at it. It seems to work best in my tests.

jhillairet (approved):

PR #807, opened 2022-11-29
Draft: Set subnetwork properties on instance instead of the class

eendebakpt:
After importing skrf and creating a 1-port network, the network as attributes `s13', `s3_3` which it should not have (since it is a 1-port network). A minimal example:
```
import numpy as np
import skrf as rf

f = rf.Frequency.from_f(np.linspace(2e6, 3e6, 10), unit="Hz")
self=n=rf.Network(frequency=f, s=np.random.rand(10,)*1j, name='test')
# now n has properties s_12, which it should not have

w=[p for p in dir(self) if p.startswith('s') and p[-1] in ['0', '1','2','3','4']]
print(w)
```
has output
```
['s11', 's12', 's13', 's1_1', 's1_2', 's1_3', 's21', 's22', 's23', 's2_1', 's2_2', 's2_3', 's31', 's32', 's33', 's3_1', 's3_2', 's3_3', 's_db10']
```

This PR fixes that by setting the attributes on the instance, instead of the class.

### DIFF
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -208,7 +208,6 @@ class Frequency:
         if isinstance(key, str):
 
             # they passed a string try and do some interpretation
-            re_numbers = re.compile(r'.*\d')
             re_hyphen = re.compile(r'\s*-\s*')
             re_letters = re.compile('[a-zA-Z]+')
 
@@ -279,7 +278,7 @@ class Frequency:
         if npy.isscalar(f):
             f = [f]
         temp_freq =  cls(0,0,0,*args, **kwargs)
-        temp_freq._f = npy.array(f) * temp_freq.multiplier
+        temp_freq._f = npy.asarray(f) * temp_freq.multiplier
         temp_freq.check_monotonic_increasing()
 
         return temp_freq
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -188,10 +188,17 @@ from .constants import S_DEFINITIONS, S_DEF_DEFAULT
 if TYPE_CHECKING:
     import pandas as pd
 
-#from matplotlib import cm
-#import matplotlib.pyplot as plt
-#import matplotlib.tri as tri
-#from scipy.interpolate import interp1d
+def add_property(inst : object, name : str, method : Callable, doc : Optional[str] = None):
+    """ Add a property to an instance
+    
+    Adapted from: https://stackoverflow.com/a/2954373/19469050
+    """
+    cls = type(inst)
+    if not hasattr(cls, '__perinstance'):
+      cls = type(cls.__name__, (cls,), {})
+      cls.__perinstance = True
+      inst.__class__ = cls
+    setattr(cls, name, property(method, doc=doc))
 
 class Network:
     r"""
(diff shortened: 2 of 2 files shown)


FranzForstmayr:
Nice catch!

eendebakpt:
@FranzForstmayr This is more complicated than I expected. Setting a property on an instance is not really possible. I used some code from stackoverflow to make this work (see commit 
fc45854), but now the equality checks are failing. I could try to fix that, but we risk adding all kinds of non-standard workarounds.

An alternative option is to subclass `Network` into `Network1port`, `Network2port`, ..., `NetworkManyPort` and set the appropriate properties for each subclass.

And a mitigation could be to modify the `__dir__` of Network to filter out values of `sm_n` that are not valid for the number of ports.

skrf/frequency.py
```diff
@@ -208,7 +208,6 @@ def __getitem__(self, key: Union[str, int, slice]) -> 'Frequency':
         if isinstance(key, str):
 
             # they passed a string try and do some interpretation
-            re_numbers = re.compile(r'.*\d')
```
jhillairet, line 211: Why does this line have been removed? Wasn't used after?

jhillairet:
> An alternative option is to subclass `Network` into `Network1port`, `Network2port`, ..., `NetworkManyPort` and set the appropriate properties for each subclass.

I'm not a big fan of this solution.
 
>  filter out values of `sm_n` that are not valid for the number of ports.

Removing properties `smn` and `sm_n` for (m,n) higher than the number of ports in `__generate_subnetworks` should solve the issue.

skrf/frequency.py
eendebakpt: No, it was not used in the file

eendebakpt:
> > An alternative option is to subclass `Network` into `Network1port`, `Network2port`, ..., `NetworkManyPort` and set the appropriate properties for each subclass.
> 
> I'm not a big fan of this solution.
Me neither.
> 
> > filter out values of `sm_n` that are not valid for the number of ports.
> 
> Removing properties `smn` and `sm_n` for (m,n) higher than the number of ports in `__generate_subnetworks` should solve the issue.
Removing the properties for high (m,n) in `__generate_subnetworks` does not work: the properties are set on the `Network` class, and not on the different `Network` instances. Since different `Network` instances can have different values of `m` and `n`, we will get incorrect `s` properties as soon as we have `Network` instances with different number of ports.

Since `dir` works on the instance we can modify that. The incorrect `smn` parameters are still part of the object, but they won't show any more in auto-completion. It is only a (partial) mitigation, but at this moment I see no other (backward compatible) solutions.

@jhillairet

FranzForstmayr:
> An alternative option is to subclass Network into Network1port, Network2port, ..., NetworkManyPort and set the appropriate properties for each subclass.

It's not possible to choose a parent class dynamically. Apart from that it's not very beautiful.

> And a mitigation could be to modify the __dir__ of Network to filter out values of sm_n that are not valid for the number of ports.

This should work probably, but isn't a clean solution either I think. I'll start the investigate later today hopefully.

FranzForstmayr:
> Since dir works on the instance we can modify that. The incorrect smn parameters are still part of the object, but they won't show any more in auto-completion. It is only a (partial) mitigation, but at this moment I see no other (backward compatible) solutions.

Does s11 and so on show up in autocompletion now?

FranzForstmayr:
image

vscode fails to show completion

eendebakpt:
In Jupyterlab the values do not show in autocompletion:

image

jhillairet:
@FranzForstmayr @eendebakpt what about this PR?

eendebakpt:
Closing in favor of #810

PR #809, opened 2022-12-06
Run secondary property generation only once

FranzForstmayr:


FranzForstmayr:
Will close this, as functionality was merged into #810

PR #810, opened 2022-12-06
Do not generate subnetworks all the time but provide them dynamically

FranzForstmayr:
Do not generate subnetworks all the time but provide them dynamically if requested.

This is an alternative approach for #807, open for discussion


FranzForstmayr:
Subnetworks aren't shown in docs anymore...

jhillairet:
> Subnetworks aren't shown in docs anymore...

and I guess neither in the autocompletion, isn't?

FranzForstmayr:
I don't think so. I work with vscode, which hadn't autocompletion for this dynamically generated properties anyway.

FranzForstmayr:
I did some research about dynamic properties and autocompletion today.

It's not possible to get static typing or type hints (as for mypy or any other typechecker) with dynamically created properties, as the type checker does not execute the code even once, that's why it's not aware of the dynamic attributes.

Interactive Sessions like jupyter or ipython have a concept which is called autosuggestion which can handle dynamic properties. The list of available properties is taken from the __dir__() method.

So this PR removes dynamic property setting and generates all available dynamic attributes on the fly, as the __getattr__() method is overloaded. Autosuggestion for interactive Sessions and Jupyter is done via the __dir__ method.

Open for Feedback :)

If this works well, this method could be expanded to replace the code, which creates the plotting methods for a Network or NetworkSet

FranzForstmayr:
Still have to investigate sphinx build

jhillairet:
what about what has been proposed in #828 concerning secondary properties?

FranzForstmayr:
The secondary property generation is error prone in general and wouldn't be needed anymore.

FranzForstmayr:
I think we can start review at this point, although I think we should not merge before 0.25 release. A follow up PR #848 will unify the plot code generation.

jhillairet (approved):
very nice addition, hats off!

FranzForstmayr:
I think this one is fine. #848 already includes this PR, but isn't finished yet.

jhillairet (approved):

PR #814, opened 2022-12-16
Changed default network interpolation from cart (re/im) to polar

graybc:
With 's' as the default basis for interpolation, it seems to me that polar might make more sense as a default.

Let me know if this doesn't jibe with existing design philosophy. This is a great package, thanks so much for all the hard work!

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2584,7 +2584,7 @@ class Network:
 
     # interpolation
     def interpolate(self, freq_or_n: Union[Frequency, NumberLike], basis: str = 's',
-                    coords: str = 'cart', f_kwargs: dict = {}, return_array: bool = False,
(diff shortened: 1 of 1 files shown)


arsenovic:
thanks!.
i would recommend against this.  for respones that trace out circles that
are not centered at 0,0 it can lead to strange and i would say unexpected
results.

happy to hear others thoughts too

FranzForstmayr:
In metrology applications also cart is used usually, not polar.
I think there's no general true or false, it depends, as so often, on your data.

Vinc0110:
Given that this change is only about the default value, not about functionality, I also don't see the need. 
Apparently, 'cart' is fine for a lot of users. Others are already able to use 'polar'.

@graybc:  Feel free to explain the problem in more detail. So far, I'm against changing it.

PR #828, opened 2023-01-25
Improve performance of Network.copy

eendebakpt:
In a copy of a `Network` new `s` parameters are assigned. This triggers a call to `__generate_secondary_properties` which is not required as `__generate_secondary_properties` depends only on class properties, not on instance properties (or the values of the s parameters)

On this data
```
import numpy as np
import skrf as rf

f = rf.Frequency.from_f(np.linspace(2e6, 3e6, 10), unit="Hz")
self=n=rf.Network(frequency=f, s=np.random.rand(10,)*1j, name='test')
``` 
a benchmark `%timeit n.copy()` is improved by:
```
Master: 1.06 ms ± 143 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
PR: 217 µs ± 32.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -343,6 +343,8 @@ class Network:
     Default interpolation method.
     """
 
+    _secondary_properties_generated = False
+
     # CONSTRUCTOR
     def __init__(self, file: str = None, name: str = None, params: dict = None,
                  comments: str = None, f_unit: str = None,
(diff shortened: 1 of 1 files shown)


jhillairet (approved):
Great and smart!

Issue #815, opened 2022-12-17
OnePort calibration leads to very sparse corrected Network when using different frequency ranges


cactorium:
So I'm using the OnePort to do one port calibration of VNA data, and it looks like the output Network tends to have very few points when different frequency range settings are used.

For example, if I were to do something like this:
```
cal = OnePort(measured=[...], ideals=[...]) # networks with frequency range like 1 MHz-2 GHz, 5000 pts
cal.run()
net = skrf.Network("test.s1p") # network with frequency range like 1MHz-10MHz, 1000 pts
net_corrected = cal.apply_cal(net)
```

`net_corrected` would only have one point at 1 MHz. I've traced this to `OnePort`, where it cascades the correction network with the given network. The issue is that the cascade operation causes the output network to only have the points where there's an exact match between the frequencies collected to do the calibration and the frequencies used in the measurement, which doesn't tend to happen very often if the frequency ranges aren't exactly the same.

Would it be possible to change `OnePort` to interpolate the correction network to match the input frequency range? I understand the point of only using the exact matching points to ensure accuracy in the general case of network cascades, but it seems very limiting to have the calibration only work as expected when the frequency ranges are precisely matched. I can write a pull request if this seems like a reasonable change

Vinc0110:
A warning would probably be a good idea in such cases.
Still, why don't you just provide both networks with identical frequency vectors, as expected by the calibration routine?

If you have measured your calibration at different frequencies than your data to be calibrated (not ideal), you can always interpolate either or both of them:
https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.network.Network.interpolate.html

cactorium:
It's for cases where I'm zooming into a particular frequency range to examine the S parameters more closely, or where I'm using a coarser frequency grid to reduce the scanning time.

Yeah, the issue is that there's no way to get the network from `OnePort` to interpolate it to match the input network's frequency range, because it's calculated and cascaded internally. I'm writing a small library to work with a MiniVNA Tiny, and I want the output network to have the frequency range the user requested, so the output network's range preferably should match the input network range. Ideally the calibration network needs to be interpolated into the input network range, but the network is never exposed so I can't perform that interpolation. I could interpolate to match the calibration's frequencies and then interpolate back into the input network's frequency range, but that would add a lot more error than necessary

[...]

PR #832, opened 2023-02-04
Interpolate calibration network to match input network

cactorium:
For https://github.com/scikit-rf/scikit-rf/issues/815 , I did some basic testing and it seems to work, would you want me to write some test cases for it?

EDIT: I guess the current test cases are doing a good job, gimme a bit

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -1106,7 +1106,9 @@ class OnePort(Calibration):
         s11 = self.coefs['directivity']
         s22 = self.coefs['source match']
         er_ntwk.s = npy.array([[s11, s12],[s21,s22]]).transpose(2,0,1)
-        er_ntwk = er_ntwk.interpolate(ntwk.frequency)
+        # if the frequencies aren't a perfect match, interpolate
+        if er_ntwk.frequency != ntwk.frequency.f and er_ntwk.frequency.npoints > 1 and ntwk.frequency.npoints > 1:
+          er_ntwk = er_ntwk.interpolate(ntwk.frequency)
         return er_ntwk.inv**ntwk
 
     def embed(self,ntwk):
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -1106,6 +1106,7 @@ class OnePort(Calibration):
         s11 = self.coefs['directivity']
         s22 = self.coefs['source match']
         er_ntwk.s = npy.array([[s11, s12],[s21,s22]]).transpose(2,0,1)
+        er_ntwk = er_ntwk.interpolate(ntwk.frequency)
         return er_ntwk.inv**ntwk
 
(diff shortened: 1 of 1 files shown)


jhillairet (approved):

Ttl:
I don't like this too much as a silent default option since interpolating raw measurements in general doesn't work. It requires that the frequency spacing is small enough that there is no aliasing in the error network. Even in the case that it is fulfilled interpolating loses accuracy since linear interpolation of S-parameters is not very accurate.

cactorium:
Would enabling it via a keyword flag be a reasonable compromise? As in
`cal.apply_cal(meas, interpolate=True)`

And would it be better to interpolate in some different space?

Ttl:
Here is an example demonstrating the problem with interpolation:

```python
import skrf
skrf.stylely()

NPTS = 101

freq1 = skrf.F(75, 100, NPTS, unit='GHz')
freq2 = skrf.F(75, 100, 2*NPTS, unit='GHz')

wg  = skrf.RectangularWaveguide(freq1, a=100*skrf.mil, z0=50, rho='gold')
wg2 = skrf.RectangularWaveguide(freq2, a=100*skrf.mil, z0=50, rho='gold')

# Error networks
E = wg.thru()**wg.line(200, 'mm', z0=40)**wg.thru()
E2 = wg2.thru()**wg2.line(200, 'mm', z0=40)**wg2.thru()

# Calibrate using freq1 media
ideals = [
        wg.short( name='short'),
        wg.delay_short( 45.,'deg',name='ew'),
        wg.delay_short( 90.,'deg',name='qw'),
        wg.match( name='load'),
        ]
measured = [E**k for k in ideals]

cal = skrf.OnePort(
    is_reciprocal = True, 
    ideals = ideals, 
    measured = measured,
    )

# Same DUT with different frequencies
dut = E**wg.line(10, unit='mm', z0=20)**wg.resistor(50)**wg.short()
dut2 = E2**wg2.line(10, unit='mm', z0=20)**wg2.resistor(50)**wg2.short()

# Calibrate the DUT using the same cal
cal.apply_cal(dut).plot_s_db(label='Original')
cal.apply_cal(dut2).plot_s_db(label='Interpolated')
```

cal_interp

The result improves if the number of points is raised, but for some reason some artifacts remain even with high number of points.

I would prefer that user would interpolate measurements themselves if they think it's possible in their case.

Vinc0110:
I agree with @Ttl.
For the problem with the frequency zoom described in #815, the right thing would probably be to create two separate calibrations: one for the coarse grid, another for the finer one.

cactorium:
Huh, I guess I didn't notice that in my testing because I was looking at a Smith chart at values fairly close to the center, it looks like the interpolation doesn't work as well as I thought. No other settings seem to improve it so I guess I'll adjust the frequency settings before starting a run to coincide with the calibration grid

PR #834, opened 2023-02-06
Overlap dut and dummy networks in deembedding

ericwrice:
This addresses the TODO comments in each of the deembedding classes to overlap the dut and dummy networks if their frequencies are not equal. Also adds an overlap_multi() convenience function in network to handle n number of overlap operations. The raised errors have been changed to runtime warnings. The tests probably need to be expanded to include networks with different frequencies.

### DIFF
--- a/skrf/calibration/deembedding.py
+++ b/skrf/calibration/deembedding.py
@@ -239,10 +239,10 @@ class OpenShort(Deembedding):
             warnings.warn('Network frequencies dont match dummy frequencies, attempting overlap.', RuntimeWarning)
             caled, op, sh = overlap_multi([ntwk, self.open, self.short])
         else:
-            caled, op, sh = ntwk.copy(), self.open, self.short.copy()
+            caled, op, sh = ntwk.copy(), self.open, self.short
 
         # remove parallel parasitics from the short dummy
-        deembeded_short = sh
+        deembeded_short = sh.copy()
         deembeded_short.y = sh.y - op.y
         # remove parallel parasitics from the dut
         caled.y = caled.y - op.y
@@ -430,11 +430,11 @@ class ShortOpen(Deembedding):
             warnings.warn('Network frequencies dont match dummy frequencies, attempting overlap.', RuntimeWarning)
             ntwk, op, sh = overlap_multi([ntwk, self.open, self.short])
         else:
-            op, sh = self.open.copy(), self.short
+            op, sh = self.open, self.short
 
         caled = ntwk.copy()
         # remove series parasitics from the open dummy
-        deembeded_open = op
+        deembeded_open = op.copy()
         deembeded_open.z = op.z - sh.z
         # remove parallel parasitics from the dut
         caled.z = ntwk.z - sh.z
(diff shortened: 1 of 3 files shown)


jhillairet (approved):
Great, thank you!

jhillairet:
@ericwrice, do you plan to extend the tests soon, or can I merge?

ericwrice:
@jhillairet, I think you're good to merge. I'd be happy to extend the tests later if that's desirable.

skrf/calibration/deembedding.py
```diff
@@ -236,17 +236,16 @@ def deembed(self, ntwk):
...
         if ntwk.frequency != self.open.frequency:
-            raise(ValueError('Network frequencies dont match dummy frequencies.'))
-
-        # TODO: attempt to interpolate if frequencies do not match
-
-        caled = ntwk.copy()
+            warnings.warn('Network frequencies dont match dummy frequencies, attempting overlap.', RuntimeWarning)
+            caled, op, sh = overlap_multi([ntwk, self.open, self.short])
+        else:
+            caled, op, sh = ntwk.copy(), self.open, self.short.copy()
```
keikawa, line 242: This code certainly works, but it might confuse ```self.short``` and ```deembeded_short```.
How about the following implementation?
```
caled, op, sh = ntwk.copy(), self.open, self.short

deembeded_short = sh.copy()
deembeded_short.y = sh.y - op.y
```

skrf/calibration/deembedding.py
```diff
@@ -428,17 +427,17 @@ def deembed(self, ntwk):
 
         # check if the frequencies match with dummy frequencies
         if ntwk.frequency != self.open.frequency:
-            raise(ValueError('Network frequencies dont match dummy frequencies.'))
-
-        # TODO: attempt to interpolate if frequencies do not match
+            warnings.warn('Network frequencies dont match dummy frequencies, attempting overlap.', RuntimeWarning)
+            ntwk, op, sh = overlap_multi([ntwk, self.open, self.short])
+        else:
+            op, sh = self.open.copy(), self.short
```
keikawa, line 433: The same as above.
```
op, sh = self.open, self.short

deembeded_open = op.copy()
deembeded_open.z = op.z - sh.z
```

skrf/calibration/deembedding.py
```diff
@@ -524,14 +523,14 @@ def deembed(self, ntwk):
 
         # check if the frequencies match with dummy frequencies
         if ntwk.frequency != self.short.frequency:
-            raise(ValueError('Network frequencies dont match dummy frequencies.'))
-
-        # TODO: attempt to interpolate if frequencies do not match
+            warnings.warn('Network frequencies dont match dummy frequencies, attempting overlap.', RuntimeWarning)
+            ntwk, sh = overlap_multi([ntwk, self.short])
+        else:
+            sh = self.short.copy()
```
keikawa, line 529: It can be as simple as ``sh = self.short`` similar to the open deembedding because it does not overwrite ```self.short```.

keikawa (commented):
Thank you for implementing this long awaited TODO😀

skrf/calibration/deembedding.py
ericwrice: Yep, thanks for pointing that out, I think this is a better way to do it!

skrf/calibration/deembedding.py
ericwrice: Good catch, no reason to make a copy here

keikawa (approved):
Thanks!

PR #835, opened 2023-02-06
Fix inductor units in media docstrings

ericwrice:


Vinc0110:
Good catch, inductance should not be in Farad.
Personally, I would prefer the singular form for the unit, i.e. "Henry".

ericwrice:
I can certainly change it to singular if that is preferred, I went with plural to be consistent with the resistor and capacitor doc strings.

Vinc0110:
It's probably not that important. Keeping it consistent is also a good thing.

Just for future reference, there exists a NIST guide with rules for spelling SI units:
https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-9-rules-and-style-conventions-spelling-unit-names

Following those rules would certainly require a major revision on many docstrings...

jhillairet (approved):
Good catch, thank you !

Issue #742, opened 2022-08-17
skrf.plotting.shade_bands(): fix y_range


joergschoebel:
skrf.plotting.shade_bands() has a parameter **y_range**, which should allow to set the y range of the shaded area.
However, in the code y_range is always overwritten:
`y_range=plt.gca().get_ylim()`

This should only occur if y_range is not passed - such as (including some rudimentary error checking):
```
if not isinstance(y_range, tuple) or (len(y_range) != 2):
    y_range=plt.gca().get_ylim()
```

PR #838, opened 2023-02-09
Fix shade_bands overwriting y_range (#742)

ericwrice:
Implements the fix to issue #742 to prevent the y_range parameter in plotting.shade_bands from being overwritten.


Vinc0110 (approved):

jhillairet (approved):
Thanks!

PR #844, opened 2023-02-11
Add test for input networks of OnePort calibration

eendebakpt:
If for a `OnePort` calibration by accident the standards are entered as `[two_port_reflect(short, short), two_port_reflect(open, open), two_port_reflect(load, load)]` the calibration succeeds, but it cannot be applied (error: `LinAlgError: Singular matrix`).
Also expressions like `calibration.coefs_ntwks['reflection tracking'].plot_s_db()` fail. 

This PR adds a check on the input networks and a test.


jhillairet (approved):
Thanks!

PR #845, opened 2023-02-13
Vector Fitting: Custom pole initialization

Vinc0110:
As [requested in the mailing list](https://groups.google.com/g/scikit-rf/c/VrX80hlDxx4), this PR adds an option for custom pole spacing by the user.

The pole spacing is a delicate part of the fitting process and this option is supposed to be used by advanced users only.

The goal was to keep the API unchanged, and I hope this is an elegant and usable way to do it.

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -11,10 +11,6 @@ import warnings
 
 class VectorFittingTestCase(unittest.TestCase):
 
-    msg_passivity_violation = 'The fitted network is passive, but the vector fit is not passive. Consider running ' \
-                              '`passivity_enforce()` to enforce passivity before using this model.'
-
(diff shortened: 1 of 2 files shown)


andree-sc:
Hey, thanks a lot, looks fine, except for one small thing I noticed: in line 187ff. I think a check should be made whether or not the instance variable "self.poles" holds any poles... I don't think it's that relevant, but it's the only thing I noticed now.
Other than that, nice work. I think it's an elegant solution and it doesn't change anything about how the API works *thumbsup*

jhillairet (approved):
OK. Would like it to be included in the next release ?

Vinc0110:
I don't understand why the tests are failing. 
@FranzForstmayr Any idea? Is it because of that decorator to filter warnings?

Vinc0110:
Ok now. I moved the tests to a separate PR (#849). This one is ready.

Vinc0110:
@andree-sc: Thanks for the review! I added the parameter check.

PR #848, opened 2023-02-13
Plot attributes

FranzForstmayr:


FranzForstmayr:
`NetworkSet` still has some strange dynamic plotting methods, but `plotting.py` is much cleaner now.

FranzForstmayr:
We should really ask our self, if we need that much methods on a class like NetworkSet.
Before the autogeneration of attributes in NetworkSet's constructor the `__dir__()` object has 81 entries. This is not less, but ok.
After the generation we have 952!! entries, most of them autogenerated plot methods, in `__dir__()`.

This PR should not make any functional difference to any user, but we really need to tighten the API in the future.

jhillairet:
> `NetworkSet` still has some strange dynamic plotting methods, but `plotting.py` is much cleaner now.

There are the same changes in the PR #852, could say essentially what/where are the difference? I'm a bit lost...

> We should really ask our self, if we need that much methods on a class like NetworkSet.

This is probably too much you're right,  but it's hard to say what users are really using or not....

FranzForstmayr:
#852 is done after this PR, that's why all the changes are in both PRs.
I'll keep #852 as draft, as long this isn't finished.

jhillairet:
what about this one for the next release ?

FranzForstmayr:
There are some docstring which need some more work. I think we should wait for the next one.
Furthermore I wanted to create unittest for the plot functions, they should not raise something unexpected

jhillairet:
@FranzForstmayr could you look to the conflicts and make a brief summary of the purpose of the the changes ?

FranzForstmayr:
I'm ready to merge if ok

jhillairet (approved):

PR #850, opened 2023-02-15
Vector Fitting: Tests

Vinc0110:
This is a branch from PR https://github.com/scikit-rf/scikit-rf/pull/845 to work on the tests separately. Work in progress...

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -12,7 +12,7 @@ class VectorFittingTestCase(unittest.TestCase):
     msg_passivity_violation = 'The fitted network is passive, but the vector fit is not passive. Consider running ' \
                               '`passivity_enforce()` to enforce passivity before using this model.'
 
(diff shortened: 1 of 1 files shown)


skrf/tests/test_vectorfitting.py
```diff
@@ -1,40 +1,50 @@
...
 import sys
-import warnings
 
 
 class VectorFittingTestCase(unittest.TestCase):
 
+    msg_passivity_violation = 'The fitted network is passive, but the vector fit is not passive. Consider running ' \
+                              '`passivity_enforce()` to enforce passivity before using this model.'
+
+    @pytest.mark.filterwarnings(f'ignore:*{msg_passivity_violation}*:UserWarning')
```
FranzForstmayr, line 15: Have you tried without the `*` character?

Vinc0110: It wasn't the `*`, but it now works in general form, without any message. That filters every UserWarning, which I think is fine in this case.

PR #852, opened 2023-03-09
Tests pass without optional dependencies

FranzForstmayr:
This is just a follow up PR to #848 to allow running all tests without any optional dependency.


tox.ini
```diff
@@ -35,4 +35,4 @@ basepython = python3.9
 extras = 
     test
 commands =
-    python -m pytest skrf/tests/test_network.py skrf/tests/test_vectorfitting.py 
\ No newline at end of file
+    python -m pytest {posargs}
```
jhillairet, line 38: Following your PR, all the setup is no more required, ins't it?

skrf/network.py
```diff
@@ -172,28 +170,25 @@
...
 if TYPE_CHECKING:
     import pandas as pd
 
-#from matplotlib import cm
-#import matplotlib.pyplot as plt
-#import matplotlib.tri as tri
-#from scipy.interpolate import interp1d
-
-class Network:
+class Network(PlottingMixin, BaseNetwork):
```
jhillairet, line 191: To understand your approach, what is the purpose of this new BaseNetwork ABC class? Properly separate the base properties from the plotting ones?

skrf/network.py
```diff
@@ -306,10 +300,18 @@ class Network:
         'time_impulse': None,
         'time_step': None,
     }
+
     """
     Component functions like 're', 'im', 'mag', 'db', etc.
     """
 
+    @classmethod
+    @lru_cache()
```
jhillairet, line 309: For my knowledge: did you test if the use or lru_cache improved performance?

skrf/circuit.py
```diff
@@ -1290,3 +1299,115 @@ def voltages_external(self, power: NumberLike, phase: NumberLike) -> np.ndarray:
         for port_idx in self.port_indexes:
             Vs.append((a[port_idx] + b[:,port_idx])*np.sqrt(z0s[:,port_idx]))
         return np.array(Vs).T
+
+    def plot_graph(self, **kwargs):
```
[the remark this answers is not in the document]
FranzForstmayr: Autocompletion and typing support works only for the `real` methods and not for the dynamically added methods, so this would be the prefered way. iPython autosuggestion works in both cases.

tox.ini
FranzForstmayr: I think it's required to test, if our function will work without matplotlib installed, but we don't have to specify a subset of the tests. Testcases which require matplotlib are skipped in this case.

skrf/network.py
FranzForstmayr: I thought about creating a base class, where Network and NetworkSet could inherit from. But I'm not sure yet. As said, NetworkSet improvements are WIP.

skrf/network.py
FranzForstmayr: No I hadn't. As the function is called only a few times during module initialization, it would not be much either.

FranzForstmayr:
Closed in favor of #848

PR #744, opened 2022-08-21
Start new VNA implementation

cafeclimber:
This PR seeks to update the virtual instrument module, but primarily the VNA submodule. The primary goals are:

- Simplify adding support for new instruments by creating a robust VNA baseclass
- Improve the channel interface for those instruments that support channels
- Remove the need to write SCPI command trees that have to be parsed separately to generate code. Instead SCPI and other commands can all be included in the source for the individual instrument's class
- Add unit tests for the currently supported instruments

There is also now updated documentation to make the process of contributing additional instrument support more simple.

Let me know what y'all think and if / how we should move forward!

Summary of changes:

- Rewrote VNA base class
- Removed vi.sa and vi.stages
- Updated documentation
- Wrote new implementations of Keysight FieldFox, Keysight PNA, and NanoVNA to fit the new architecture

TODO:

- [x] Write docs
- [x] Write tests
- [ ] Implement those instruments which were already implemented in the previous version
    - [X] Implement Keysight FieldFox
    - [X] Implement Keysight PNA
    - [X] Implement NanoVNA
    - [ ] Implement HP 8510c
    - [ ] Implement Rhode&Schwarz ZVA
- [ ] Test on actual hardware
    - [X] Test Keysight FieldFox
    - [x] Test Keysight PNA
    - [ ] Test NanoVNA
    - [ ] Test HP 8510c
    - [ ] Test Rhode&Schwarz ZVA

PR #855, opened 2023-03-16
make nanovna2 idn function a property

zarath:
idn should be a property as in the abcvna base class

### DIFF
--- a/skrf/vi/vna/nanovna_v2.py
+++ b/skrf/vi/vna/nanovna_v2.py
@@ -64,6 +64,7 @@ from time import sleep
 # 0x21 0x20: number of sweep frequency points (2 bytes; uint16)
 # 0x23 0x22: number of data points to output for each frequency (2 bytes; uint16)
 
+
 class NanoVNAv2(abcvna.VNA):
     """
     Python class for NanoVNA V2 network analyzers [#website]_.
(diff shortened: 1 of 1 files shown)


jhillairet:
There is currently a larger reimplementation of the instruments/vi parts, cf #744. Please have a look to it, as it will be probably the new default in future version (nanoVNA has been implemented).

zarath:
Ok. I'll take a look.

Btw. Do you consider renaming the nanavna part more clearly to nanovna_v2 or saa2? It's a little bit misleading, as the s-a-a-2/v2 has a binary protocol, opposed to the original and more common nanovna's (and -H/-F..)

cafeclimber:
> Ok. I'll take a look.
> 
> Btw. Do you consider renaming the nanavna part more clearly to nanovna_v2 or saa2? It's a little bit misleading, as the s-a-a-2/v2 has a binary protocol, opposed to the original and more common nanovna's (and -H/-F..)

Hi, I'm the one writing the new implementation. In the vi.vna module, everything will be grouped by manufacturer, but you make a good point about changing the name of the specific instrument class. I'll rename it to nanovna.NanoVNAV2

cafeclimber:
Also, if you are a user of the NanoVNA v2, it would be great if you had a chance to test the new implementation! Just clone my fork, install, and try running through the various commands. I don't have one on hand so was writing a lot based on the previous implementation.

zarath:
> Also, if you are a user of the NanoVNA v2, it would be great if you had a chance to test the new implementation! Just clone my fork, install, and try running through the various commands. I don't have one on hand so was writing a lot based on the previous implementation.

Yes, I'll check it! ... I close this PR here now!

zarath:
@cafeclimber oh, I've just seen that the request for testing was from you.... Is there a document, which methods should be mandatory and which could be optional in each driver. I've some concerns making frequency_start, stop, and step methods propertiies. They are harder to derive. despite to "normal" methods. One can discuss if it is better and more pythonic to have set_frequency_start/get_frequency_stop. Most pythonic way in my opinion would be to have just 3 attributes (start, stop, npoints) and and a method(s) to apply that values to the instrument or calculating them from a Frequency object.

cafeclimber:
You can build the docs in my fork and there's a page with expected names. It's still very much in active development though

PR #858, opened 2023-03-20
Citi fix

Nemesis82:
Please use this instead of #854 .

### DIFF
--- a/skrf/io/citi.py
+++ b/skrf/io/citi.py
@@ -199,6 +199,15 @@ class Citi():
                     raise NotImplementedError('Not implemented format case')
 
                 self._data[cur_name]['values'] = values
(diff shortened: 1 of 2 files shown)


jhillairet:
> Please use this instead of #854 .

?

Actually #854 is solving additional issues with CITI files, tests, and the support of the characterics impedances.

Nemesis82:
Then please take this. Unfortunately, I currently don't have time to
further look into this topic. I am using the code from my pull request
since several weeks bit I don't care which one will be taken.

Julien Hillairet ***@***.***> schrieb am Mo., 20. März 2023,
21:03:

jhillairet:
superseded by #854

PR #859, opened 2023-03-23
Create variables to modify nudge_eig thresholds

### LINKED ISSUE #741: nudge_eig function is causing wrong conversion of S->Z
I understand why the **nudge_eig** function exists. but In my case the eigenvalues are borderline and fail the condition as given here 
Max(3.14e-09, 1e-12) and nudges to 3.14e-09. 
This causes wrong results. I can monkey patch this but it would be good if we can have a global variable that sets this limit. 

Amy thoughts 

```
def nudge_eig(mat: npy.ndarray, cond: float = 1e-9, min_eig: float = 1e-12) -> npy.ndarray:
    r"""Nudge eigenvalues with absolute value smaller than
    max(cond * max(eigenvalue), min_eig) to that value.
    Can be used to avoid singularities in solving matrix equations.
    Input should have dimension of similar to (nfreqs, nports, nports).

```

jhillairet:
Fix #741
Create new variables in the `constants` part that can be tuned to modify `nudge_eig` thresholds

### DIFF
--- a/skrf/constants.py
+++ b/skrf/constants.py
@@ -85,6 +85,17 @@ T0 = 290.
 Room temperature (kind of)
 """
 
+EIG_COND = 1e-9
+"""
+Eigenvalue ratio compared to the maximum eigenvalue in :meth:`~skrf.mathFunctions.nudge_eig`.
+
+EIG_COND * max(eigenvalue)
+"""
+
+EIG_MIN = 1e-12
+"""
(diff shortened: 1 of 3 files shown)


jhillairet:
@vishak-venkatraman, could you test this PR and let me know if by changing the following variables:
`rf.constants.EIG_COND` or `rf.constants.EIG_MIN` it solves your issue?

Vinc0110 (approved):

PR #863, opened 2023-03-26
[pre-commit] pre-commit configuration

jhillairet:
Add a minimal configuration file to use with pre-commit. (Will lead to many small trailing spaces and etc in a foreseen distinct PR) 

To be discussed, if we should use additional hooks:
- [black](https://github.com/psf/black) formatting ([blacken-docs](https://github.com/asottile/blacken-docs))?
- [isort](https://github.com/pycqa/isort)?
- [codespell](https://github.com/codespell-project/codespell) ?
- [notebook quality](https://github.com/nbQA-dev/nbQA)?
- other?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,12 @@
+exclude: |
+     (?x)(
+          ^conda.recipe/meta.yaml
+     )
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
     hooks:
-    -   id: trailing-whitespace
-    -   id: end-of-file-fixer
-    -   id: check-added-large-files
+    - id: trailing-whitespace
+    - id: end-of-file-fixer
+    - id: check-added-large-files
+    - id: check-yaml
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-added-large-files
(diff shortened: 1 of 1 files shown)


FranzForstmayr:
+1 for using a code formatter, I'm also using black most of the time.
Instead of isort I would recommend to use ruff, which is a de facto successor of flake8.
ruff has lots of plugins builtin (including isort) and is really fast.

eendebakpt:
+1 for a code formatter. black is nice (with perhaps max-line length increased a bit), although autopep8 is also good
+1 for ruff. Instead of having multiple tools (isort, flake, etc.) one just needs a single tool. 

Here is the configuration file I use
```
exclude: 'doc/conf.py'

repos:

-   repo: https://github.com/psf/black
    rev: 23.1.0
    hooks:
      - id: black

- repo: https://github.com/nbQA-dev/nbQA
  rev: 1.6.3
  hooks:
    - id: nbqa-black
      additional_dependencies: [black==23.1]
    - id: nbqa-ruff
      args: [--fix, --exit-non-zero-on-fix]

- repo: https://github.com/charliermarsh/ruff-pre-commit
  # Ruff version.
  rev: 'v0.0.253'
  hooks:
    - id: ruff
      args: [--fix, --exit-non-zero-on-fix]
```
with in `pyproject.toml`:
```
... 

[tool.ruff]
select = ["E", "F", "UP", "I"]
ignore = [ #"UP031", # use format specifiers instead of percent format
            #"F401", # imported but unused
        ]
line-length = 120

target-version = "py38"
```
To reduce the amount of changes in the codebase, one could remove the "UP" or add more rules to the `ignore`

cafeclimber:
Another +1 for Black and Ruff!

jhillairet:
Hum, I did not intend to merge it. Seems it autocally merged it when I merged #863 (which was branched from from)..

I'll do another PR to add ruff and black

PR #864, opened 2023-03-27
Draft: Embed line if impedances differ

eendebakpt:
@mhuser @jhillairet This is a followup of https://github.com/scikit-rf/scikit-rf/pull/715. The deprecation warning is showing up in some example notebooks (e.g. https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Multiline%20TRL.html) 
It seems there is currently no way to have the desired behaviour, without a warning.

This PR updates the default value of `embed` to `None` with the (I think) intended behaviour . This allows to remove the `embed` option at a later moment.

### DIFF
--- a/doc/source/examples/metrology/Multiline TRL.ipynb
+++ b/doc/source/examples/metrology/Multiline TRL.ipynb
@@ -446,8 +446,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "X = coax1mm.line(1, 'm', z0=58, name='X', embed=True)\n",
-    "Y = coax1mm.line(1.1, 'm', z0=40, name='Y', embed=True)\n",
+    "X = coax1mm.line(1, 'm', z0=58, name='X')\n",
+    "Y = coax1mm.line(1.1, 'm', z0=40, name='Y')\n",
     "\n",
     "plt.figure()\n",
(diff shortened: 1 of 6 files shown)


eendebakpt:
@mhuser @Vinc0110 @jhillairet  To make this PR pass the tests, several of them would need to be modified. For example in `test_coaxial.py` there is a test failing because
```
a_media = Coaxial(
            frequency = qucs_ntwk.frequency,
            Dint=1e-3, Dout=3e-3, epsilon_r=2.29, \
            tan_delta=4e-4, sigma=1./1.68e-8 \
            )
a_media.thru(z0=50)
```
In current master the `a_media.thru(z0=50)` is not embedded (default is is `False`), in this PR the thru is embedded with a different result (since `z0=50` is not equal to the `a_media.z0`). It shows that the PR is not backwards compatible, perhaps more than we would want. Since the `line(..., embed = ...)` is called via `thru` (which is called via several other functions), it is hard for users to get rid of the warnings with the current master)

 Can you advice on how to more forward? The current deprecation warning has been in place for 9 months, which is not a very long time.

eendebakpt:
@mhuser Thanks for the pointers. I will close this PR in favor of #727.

Issue #865, opened 2023-03-28
Issue with TRL and determine_reflect


eendebakpt:
The TRL calibration procedure is giving some unexpected (incorrect?) results. I tracking it down to the following minimal example.
I define a perfect thru, reflect and line and then run `determine_reflect`. I would expect it to return a reasonable estimate for the reflect, but instead I get `NaN` values.
```
import skrf
from skrf.media import CPW, Coaxial
import numpy as np
import matplotlib.pyplot as plt
import matplotlib.pyplot as plt
from skrf.calibration import determine_reflect
np.set_printoptions(threshold=8)

# define medium
freq = skrf.F(.5,.7,801, unit='GHz')
att_db_per_meter = 3.0
medium=Coaxial.from_attenuation_VF(freq, att=att_db_per_meter, VF=.69)

# define measurements
thru = medium.line(0, 'm')
line=medium.line(0.12, 'm')
short = medium.short()
short2port = skrf.two_port_reflect(short, short)
    
# check our frequency is in the valid range for TRL
w = np.rad2deg(np.imag( medium.electrical_length(.12)))
print(f'TRL bandwidth {w[0], w[-1]}')


rr=determine_reflect(thru, short2port, line, short2port, line, return_all=True)

for r in rr:
    print(r.s)
    
# Do not give any hints
rr=determine_reflect(thru, short2port, line, return_all=True)
for r in rr:
    print(r.s)
````

**Comment by Ttl** [comment]
The issue is caused by division by zero when thru and line are both perfectly matched. You can work around it in this case by adding a tiny amount of mismatch:

```python
thru.s[:,0,0] += 1e-7
thru.s[:,1,1] += 1e-7
```

eendebakpt:
@Ttl Adding the small mismatch indeed allows the run the algorithm with good results. Often when near singularities algorithms become a bit unstable, but that does not appear to be the case here. I tested with 
```
thru.s[:,0,0] += 1e-7
thru.s[:,1,1] += 1e-7
thru.add_noise_polar(1e-7, 1e-7)
```
and that gives good results as well.

Do you have any insight why the division by zero occurs for matched thru and line? I seems to me it should be an easy case, and I am a bit wary of algorithms that have issues with easy cases.

[...]

PR #868, opened 2023-03-29
Fix division by zero warnings in determine_line

eendebakpt:
* Fix division by zero warnings in `determine_line`
* Add documentation for `determine_reflect` (see https://github.com/scikit-rf/scikit-rf/issues/865)
* Remove unused variables

The division by zero warnings are caused by the elementwise division of networks S parameters in `line_m/thru_m`, where only the result for S21 is required.


jhillairet:
Thank you!

PR #871, opened 2023-04-01
improve performance of Coaxial.from_attenuation

eendebakpt:
The `Coaxial.from_attenuation`  check on input sizes converted the `Frequency`  argument to an array with a very slow loop including a call to `__getattr__` on each element. We replace the conversion with a lookup of `Frequency.f` 

Benchmark:
```
freq = skrf.F(.15,1.7,801, unit='GHz') # fails   
att_db_per_meter = 3.0
medium=Coaxial.from_attenuation_VF(freq, att=att_db_per_meter, VF=.69)
```
Master: 3.5 seconds, this PR: < 0.1 seconds.

@FranzForstmayr

### DIFF
--- a/skrf/media/coaxial.py
+++ b/skrf/media/coaxial.py
@@ -121,7 +121,7 @@ class Coaxial(DistributedCircuit, Media):
 
         """
         # test size of parameters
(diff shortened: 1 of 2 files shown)


jhillairet (approved):
Nice catch!

PR #875, opened 2023-04-04
Improve performance network.copy

eendebakpt:
The `network.copy` makes a deep copy of the frequency object, and this copy happens again in the `Network`  constructor.
This PR removes the explicit copy of the frequency object and adds a test.

Note: are the copies at https://github.com/scikit-rf/scikit-rf/blob/master/skrf/calibration/calibration.py#L245 required? If not, we could remove them as well

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1901,7 +1901,7 @@ class Network:
 
         """
         ntwk = Network(s=self.s,
-                       frequency=self.frequency.copy(),
+                       frequency=self.frequency,
                        z0=self.z0, s_def=self.s_def,
                        comments=self.comments, params=self.params
                        )
@@ -1909,9 +1909,6 @@ class Network:
         ntwk.name = self.name
 
         if self.noise is not None and self.noise_freq is not None:
-          if False :
-              ntwk.noise = npy.copy(self.noise)
-              ntwk.noise_freq = npy.copy(self.noise_freq)
           ntwk.noise = self.noise.copy()
           ntwk.noise_freq = self.noise_freq.copy()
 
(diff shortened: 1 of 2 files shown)


jhillairet:
I guess that, originally, the `copy()` was made to avoid modifying the frequency objects of the copies if the original one was modified... Can you check it is not the case after the change?

eendebakpt:
> I guess that, originally, the `copy()` was made to avoid modifying the frequency objects of the copies if the original one was modified... Can you check it is not the case after the change?

I added test cased to cover this

FranzForstmayr:
Just thinking loud....
Do we still need copy(), as a frequency object should be immutable?
I know that we intended to change the frequency object to be immutable, but I'm not sure if we checked every case.

jhillairet:
> Just thinking loud.... Do we still need copy(), as a frequency object should be immutable? I know that we intended to change the frequency object to be immutable, but I'm not sure if we checked every case.

You're right, was supposed to be done in #700. But it seems that changing an element of the array is still possible, as done in the test of this PR.

eendebakpt:
> Just thinking loud.... Do we still need copy(), as a frequency object should be immutable? I know that we intended to change the frequency object to be immutable, but I'm not sure if we checked every case.

The `Frequency` object is not immutable right now. One can change the unit and also the frequency (via the `Frequency.f` numpy array). There are three setter methods, of which 2 are already deprecated. I created https://github.com/scikit-rf/scikit-rf/pull/878 to see whether we can actually make it immutable (except for the unit).

eendebakpt:
@FranzForstmayr @jhillairet I think this PR is ready. The Frequency is not immutable right now, and will not be during some deprecation time.

jhillairet (approved):

PR #881, opened 2023-04-06
URL updated.

wollmich:


jhillairet (approved):

jhillairet:
Thank you @wollmich. 

By the way these IO routines related to METAS probably need some updates. Let us know how scikit-rf could be useful to METAS products.

PR #885, opened 2023-04-08
Fix incorrect z0 handling in SeriesImpedance and ShuntAdmittance

Ttl:
Reported on the mailing list. `z0` was not set in `a2s`.


jhillairet (approved):
I started working on it, but you have been faster then me, thanks!

Issue #886, opened 2023-04-10
Network.copy() does not copy the params dictionary.


cafeclimber:
Calling Network.copy() does not call copy on the params dictionary. This causes issues when trying to copy and then mutate the copy's params.

**Example**

```python
from skrf.data import ring_slot
a = ring_slot.copy()
a.params = {'foo':'bar'}
b = a.copy()
print(f"{id(a)=} {id(a.params)=}")
# id(a)=139976964164768 id(a.params)=139976964060544 <------
print(f"{id(b)=} {id(b.params)=}")
# id(b)=139976964163136 id(b.params)=139976964060544 <------
```

**Expected behavior**
params should also be copied

(I'll try to submit a fix asap)

PR #887, opened 2023-04-10
Make Network.copy also copy params dict

cafeclimber:
Currently, `Network.copy()` does not copy `params` which was extremely confusing in my case. `copy` should be a full deep copy. This PR makes that the case.


cafeclimber:
Addresses #886

jhillairet (approved):
Indeed, good catch

PR #888, opened 2023-04-10
Prepare media tutorial for characteristic impedance implementation change

mhuser:
Update the `Media` tutorial to show how the port and characteristic impedance of `Media` should be handled and describe use cases with figures.


denzchoe:
wow these illustrations are impressive man...

doc/source/tutorials/Media.ipynb
jhillairet, line 1: Great tutorial! And super nice figures!

Just a couple of small comments:

1. Ghz -> GHz in:

```
f_mlin = Frequency(0.1, 10,1001, 'Ghz')
f_wr10 = Frequency(75, 110, 1001,'Ghz')
```


2. for the sake of the tutorial, add `unit=` in 
```
mlin_100 = mlin.line(100e-3, unit='m', name = 'mlin 100mm')
wr10_100 = wr10.line(100e-3, unit='m', name = 'wr10 100mm')
```

3. in:
> This is because no *port impedance* `z0` (lowercase) was specified at media construction. The *characteristic impedance* `Z0` (upper case) was used as port impedance and the network is perfectly matched with itself. This result is equivalent to an electromagnetic simulation with complex *port impedance* instead of a real-world measurement.

I would say something like:

>This is because no *port impedance* `z0` (lowercase) was specified at media construction. In this cas, the *characteristic impedance* `Z0` (upper case) was used by default as port impedance and the network is perfectly matched with itself on the whole frequency band, as the characteristic impedance is a (complex) frequency dependant parameter. This behaviour also arises in electromagnetic simulations with complex *port impedance* instead of a real-world measurement.
>
>print(mlin_100.z0)  # the port impedance is frequency dependant

I'm not sure the remark concerning the complex port impedance is clear for everyone.

4. print(wr10_200_measured1) -> print(wr10_100_measured1) ?

jhillairet: 5. ohm -> Ohm
6. "The measurement will store the port impedance instead of the characteristic impedance which is lost. This is not normalization. The actual characteristic impedance of the line is not measured. This method is specific to hollow waveguides." -> yes, and measuring the char impedance of a rectangular waveguide has no meaning when using multimode waveguides.

jhillairet (commented):
Great tutorial.

I wonder if we should drop the `embed` deprecation from the next release.

doc/source/tutorials/Media.ipynb
mhuser: Thank you a lot for your review @jhillairet !

1. Oh well spotted! I probably wrote 'ghz' first and did only correct it halfway. Corrected.
2. Added `unit = ` in `.line(...)` in whole notebook.
3. Proposition without mention of complex numbers:
This is because no *port impedance* `z0` (lowercase) was specified at media construction. In this case, the *characteristic impedance* `Z0` (upper case) is used as *port impedance* and the network is perfectly matched with itself on the whole frequency band.
The *characteristic impedance* is usually a frequency-dependent parameter. Having a frequency-dependent *port impedance* is often encountered in electromagnetic simulations. Real-world measurements use a constant *port impedance* instead.
4. Indeed!
5. ohm->Ohm in the whole notebook.
6. Thank you for your confirmation. I was not really sure here because I just never used a rectangular waveguide!

jhillairet (approved):

PR #892, opened 2023-04-12
[pre-commit] ruff

jhillairet:
Adding ruff to the pre-commit configuration and fixes the found issues for each rules


jhillairet:
Tests are passing but there is problem with coveralls for some reasons (cf https://github.com/lemurheavy/coveralls-public/issues/1708).

Most of rules are now checked, except:

- F403: undefined-local-with-import-star
- F405: undefined-local-with-import-star-usage
- F841: unused-variable
- E741: ambiguous-variable-name

which require more work.

mhuser (approved):
Looks great! Lot of syntax ehancement done here.

Issue #889, opened 2023-04-11
TRL calibration fails with ideals argument


eendebakpt:
**Describe the bug**

The TRL calibration fails when ideals are provided. This is counterintuitive, since providing more information should help the algorithm.

**To Reproduce**
The following minimal example shows the behaviour.

```python
import matplotlib.pyplot as plt

import numpy as np
import skrf as rf
from skrf import Network, two_port_reflect
from skrf.media import Coaxial
from skrf.calibration import TRL

def add_noise(networks: list[Network], mag_dev : float, phase_dev : float) -> list[Network]:
    """  Add noise to list of networks """
    result = [n.copy() for n in networks]
    [n.add_noise_polar(mag_dev, phase_dev) for n in result]
    return result

#%% Media and components
frequency = rf.Frequency(1,10,1001,'ghz')
att_db_per_meter = 4*frequency.f/np.max(frequency.f) # data roughly estimated from #https://www.minicircuits.com/pdfs/086-10SBSM+.pdf
media=Coaxial.from_attenuation_VF(frequency, att=att_db_per_meter, VF=.69)
thru = media.line(d=2, unit='cm', name='thru', z0=50+1e-6)
line = media.line(d=12, unit='cm', name='line')

reflective_standards = {'short': two_port_reflect(media.short(), media.short(), name='short'),
                       'open': two_port_reflect(media.open(), media.open(), name='open'),
                       'load': two_port_reflect(media.load(Gamma0=0), media.load(Gamma0=0), name='load')}

fixture_left=media.attenuator(s21=-2, dB=True)
fixture_right=media.attenuator(s21=-2, dB=True)

length_thru=.1
thru = media.line(d=length_thru, unit='m', name='thru') # good at small d

delta_line=1.71*1e-2
line = media.line(d=delta_line, unit='m',name='line', z0=50. +1e-5) ** thru
ll=[length_thru, length_thru+0.171]

dut = media.line(d=5, unit='cm', name='DUT')

eline = media.electrical_length(delta_line)
el_deg = np.rad2deg(np.imag(eline))   
trl_singularity=frequency.f[np.argmin(el_deg<180)]

#%% Perform TRL calibration
my_ideals=[thru,reflective_standards['short'],  line]
my_measured=[fixture_left ** n ** fixture_right for n in my_ideals]
my_measured=add_noise(my_measured, mag_dev=1e-4, phase_dev=.001)

[...]

PR #895, opened 2023-04-13
Allow NISTMultilineTRL to work with non-exact floats

eendebakpt:
The `NISTMultilineTRL` checks whether variables are scalar with the construction
``` 
try:
   v = x[0]
except TypeError:
    # treat x as scalar
    ...
```
This works for float, since `v = 1.0; v[0]` results in a `TypeError`. But for `v = npy.float64(1.0); v[0]` an `IndexError` is raised. This PR allows `NISTMultilineTRL` to work with numpy scalars by checking arguments with `numpy.isscalar`  and converting if required. The PR also improves performance of `s2t`  by factoring out a common expression.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -2638,7 +2638,7 @@ class NISTMultilineTRL(EightTerm):
         self.refl_offset = refl_offset
         self.ref_plane = ref_plane
         self.er_est = er_est
-        self.l = l
+        self.l = [float(v) for v in l] # cast to float, see gh-
         self.Grefls = Grefls
         self.gamma_root_choice = gamma_root_choice
         self.k_method = k_method
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -2638,7 +2638,7 @@ class NISTMultilineTRL(EightTerm):
         self.refl_offset = refl_offset
         self.ref_plane = ref_plane
         self.er_est = er_est
-        self.l = [float(v) for v in l] # cast to float, see gh-
+        self.l = [float(v) for v in l] # cast to float, see gh-895
         self.Grefls = Grefls
         self.gamma_root_choice = gamma_root_choice
         self.k_method = k_method
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -3234,9 +3234,7 @@ class NISTMultilineTRL(EightTerm):
         #Reference impedance renormalization
         if self.z0_ref is not None and npy.any(z0 != self.z0_ref):
             powerwave = self.kwargs.get('powerwave', False)
-            if npy.shape(self.z0_ref) != (fpoints,):
-                z0_ref = self.z0_ref*npy.ones(fpoints, dtype=complex)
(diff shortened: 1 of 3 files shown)


Ttl (changes requested):
I would prefer making them arrays in `__init__` so that try except is not necessary. `npy.isscalar` could be useful for it as it handles both numpy types and ordinary python floats.

Just adding a comment is not good enough fix for #889. I'll submit a PR for it later.

skrf/calibration/calibration.py
```diff
@@ -2636,28 +2637,30 @@ def __init__(self, measured, Grefls, l,
 
         """
         self.refl_offset = refl_offset
+
+        if npy.isscalar(ref_plane):
+                ref_plane = [ref_plane, ref_plane]
```
Ttl, line 2642: Extra indentation.

skrf/calibration/calibration.py
```diff
@@ -3163,20 +3158,20 @@ def solve_A(B1, B2, CoA1, CoA2):
                     raise ValueError('Only one of c0 or z0_line can be given.')
                 try:
                     c0m = self.c0[m]
-                except TypeError:
+                except (TypeError, IndexError):
```
Ttl, line 3161: Could this also be changed to use `isscalar` instead of `try ... else`? Either at `__init__` set it to array or `None` or check it here.

skrf/calibration/calibration.py
```diff
@@ -3163,20 +3158,20 @@ def solve_A(B1, B2, CoA1, CoA2):
...
                     c0m = self.c0
                 z0[m] = gamma[m]/(1j*2*npy.pi*freqs[m]*c0m)
             else:
                 #Set the known line characteristic impedance
                 if self.z0_line is not None:
                     try:
                         z0[m] = self.z0_line[m]
-                    except TypeError:
+                    except (TypeError, IndexError):
```
Ttl, line 3169: Same here too

skrf/calibration/tests/test_calibration.py
```diff
@@ -761,6 +761,29 @@ def test_shift(self):
             self.assertTrue(all(npy.abs(self.cal.coefs[k] - self.cal_shift.coefs[k]) < 1e-9))
 
 
+    def test_non_exact_float(self):
```
Ttl, line 764: `test_numpy_float` sounds better.

skrf/calibration/tests/test_calibration.py
```diff
@@ -761,6 +761,29 @@ def test_shift(self):
...
+        # see gh-895
+        cal= NISTMultilineTRL(
```
Ttl, line 766: Spaces around equals sign and the same below.

Ttl (commented):
Looks mostly fine, just few comments.

skrf/calibration/calibration.py
eendebakpt: I decided to convert scalar values to a list in the `__init__`. It takes a bit extra memory, but I think it is rare for the frequency to be so large this causes problems.

Note that the `z0_line`, etc. are public attributes so this slightly changes the public behavior of the TRL.

Ttl (approved):

PR #897, opened 2023-04-19
Fix doc links from Circuit tutorial to Circuit convenience methods

caryan:
Currently they all link to `Circuit.Ground`


jhillairet (approved):

jhillairet:
Indeed... Thank you @caryan and welcome as a contributor to scikit-rf!

PR #898, opened 2023-04-23
Create dependabot.yml

jhillairet:
Dependabot to automatically keep the dependencies and packages used in your repository updated to the latest version

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+
+# Maintain dependencies for GitHub Actions
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
+  assignees:
+  - jhillairet
+
+# Maintain dependencies for python
+- package-ecosystem: pip
+  directory: /
+  schedule:
+    interval: monthly
+  assignees:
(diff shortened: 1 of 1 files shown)


FranzForstmayr:
I'm not really sure how this should work.

For example we use `numpy >= 1.21`, the current version is `1.24`. Will dependabot automatically increase the minimum required numpy if allowed by the other requirements? If that's the case I'm against, as we would disallow users of older numpy versions to use `skrf` without any technical reason. 

I'm perfectly fine to update a package if needed from our side.

jhillairet:
> For example we use `numpy >= 1.21`, the current version is `1.24`. Will dependabot automatically increase the minimum required numpy if allowed by the other requirements? If that's the case I'm against, as we would disallow users of older numpy versions to use `skrf` without any technical reason.

Yes, I agree with you. We are making a Python package, not a webapp in javascript, so we don't require to be on the cutting edge of the required packages.

There is a[ `version-strategy: increase-if-necessary` option](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) I should include in the config file,  which increases the version requirement only when required by the new version. To be tested.

Issue #860, opened 2023-03-24
Time scale in time domain


wollmich:
I think the following assumption is wrong:

https://github.com/scikit-rf/scikit-rf/blob/cc00e879f615c09a5f84b17355b8362e19c5ae52/skrf/time.py#L410

In the band pass mode the time resolution, maximum time and minimum time are given by:

```
dt_bp = 1 / (n * df)

tp_bp = floor((n - 1) / 2)  * dt_bp
-tn_bp = -floor(n / 2) * dt_bp

tp_bp + tn_bp = (n – 1) / (n * df)
```

In the low pass mode, when the frequency domain data do not include a DC point, the time resolution, maximum time and minimum time are given by:

```
dt_lp = 1 / (2 * n + 1) / df

tp_lp = n * dt_lp
-tn_lp = -n* dt_lp

tp_lp + tn_lp = 2 * n / (2 * n + 1) / df
```

In the low pass mode, when the frequency domain data include a DC point, the time resolution, maximum time and minimum time are given by:

```
dt_lp = 1 / (2 * n - 1) / df

tp_lp = (n - 1) * dt_lp
-tn_lp = -(n - 1) * dt_lp
```

So the anti-alias-time is little less than `1/df`.

Vinc0110:
Could you explain this some more? What exactly is the issue? Did you run into problems with the existing code? Did the time-gating fail in your tests?
You claim the time scale is wrong (it might be), but your list of equations don't seem to proof it. By the way: the time-domain plot created with `Network.plot_s_db_time()` is completely independent from the `time_gate()`.

For further discussion, I propose to treat the low-pass transform without a dc point as a (border line) case of a band-pass transform. Then, we only have to deal with those two cases:

### Band-pass mode:
The frequency-domain signal is treated as a band-pass signal including frequency samples from $f_{min}$ to $f_{max}$. 
For time-gating purposes, the signal is transformed into time-domain using an [inverse FFT](https://numpy.org/doc/stable/reference/generated/numpy.fft.ifft.html), which yields a complex-valued time-domain signal. After gating (multiplication with the time-gate), it gets transformed back into frequency domain using a [FFT](https://numpy.org/doc/stable/reference/generated/numpy.fft.fft.html).
The length of the time-domain signal equals the length of the frequency domain signal: $N_{time} = N_{freq}$.

### Low-pass mode:
[...]

Issue #757, opened 2022-09-22
Time-domain gating: Frequency-domain window required?


Vinc0110:
This is a branch from the discussion on the mailing list: https://groups.google.com/g/scikit-rf/c/IoH8hNV58sI

Possibly related to issue #734.

The problem is with the roll-off in the edges of the frequency-domain results after gating (at f_min and f_max). A perfectly straight line is provided by PNA transforms. They probably apply a frequency-domain window before the gating. 

Discussion: Is this required or advised to do by default?

The theoretical background is covered in an application note by Agilent/Keysight: 
- [new version; login required](https://www.keysight.com/de/de/assets/7018-01451/application-notes/5989-5723.pdf)
- [old version; no login required](https://www.eeweb.com/wp-content/uploads/articles-app-notes-files-time-domain-analysis-using-a-network-analyzer-1304976108.pdf)

Example code based on the data by Roger in the mailing list:

```python
from skrf.media.coaxial import Coaxial
import skrf
import matplotlib.pyplot as plt
import numpy as np

# reflection delays in nanoseconds
t1 = 758.725e-6
t2 = 82.759e-3
t3 = 416.759e-3
t4 = 508.759e-3

# window settings
freq = skrf.Frequency(0.05, 33, 1000, 'GHz')
print(freq.f)
window_type = ('tukey', 0.5)
t_center = t2
t_span = 0.25   # 150 ps

# create network
media = Coaxial(frequency=freq, z0=50)
coax_line_1 = media.line(0.5 * t2, 'ns', z0=50, embed=True)
coax_line_2 = media.line(0.5 * (t3 - t2), 'ns', z0=25, embed=True)
coax_line_3 = media.line(0.5 * t2, 'ns', z0=50, embed=True)
nw_total = coax_line_1 ** coax_line_2 ** coax_line_3

# apply gate
nw_gated = skrf.time_gate(nw_total.s11, window=window_type, center=t_center, span=t_span)

# plotting
fig, ax = plt.subplots(3, 1)

nw_total.s11.plot_s_time(ax=ax[0], label='Original')
nw_gated.s11.plot_s_time(ax=ax[0], label='Gated')
ax[0].set_xlim(-0.5, 1.5)
ax[0].legend(loc='upper right')

nw_total.s11.plot_z_time_step(ax=ax[1], label='Original')
nw_gated.s11.plot_z_time_step(ax=ax[1], label='Gated')
ax[1].set_xlim(-0.5, 1.5)

[...]

PR #900, opened 2023-04-24
Timescale

FranzForstmayr:
This PR fixes several issues regarding time transformation.

to be continued

### DIFF
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -628,16 +628,16 @@ class Frequency:
         """
         Time vector in s.
 
-        t_period = 1/f_step
+        t_period = (n-1)/f_step
         """
-        return linspace(-.5/self.step , .5/self.step, self.npoints)
+        return npy.fft.fftshift(npy.fft.fftfreq(self.npoints, self.step))
 
     @property
     def t_ns(self) -> npy.ndarray:
         """
         Time vector in ns.
 
-        t_period = 1/f_step
+        t_period = (n-1)/f_step
         """
         return self.t*1e9
 
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -628,7 +628,7 @@ class Frequency:
         """
         Time vector in s.
 
-        t_period = (n-1)/f_step
+        t_period = 2*(n-1)/f_step
         """
         return npy.fft.fftshift(npy.fft.fftfreq(self.npoints, self.step))
 
@@ -637,7 +637,7 @@ class Frequency:
         """
         Time vector in ns.
 
-        t_period = (n-1)/f_step
+        t_period = 2*(n-1)/f_step
         """
         return self.t*1e9
 
--- a/skrf/mathFunctions.py
+++ b/skrf/mathFunctions.py
@@ -1054,7 +1054,7 @@ def ifft(x: npy.ndarray) -> npy.ndarray:
     --------
     irfft
     """
-    return npy.fft.fftshift(npy.fft.ifft(x, axis=0), axes=0)
+    return npy.fft.fftshift(npy.fft.ifft(npy.fft.ifftshift(x, axes=0), axis=0), axes=0)
 
 
 def irfft(x: npy.ndarray, n:int = None) -> npy.ndarray:
--- a/skrf/plotting.py
+++ b/skrf/plotting.py
@@ -1425,8 +1425,8 @@ def plot_reciprocity2(self, db=False, *args, **kwargs):
     plt.draw()
 
 
-def plot_s_db_time(self, window: Union[str, float, Tuple[str, float]]=('kaiser', 6),
-        normalize: bool = True, center_to_dc: bool = None, *args, **kwargs):
+def plot_s_db_time(self, *args, window: Union[str, float, Tuple[str, float]]=('kaiser', 6),
+        normalize: bool = True, center_to_dc: bool = None, **kwargs):
     return self.windowed(window, normalize, center_to_dc).plot_s_time_db(*args,**kwargs)
 
 
--- a/skrf/plotting.py
+++ b/skrf/plotting.py
@@ -1425,8 +1425,9 @@ def plot_reciprocity2(self, db=False, *args, **kwargs):
     plt.draw()
 
 
-def plot_s_db_time(self,center_to_dc=None,*args,**kwargs):
-    return self.windowed(center_to_dc=center_to_dc).plot_s_time_db(*args,**kwargs)
+def plot_s_db_time(self, window: Union[str, float, Tuple[str, float]]=('kaiser', 6),
+        normalize: bool = True, center_to_dc: bool = None, *args, **kwargs):
+    return self.windowed(window, normalize, center_to_dc).plot_s_time_db(*args,**kwargs)
 
 
 # plotting
--- a/skrf/time.py
+++ b/skrf/time.py
@@ -407,19 +407,18 @@ def time_gate(ntwk: 'Network', start: float = None, stop: float = None, center: 
     ntwk_gated.s[:, 0, 0] = ntwk_gated.s[:, 0, 0] * window_fd
 
     # create time vector
-    t = npy.linspace(-0.5 / df, 0.5 / df, n_td)
-
+    t = npy.fft.fftshift(npy.fft.fftfreq(n_td, df))
     # find start/stop gate indices
     start_idx = find_nearest_index(t, start)
     stop_idx = find_nearest_index(t, stop)
 
     # create gating window
-    window_width = abs(stop_idx - start_idx)
+    window_width = abs(stop_idx - start_idx) + 1
     window = signal.get_window(window, window_width)
 
     # create the gate by padding the window with zeros
     gate = npy.zeros_like(t)
-    gate[start_idx:stop_idx] = window
+    gate[start_idx:stop_idx+1] = window
 
     if method == 'convolution':
         # frequency-domain gating


FranzForstmayr:
Some ongoing work to match the `t` vector from `impulse_response` (which looks good) with `Frequency.t`

mhuser:
> Some ongoing work to match the `t` vector from `impulse_response` (which looks good) with `Frequency.t`

Beware pitfall with [fft](https://numpy.org/doc/stable/reference/generated/numpy.fft.fftfreq.html) vs [rfft](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfftfreq.html). They do not return the same number of samples. I often stumble inside 😉

FranzForstmayr:
I am aware of the computational difference, but both methods expect the first sample to be the DC component. 
In a band pass measurement I'm not sure if the interpretation of time signals make any physical sense without DC extrapolation.
Correct me if I'm wrong.

FranzForstmayr:
Copied from #860 

```python
# Current behaviour
f = rf.Frequency(10, 50, 5)
f.t 
array([-5.0e-11, -2.5e-11,  0.0e+00,  2.5e-11,  5.0e-11]) # 25 ps would imply 40 GSps = 20GHz

# PR 900 now
f = rf.Frequency(10, 50, 5)
f.t 
array([-4.0e-11, -2.0e-11,  0.0e+00,  2.0e-11,  4.0e-11]) # 20 ps would imply 50 GSps = 25 GHz

# I would like to have when PR900 is finished
f = rf.Frequency(10, 50, 5)
f.t 
array([-5.0e-11, -4.0e-11, -3.0e-11, -2.0e-11, -1.0e-11 0.0e+00,  1.0e-11, 2.0e-11, 3.0e-11, 4.0e-11]) # 10 ps would imply 100 GSps = 100 GHz
```

mhuser:
> In a band pass measurement I'm not sure if the interpretation of time signals make any physical sense without DC extrapolation.

According to the Agilent note presented by @Vinc0110  in #757, the bandpass mode without dc still show location and amplitude of discontinuities in impulse response, but not their capacitive or inductive nature. The frequency requirement are less stringent, e.g. no harmonic sweep required. 
 
image

FranzForstmayr:
Can you calculate this with an ordinary `ifft`? I think one would need something like [STFT](https://en.wikipedia.org/wiki/Short-time_Fourier_transform) to calculatate arbitrary ranges. We do not pass the frequency vector to the `ifft` currently, so we assume frequency is starting at 0.

FranzForstmayr:
> But in the above code to plot the time domain of s-parameters, which should be equivalent to bandpass, it is like half of the frequency vector will be considered as negative frequencies, which left me confused. It probably works, but how?

I'm also not sure how to interprete the result.
> In a band pass measurement I'm not sure if the interpretation of time signals make any physical sense without DC extrapolation.

The current implementation does not care if I pass 0-5GHz, 1-6GHz or 50-55GHz, which seems strange to me.

mhuser:
> > But in the above code to plot the time domain of s-parameters, which should be equivalent to bandpass, it is like half of the frequency vector will be considered as negative frequencies, which left me confused. It probably works, but how?
> 
> I'm also not sure how to interprete the result.

Ok, in [this source](https://www.researchgate.net/publication/261779218) I think I found half of the answer:
Du to the periodic nature of fft (that's also why we can shift it), it can also be seen as positive-only frequency vector only from 0 to Fs. This would be consistant with skrf implementation.
image

But it does not answer our question on how to interpret the bandpass nature of the frequency vector when it start far away 0.

Vinc0110:
Hi guys.
I'm loving the discussion, but I currently don't have time to dig into the details. Hopefully I can have a closer look over the weekend. 

This whole uncertainty is exactly what irritated me in issue #860. A benchmark result from another tool - known to be correct - would be perfect to compare our implementation against.

mhuser:
> This whole uncertainty is exactly what irritated me in issue #860. A benchmark result from another tool - known to be correct - would be perfect to compare our implementation against.

Maybe a brave VNA measurer could measure a length + short and save .s1p as well as bandpass time serie?

In the meantime, I have found that bandpass frequency data seems to be related to *downsampling* if the fft coefficients are 0 between 0 and a start frequency. The source [Oppenheim, Schafer, Discrete-Time Signal Processing](https://research.iaun.ac.ir/pd/naghsh/pdfs/UploadFile_2230.pdf) is often cited and contains following figure:
image

Unfortunately, I did not found in the book neither how the time vector is reconstructed if the inverse fft is done nor how the time data are supposed to looks like.

FranzForstmayr:
We could also make a separate Issue/PR for the bandpass stuff and finish this one.
This PR fixes a one off error for the window, the sampling frequency for baseband measured time visualizations and passes the window specification through `plot_s_db_time` for now.

mhuser:
I think I have got it. I asked a colleague that is much more skilled in signal processing than me and he gave me the interpretation that was later confirmed by the appendix B of [Agilent AN 1287-8 Simplified Filter Tuning Using Time Domain](https://xdevs.com/doc/HP/pub/5968-5328E.pdf)

Radio signal transmission - e.g. AM - modulates a carrier frequency with data, which creates sidebands around it. For demodulation, the time-domain data can be multiplied by the carrier frequency, which centers the side bands around 0 Hz and then low-passed to remove the higher-frequencies content.
Bandpass TDR just does that, but because we already are in the frequency domain we just have to pretend that the center frequency of the band is 0 Hz *et voilà*, demodulation done. Because the spectrum is not symmetric, this will give complex data representing the demodulated [I and Q signals](https://en.wikipedia.org/wiki/In-phase_and_quadrature_components). Hence, to locate discontinuities, magnitude should be observed.

> In this mode, the center frequency of the frequency
> sweep is effectively translated to DC, and the
> inverse Fourier transform is applied from minus
> one-half of the frequency span to plus one-half of
> the span. This is important when looking at a band-
> pass filter with a frequency response that is the
> same as a low-pass filter response translated up in
> frequency to the center of the bandpass filter.
> The time-domain transform represents the return
> loss as a function of length through the device
> under test. For time-domain transforms to be
> useful, they must have enough resolution to resolve
> the distinguishing characteristics of the network
> being measured. In general, the resolution of a
> transform is inversely proportional to the frequency
> span, although in bandpass mode the resolution
> is reduced by half because half the span is for neg-
> ative frequencies and half for positive frequencies.

mhuser:
> We could also make a separate Issue/PR for the bandpass stuff and finish this one.

Indeed. Maybe should we still wait Monday to let @Vinc0110 have a look if he meet the occasion.

mhuser:
```Python
import skrf as rf
import numpy as np

f0 = 10e9
f1 = 50e9
n = 5
f = rf.Frequency(f0, f1, n, unit = 'Hz')

# bandpass frequency and time step
df = (f1 - f0) / (n - 1)
dt = 1 / df

# fft frequency vector
ff1 = np.array([-2, -1, 0, 1, 2]) * df / 5
print(f'{ff1} [Hz]')
ff2 = np.fft.fftshift(np.fft.fftfreq(n, 1/df))
print(f'{ff2} [Hz]')

# fft time vector
tt1 = np.array([-2, -1, 0, 1, 2]) * dt / 5
print(f'{tt1} [s]')
tt2 = np.fft.fftshift(np.fft.fftfreq(n, 1/dt))
print(f'{tt2} [s]')

# output:
[-4.e+09 -2.e+09  0.e+00  2.e+09  4.e+09] [Hz]
[-4.e+09 -2.e+09  0.e+00  2.e+09  4.e+09] [Hz]
[-4.e-11 -2.e-11  0.e+00  2.e-11  4.e-11] [s]
[-4.e-11 -2.e-11  0.e+00  2.e-11  4.e-11] [s]
```

FranzForstmayr:
Ok, thank you for this nice paper :)

But then I think that we serve the values for `ifft` in a wrong order.

From numpy's doc:

>The input should be ordered in the same way as is returned by [fft](https://numpy.org/doc/stable/reference/routines.fft.html#module-numpy.fft), i.e.,
>a[0] should contain the zero frequency term,
>a[1:n//2] should contain the positive-frequency terms,
>a[n//2 + 1:] should contain the negative-frequency terms, in increasing order starting from the most negative frequency.

We calculate the ìfft`like this:
```
return npy.fft.fftshift(npy.fft.ifft(x, axis=0), axes=0)
```
where `x[0]` is our lowest frequency, which is the most negative one. So I think we have to `ifftshift` the values first.

skrf/frequency.py
```diff
@@ -628,16 +628,16 @@ def t(self) -> npy.ndarray:
         """
         Time vector in s.
 
-        t_period = 1/f_step
+        t_period = (n-1)/f_step
```
mhuser, line 631: `t_period = 2*(n-1)/f_step` because half of the frequency vector is considered like the negative frequencies

skrf/frequency.py
```diff
@@ -628,16 +628,16 @@ def t(self) -> npy.ndarray:
...
-        return linspace(-.5/self.step , .5/self.step, self.npoints)
+        return npy.fft.fftshift(npy.fft.fftfreq(self.npoints, self.step))
 
     @property
     def t_ns(self) -> npy.ndarray:
         """
         Time vector in ns.
 
-        t_period = 1/f_step
+        t_period = (n-1)/f_step
```
mhuser, line 640: `t_period = 2*(n-1)/f_step` because half of the frequency vector is considered like the negative frequencies

mhuser (changes requested):
Best if my understanding would be confirmed by someone else.

mhuser:
> But then I think that we serve the values for `ifft` in a wrong order.

Oh! I think you are right! Nice spotted!

mhuser (approved):

mhuser (changes requested):
image

Plot of s-parameter n,m is broken.

FranzForstmayr:
> @FranzForstmayr Can you try this? (edit: plot without dc extrapolation for bandpass)
> 
> ```python
> import skrf as rf
> from skrf.media import Freespace
> from matplotlib import pyplot as plt
> rf.stylely()
> 
> f = rf.Frequency(10, 50, 5, unit = 'GHz')
> m = Freespace(f, z0 = 50)
> n = m.delay_short(10, unit = 'ps')
> n_dc = n.extrapolate_to_dc()
> window = 'boxcar'
> plt.figure()
> plt.title('The 10ps short network')
> n_dc.plot_s_time_impulse(window = window, pad = 0, label = 'plot_s_time_impulse')
> n_dc.plot_s_time_step(window = window, pad = 0, label = 'plot_s_time_step')
> n.plot_s_time_mag(label = 'plot_s_time_mag')
> ```

This looks good!
test

mhuser (approved):

mhuser:
Great! Nice, the time for the reflexion going back and forth and the resolution divided by two for bandpass are clearly visibles!

Issue #379, opened 2020-07-31
Network.write_touchstone doesn't write the noise information


FournierVincent:
The Network.write_touchstone does write the S parameters but not the noise parameters as specified in the touchstone spec. 
https://ibis.org/connector/touchstone_spec11.pdf

```
    nw = rf.Network(f=[76.0],s=[[[0,1 ],[0,1]]],z0=[[50, 50]],name='noisynw',f_unit='GHz')
    nw.set_noise_a(nw.frequency, 3.3, 0.2+0.2j, 250 )
    nw.noisy
    nw.write_touchstone(return_string=True)

```
'!Created with skrf (http://scikit-rf.org).\n# GHz S RI R 50.0 \n!freq ReS11 ImS11 ReS21 ImS21 ReS12 ImS12 ReS22 ImS22\n76.0 0.0 0.0 0.0 0.0 1.0 0.0 1.0 0.0\n'

FournierVincent:
The methods for plotting (on a smith chart or ni rectangular) the noise parameters g_opt,  nfmindb, g_opt, rn are missing.
Idem for NetworkSet.

PR #906, opened 2023-05-02
Add noise data to write_touchstone

ericwrice:
Quick addition to write noise data to touchstone files as raised in #379. This follows the touchstone 1.1 [specification](https://ibis.org/connector/touchstone_spec11.pdf). Tested by reading in the following transistor touchstone and writing out the test file.

[LN300.s2p.txt](https://github.com/scikit-rf/scikit-rf/files/11377006/LN300.s2p.txt)
[test.s2p.txt](https://github.com/scikit-rf/scikit-rf/files/11377016/test.s2p.txt)

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2363,8 +2363,10 @@ class Network:
 
                 # write noise data if it exists
                 if ntwk.noisy:
-                   for f, nf, g_opt, rn, z0 in zip(ntwk.f_noise.f_scaled, ntwk.nfmin_db, ntwk.g_opt, ntwk.rn, ntwk.z0):
-                       output.write(f"{f} {nf} {mf.complex_2_magnitude(g_opt)} {mf.complex_2_degree(g_opt)} {rn/z0[0].real}\n") 
+                    new = ntwk.copy()
+                    new.resample(ntwk.f_noise) # only write data from original noise freqs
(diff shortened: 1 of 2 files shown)


jhillairet (approved):
LFTM.

Eventually if you could add associated tests?

ericwrice:
Yep, good idea, I'd be happy to add some tests. That made me notice it works fine with the file I used that had equal length scattering and noise parameters, but not with the noisy networks in the test folder that have different lengths.  

It looks to me that the original noise information is not stored in the network, only the interpolated data. Do we want write_touchstone() to use the interpolated data or find a way to maintain the original?

Probably a separate issue, but noise interpolation in Network.n fails on skrf/tests/ntwk4_n.s2p because the lowest noise frequency is less than the lowest s frequency, even though this is allowed in the touchstone specification. It's probably good practice to avoid extrapolating outside of provided data, but could make some valid touchstone files more difficult to work with.

jhillairet:
> It looks to me that the original noise information is not stored in the network, only the interpolated data. Do we want write_touchstone() to use the interpolated data or find a way to maintain the original?

I have no advice on this. Anyone?

ericwrice:
I decided it was best to only write the original data for now.  That way measured data doesn't get mixed with interpolated data for example. Also added a test.

jhillairet (approved):

jhillairet:
Thanks !

PR #908, opened 2023-05-06
Add gain properties

keikawa:
The maximum stable gain, available gain, and Mason's unilateral gain were added to the Network object.


skrf/network.py
```diff
@@ -1709,6 +1709,75 @@ def stability(self) -> npy.ndarray:
...
+                gmax = |S_{21}| / |S_{12}| * (K - sqrt(K^2 - 1))
+
+        Returns
+        -------
+        gmax : :class:`numpy.ndarray` of shape `f`
+
+        References
+        ----------
+        M. S. Gupta, "Power gain in feedback amplifiers, a classic revisited," in IEEE Transactions on Microwave Theory and Techniques, vol. 40, no. 5, pp. 864-879, May 1992, doi: 10.1109/22.137392.
+        """
```
jhillairet, line 1749:
```suggestion
        r"""
        Maximum stable power gain for K < 1 or Maximum available power gain for K >= 1 in linear value.

        .. math::

                G_{max} = \frac{|S_{21}|}{|S_{12}|} \times (K - \sqrt{K^2 - 1})

        Returns
        -------
        gmax : :class:`numpy.ndarray` of shape `f`

        References
        ----------
        M. S. Gupta, "Power gain in feedback amplifiers, a classic revisited," in IEEE Transactions on Microwave Theory and Techniques, vol. 40, no. 5, pp. 864-879, May 1992, doi: 10.1109/22.137392.
        
        See Also
        --------
        max_stable_gain : Maximum stable power gain
        unilateral_gain : Mason's unilateral power gain.
        stability : Stability factor
        """
```

skrf/network.py
```diff
@@ -1709,6 +1709,75 @@ def stability(self) -> npy.ndarray:
...
+                msg = |S_{21}| / |S_{12}|
+
+        Returns
+        -------
+        msg : :class:`numpy.ndarray` of shape `f`
+
+        References
+        ----------
+        M. S. Gupta, "Power gain in feedback amplifiers, a classic revisited," in IEEE Transactions on Microwave Theory and Techniques, vol. 40, no. 5, pp. 864-879, May 1992, doi: 10.1109/22.137392.
+        """
```
jhillairet, line 1728:
```suggestion
        r"""
        Maximum stable power gain in linear value.

        .. math::

                G_{ms} = |S_{21}| / |S_{12}|

        Returns
        -------
        msg : :class:`numpy.ndarray` of shape `f`

        References
        ----------
        M. S. Gupta, "Power gain in feedback amplifiers, a classic revisited," in IEEE Transactions on Microwave Theory and Techniques, vol. 40, no. 5, pp. 864-879, May 1992, doi: 10.1109/22.

        See Also
        --------
        max_gain : maximum stable power gain
        unilateral_gain : Mason's unilateral power gain.
        stability : Stability factor        
        """
```

skrf/network.py
```diff
@@ -1709,6 +1709,75 @@ def stability(self) -> npy.ndarray:
...
+        assert self.nports == 2, "Maximum stable gain is only defined for two ports"
+        msg = npy.abs(self.s[:, 1, 0]) / npy.abs(self.s[:, 0, 1])
+        return msg
+    
+    @property
+    def max_gain(self) -> npy.ndarray:
+        """
+        Returns maximum stable power gain for K < 1 or Maximum avalable power gain for K >= 1 in linear value
```
jhillairet, line 1736: isn't it the "Maximum Available Gain" instead?

skrf/network.py
```diff
@@ -1709,6 +1709,75 @@ def stability(self) -> npy.ndarray:
...
+                U = |S_{21} / S_{11} - 1| ^ 2 / (2 * K * |S_{21}| / |S_{12}| - 2 * Re[S_{21} / S_{12}])
+
+        Returns
+        -------
+        U : :class:`numpy.ndarray` of shape `f`
+
+        References
+        ----------
+        M. S. Gupta, "Power gain in feedback amplifiers, a classic revisited," in IEEE Transactions on Microwave Theory and Techniques, vol. 40, no. 5, pp. 864-879, May 1992, doi: 10.1109/22.137392.
+        """
```
jhillairet, line 1773: Please also add a See Also section

jhillairet (commented):
Thanks for contributing!

I've made some suggestion on the documentation, to be discussed.

Would you mind adding some associated tests eventually?

skrf/network.py
keikawa: Thank you for the suggestion.
If K > 1, max_gain returns the maximum available gain. On the other hand, if K <= 1, max_gain returns the maximum stable gain. This behavior is the same as the max_gain() function of Keysight's ADS. Thus, I defined the property name as simple as max_gain.

keikawa:
@jhillairet 
Thank you very much for review and suggestion!
I would like to add a test code, please wait a little while longer.

Ttl:
I fixed some issues with stability in #910 when S12 or S21 is zero. Can you make sure that these methods also work correctly when S12 or S21 are zero? Also I would prefer raising `ValueError` to `assert`.

keikawa:
Nice catch. I need to fix zero division issues in the gain properties.

@jhillairet 
Could you firstly merge #910 to the master branch bacause the gain properties depend on the stability?

jhillairet:
> @jhillairet Could you firstly merge #910 to the master branch bacause the gain properties depend on the stability?

OK, merged.

keikawa:
I added test codes and chaged ```assert``` to ```ValueError```.
For the zero division case, the gain properties return ```npy.inf``` or ```npy.nan``` with runtime warning (This operaiton is depend on Numpy).
Please advise me if you have any better way other than above.

jhillairet (commented):
Great, thank you. Small additionnal comments:
- could you comment your tests to explain in a few words their purpose?
- Eventually, if you could reduce the number of points of the test files (fet.s2p, maxgain_ads.csv). ~hundred of frequency points is enough for testing and it saves some space.

keikawa:
Thank you for comment!

@Ttl 
In my first commit, I used this equation for max_gain calculation.
Nevertheless, I changed the equation with no reason in the second commit.
I seem to have been writing code in my sleep...

By the way, does ```inf``` really make sense as a return value for ```max_stable_gain```?
If S12 = S21 = 0, msg is not always infinity.
Wouldn't it be more user-friendly to have nan as the return value like the default behavior of Numpy?

jhillairet (approved):

keikawa:
I understand that the default behavior of Numpy is to return `inf` when dividing a finite value by 0, and `nan` when dividing 0 by 0.
In either case, a runtimeWarning is printed.

The current test only detects runtime warnings, do I need to check `inf` and `nan` to see if they are output correctly?
I'm not familiar with testing, so if there is a better way, please let me know.

Ttl (approved):

jhillairet:
Thank you @keikawa 

I forgot to merge this PR before merging the new version, I'll make a minor version to integrate it!

Issue #914, opened 2023-05-24
__eq__ does not compare frequency


FranzForstmayr:
**Describe the bug**
Comparing two networks with identical `s` matrix but different `frequency` object returns True.

**To Reproduce**
```python
import skrf as rf
import numpy as np

s = np.random.randn(10, 2, 2)
f1 = np.arange(10)
f2 = f1 + 1

n1 = rf.Network(s=s, f=f1)
n2 = rf.Network(s=s, f=f2)

print(n1 == n2)  # True
print((f1 == f2).all())  # False
```

**Expected behavior**
Networks without different frequency should compare as false. Same with different `z0` and maybe other properties.

**System**
 - OS: independent
 - scikit-rf version: 0.27.1

PR #915, opened 2023-06-06
Fixed __eq__ issue

Hadimius:
Changed __eq__ so it accounts for differences in frequency and all primary properties. Refers to and solves issue #914

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -786,14 +786,23 @@ class Network:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, self.__class__):
             return False
-        if not (self.f == other.f).all():
+        if len(self.f) != len(other.f):
             return False
-        for prop in self.PRIMARY_PROPERTIES:
-            if npy.all(npy.abs(getattr(self,prop)-getattr(other,prop))< ZERO):
-                return True
-            else:
+        for prop in ['f','s','z0']:
+            if not npy.all(npy.abs(getattr(self,prop)-getattr(other,prop))< ZERO):
                 return False
-        
+        # if z0 is imaginary s_def is compared. If real of z0 is equal but s_def differs, networks are still equal
+        if ((npy.imag(self.z0) != 0).all()) or ((npy.imag(other.z0) != 0).all()):
+            if self.s_def == other.s_def:
+                return True
+            else:            
+                if npy.all(npy.abs(npy.real(self.z0)) - npy.abs(npy.real(other.z0)) < ZERO):
+                    return True 
+                else:
+                    return False
+        else:
+            return True
+
     def __ne__(self, other:object) -> bool:
         return (not self.__eq__(other))
 
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -795,11 +795,8 @@ class Network:
         if ((npy.imag(self.z0) != 0).all()) or ((npy.imag(other.z0) != 0).all()):
             if self.s_def == other.s_def:
                 return True
-            else:            
-                if npy.all(npy.abs(npy.real(self.z0)) - npy.abs(npy.real(other.z0)) < ZERO):
-                    return True 
-                else:
-                    return False
+            else:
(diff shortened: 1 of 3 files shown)


jhillairet (approved):
Great contribution, thank you!

Ttl:
Comparing all `PRIMARY_PROPERTIES` is unnecessary. `Network` object stores only S-parameters and other parameters are converted from them as necessary. It's enough to compare `f`, `s`, `s_def` and `z0`. `s_def` comparison is only necessary if any of `z0` has non-zero imaginary component. Different network with the same real `z0` and `s` are equal even if `s_def` is different. Also noise parameters should be compared if they exist.

`numpy` will raise error in future when comparing two different length arrays by equality. It would be better to first compare the number of frequency points. Tests would be nice to have to check all these different cases.

jhillairet:
I agree that comparing *all* properties is maybe overkill. Is the more the better? 

Or should we really restrict the comparisons to:
- `f`
-  `s`
- `z0`
- `s_def` (if any of z0 has non-zero imaginary component)

for noise parameter, eventually too.

skrf/network.py
```diff
@@ -786,10 +786,22 @@ def __div__(self, other: 'Network') -> 'Network':
...
             return False
+        for prop in ['f','s','z0']:
+            if not npy.all(npy.abs(getattr(self,prop)-getattr(other,prop))< ZERO):
+                return False
+        # if z0 is imaginary s_def is compared. If real of z0 is equal but s_def differs, networks are still equal
+        if ((npy.imag(self.z0) != 0).all()) or ((npy.imag(other.z0) != 0).all()):
+            if self.s_def == other.s_def:
+                return True
+            else:            
+                if npy.all(npy.abs(npy.real(self.z0)) - npy.abs(npy.real(other.z0)) < ZERO):
```
FranzForstmayr, line 799: The `abs` is at the wrong place.
Maybe use something like `return npy.allclose(self.z0.real, other.z0.real, atol=ZERO)`

skrf/tests/test_network.py
```diff
@@ -1743,6 +1743,28 @@ def test_stability(self):
...
+        s = npy.random.randn(10, 2, 2)
+        f1 = npy.arange(10)
+        f2 = npy.arange(10)
+        n1 = rf.Network(s=s,f=f1)
+        n2 = rf.Network(s=s,f=f2)
+        self.assertTrue(n1 == n2)
+
+        f2 = npy.arange(11)
+        n2 = rf.Network(s=s,f=f2)
+        self.assertFalse(n1 == n2)
```
FranzForstmayr, line 1756: Maybe one more test with f2 = f1 * 10, as this test returns when checking the length of the vectors.

Hadimius:
It seams to be failing because of the following test in test_media:
```python
    def test_impedance_mismatch(self):
        """
        Test the naming of the network. When circuit is used to connect a
        topology of networks, they should have unique names.
        """
        name = 'impedanceMismatch,50to25'
        qucs_ntwk = Network(os.path.join(self.files_dir, name + '.s2p'))
        self.dummy_media.frequency = qucs_ntwk.frequency
        skrf_ntwk = self.dummy_media.thru(z0=50, name = name)**\
            self.dummy_media.thru(z0=25)

        self.assertEqual(qucs_ntwk, skrf_ntwk)
        self.assertEqual(qucs_ntwk.name, skrf_ntwk.name)

``` 
I'm not 100% sure about this, but the Networks in this test should not be equal, since they differ in `z0`, or?
So should the test be adapted to compare only `f`, `s` and `name`?

FranzForstmayr:
I think the problem is, that the file stored in sNp has always the same port impedance on all ports, in this case 50.
You cannot store 50/25 Ohm in sNp, so this test has to fail by definition.

You could manually set qucs_ntwk.z0 to [50, 25] to avoid this issue. If you do, please write a small note :)

FranzForstmayr:
Can you merge in master please?

PR #918, opened 2023-06-16
Fix transposed circuit voltages

### LINKED ISSUE #917: Circuit S-parameter and voltage Calculation
I have defined a very simple two-port network with all S parameters at 0 except S21 which is 1. Then I set the power of port 1 to 1. But the voltage at port2 is 0. The code is shown below.

```
import skrf as rf
from pylab import *

f = np.array([1]) # in GHz
freq = rf.Frequency.from_f(f, unit='GHz')
s = array([[0, 0],
                  [1, 0]])
network = rf.Network(frequency = freq, s = s, name='n2')
port1 = rf.Circuit.Port(frequency=freq, name='port1', z0=50)
port2 = rf.Circuit.Port(frequency=freq, name='port2', z0=50)
cnx = [
    [(network,0),(port1,0)],
    [(network,1),(port2,0)]
]
crt = rf.Circuit(cnx)
power = [1,0]  # 1 Watt at port1 and 0 at port2
phase = [0,0]  # 0 radians
V_at_ports = crt.voltages_external(power, phase)
print(V_at_ports)
```
The expected result is `[[10.+0.j 10.+0.j]]`. But the result is` [[10.+0.j 0.+0.j]]`

But if I set S21 to 0 and S12 to 1,it's
```
s = array([[0, 1],
       [0, 0]])
```
The result is `[[10.+0.j 10.+0.j]]`

From the defination of S-parameter, S21 is V2-/V1+, So I think the results are reversed.
In circuit.py, I find the code uses np.transpose when calculating S-parameters and the comment says 'transpose is necessary to get expected result'. But I find if transpose isn't used, the result is correct.

**System**
 - Win10:
 - scikit-rf 0.27.1

Ttl:
Should fix #917. I'm not too familiar with the circuit code but this seems much more reasonable code than before. @jhillairet

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -1143,8 +1143,7 @@ class Circuit:
         phase excitation at "external" ports using `_a(power, phase)` method.
 
         """
-        # return self.s @ a_internal
-        return np.matmul(self.s, a_internal)
+        return self.s @ a_internal
 
     def currents(self, power: NumberLike, phase: NumberLike) -> np.ndarray:
         """
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -799,11 +799,7 @@ class Circuit:
         S : :class:`numpy.ndarray`
             global scattering parameters of the circuit.
         """
-        # transpose is necessary to get expected result
-        #return np.transpose(self.X @ np.linalg.inv(np.identity(self.dim) - self.C @ self.X), axes=(0,2,1))
-        # does not use the @ operator for backward Python version compatibility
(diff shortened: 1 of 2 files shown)


jhillairet:
Ho nice catch ! I was not aware of this indexing parameter... Thank you @Ttl 

And yes, this is a good opportunity to remove the old numpy code. Since you are here, can you also remove the line 1146?

jhillairet (approved):

jhillairet (approved):

jhillairet:
Thank you @Ttl .

Issue #921, opened 2023-06-19
CI fails from Numpy 1.25.0


jhillairet:
The test suite is failing from the version 1.25.0 of Numpy, in which the function `run_module_suite` [has been removed](https://github.com/numpy/numpy/releases/tag/v1.25.0). 

> NumPy switched to using pytest in 2018 and nose has been unmaintained for many years. We have kept NumPy's nose support to avoid breaking downstream projects who were might have been using it and not yet switched to pytest or some other testing framework. With the arrival of Python 3.12, unpatched nose will raise an error. It is time to move on.)

PR #922, opened 2023-06-20
Numpy1.25

FranzForstmayr:
Try to make skrf compatible to numpy 1.25


Ttl:
All the `[0]` indexing in `qfactor` looks suspicious. If they are actually one element vectors then it would make sense to instead make them scalars when they are created which I think is at the end of `_initial_fit` function.

FranzForstmayr:
I'm also not really a fan of this solution...
I will think of your approach the next days.

FranzForstmayr:
Will close #921

Ttl (approved):

PR #925, opened 2023-06-27
Allow to pass a callable fucnction as window argument

FranzForstmayr:
This is useful to define more customized windows and not just the window functions from scipy.

### DIFF
--- a/skrf/time.py
+++ b/skrf/time.py
@@ -228,13 +228,18 @@ def detect_span(ntwk: 'Network', t_unit: str = "") -> float:
 def get_window(window: Union[str, tuple, Callable], Nx: int, **kwargs) -> npy.ndarray:
     """Calls a custom window function or `scipy.signal.get_window()` depending on the window argument.
 
-    Args:
-        window (Union[str, tuple, Callable]): The callable window function or a valid window string
-            for `scipy.signal.get_window()`.
-        Nx (int): Number of samples
+    Parameters
+    ----------
+    window : str, tuple or Callable
+        The callable window function or a valid window string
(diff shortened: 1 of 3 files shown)


FranzForstmayr:
could we get this into #924 ? :)

skrf/time.py
```diff
@@ -225,7 +225,22 @@ def detect_span(ntwk: 'Network', t_unit: str = "") -> float:
...
+    """Calls a custom window function or `scipy.signal.get_window()` depending on the window argument.
 
+    Args:
+        window (Union[str, tuple, Callable]): The callable window function or a valid window string
+            for `scipy.signal.get_window()`.
+        Nx (int): Number of samples
+
+    Returns:
+        npy.ndarray: Window samples
+    """
```
jhillairet, line 238: switch to Numpy style docstring:
```suggestion
    Parameters
    ----------
    window : str, tuple or Callable
        The callable window function or a valid window string
        for `scipy.signal.get_window()`.
    Nx : int
        Number of samples

    Returns
    -------
    window : npy.ndarray
        Window samples.
    """
```

jhillairet (approved):
sure we can include it in the next release.

jhillairet (approved):

FranzForstmayr:
The function is currently not in the public API, do we need this for this case?
https://scikit-rf--925.org.readthedocs.build/en/925/api/time.html

jhillairet:
Eventually. You can add the function name in the docstring at the top of time.py

FranzForstmayr (commented):
Add `get_window` to sphinx doc

PR #926, opened 2023-06-28
Update calibration.py

jmcelroy01:
dedent typo


jhillairet (approved):
Thanks. Means this part has never been used or tested...

jhillairet:
merged. Seems it was my bad... strange that the linter does not spot it.

PR #929, opened 2023-07-03
Add TUG multiline TRL and a method to compute switch terms indirectly 

ZiadHatab:
Hi!

I have included my multiline TRL method. In the attached file, you will find test code using both measurements and synthetic data I copied from [this example](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Multiline%20TRL.html). The plots below show the Monte Carlo test I performed with additive noise.

I did not adopt the structure of the `NISTMultilineTRL` with regard to inputs. In my case, you pass the line measurements and reflect measurements in different variables. Also, in my case, you can run the code without reflect measurements if only the propagation constant and S21 and S12 of the calibrated DUT are relevant. The mathematical details of the calibration procedure can be found [here](https://ziadhatab.github.io/posts/multiline-trl-calibration/).

Also, I have included a method to compute the switch terms using reciprocal devices without prior calibration of the VNA. Tests are included in attached file. You can read more about the math behind this [here](https://ziadhatab.github.io/posts/vna-switch-terms/).

error_terms_multiplicative_noise

dut_white_noise

[PR_tests.zip](https://github.com/scikit-rf/scikit-rf/files/11939754/PR_tests.zip)


Ttl:
Code looks good but it needs tests (for both TRL and switch term solving). See [here](https://github.com/scikit-rf/scikit-rf/blob/master/skrf/calibration/tests/test_calibration.py#L756-L867) for `NISTMultilineTRL` tests. A [notebook](https://scikit-rf.readthedocs.io/en/latest/examples/metrology/Multiline%20TRL.html) similar to `NISTMultilineTRL` that would show how it's used and its advantages would be also welcome though not absolutely necessary.

ZiadHatab:
I didn't know about the test. I will adapt the ones made for `NISTMultilineTRL`.

Regarding the notebook, do you think I should rewrite the current notebook to include both multiline TRLs with comparison, or add a new notebook?

ZiadHatab:
yes, I can use the same data. I already was using the same data in the test files I attached when I opened this PR.

skrf/calibration/calibration.py
```diff
@@ -3470,7 +3475,424 @@ def load_calibration_archive(cls, filename):
...
+            
+    .. [3] https://ziadhatab.github.io/posts/multiline-trl-calibration/
+
+    See Also
+    --------
+    NISTMultilineTRL
+    """
+
+    family = 'TRL'
+    def __init__(self, line_meas, line_lengths, ereff_est=1-.0j,
```
Ttl, line 3530: The interface is different from `NISTMultilineTRL` and I think it's fine. However I would like to change `ereff_est` to `er_est` to be the same as `NISTMultilineTRL` so that they match instead of being slightly different to avoid mistakes when changing between the methods.

Ttl (changes requested):

skrf/calibration/calibration.py
```diff
@@ -3794,7 +3795,10 @@ def solve_quadratic(v1, v2, inx, x_est):
             else:
                 # solve for a11/b11, a11 and b11 (use redundant reflect measurement, if available)
                 reflect_est_offset = reflect_est*npy.exp(-2*gamma*reflect_offset) # shift estimated reflect
-                a11_b11 = (reflect_meas_S[:,m,0,0] - a12)/(1 - reflect_meas_S[:,m,0,0]*a21_a11) * (1 + reflect_meas_S[:,m,1,1]*b12_b11)/(reflect_meas_S[:,m,1,1] + b21)
+                # a11_b11 = (reflect_meas_S[:,m,0,0] - a12)/(1 - reflect_meas_S[:,m,0,0]*a21_a11) * (1 + reflect_meas_S[:,m,1,1]*b12_b11)/(reflect_meas_S[:,m,1,1] + b21)
```
Ttl, line 3798: What's this? If the new code is better this can be removed.

Ttl (approved):

skrf/calibration/calibration.py
ZiadHatab: To be honest, the first code shouldn't have been there, I initially added it while doing some debugging when I first opened the PR and then realized I didn't delete it. Even now, somehow I ended up commenting it instead of deleting it. I will delete it.

skrf/calibration/calibration.py
```diff
@@ -3470,7 +3475,427 @@ def load_calibration_archive(cls, filename):
 
         return cal
 
+class TUGMultilineTRL(EightTerm):
+    """
+    TUG Multiline TRL calibration.
+
+    An improved multiline TRL calibration procedure that generalizes the calibration process 
+    by solving a single 4x4 weighted eigenvalue problem.
```
jhillairet, line 3483:
```suggestion
    by solving a single 4x4 weighted eigenvalue problem. TUG stands for Technical University of Graz (Austria).
```

PR #930, opened 2023-07-08
Update Calibration.ipynb

mjpelmear:
Apparently the function name changed at some point.


FranzForstmayr:
Good Catch!
Usually the notebooks should get executed after each PR, but obviously this one wasnt.
We should find the reason why this notebook is exclude from testing.

jhillairet (approved):

Ttl:
It looks like all of the cells in this notebook are markdown and there is no executable code. This should be rewritten so that the code is actually executed.

jhillairet:
Yes indeed, that's because it's a generic explanation, not a specific example. But eventually it can be transformer to a practical example.

Issue #362, opened 2020-07-04
Stability and gain circles


miguel9554:
Is there any way of plotting satbility and gain circles of a network?

jhillairet:
There is a [stability parameter](https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.network.Network.stability.html) of a Nework, but no plotting things (yet). Not yet. Of course, new contributions are welcome :)

miguel9554:
I've managed to plot them myself, it's a matter of computing the center and radius of the circles from the S-parameters, and plotting. I will try to add it to the project.

curtisma:
Any news on this feature request?  We also have an interest in being able to plot circles on the smith chart for gain and noise figure.

@miguel9554 Can you share how you implemented it previously?  Maybe share the code?

Vinc0110:
Hi @keikawa, in your PR #943 you announced some plans to implement gain circles as well. Is there any progress already? It would be awesome if you could help to finally finish and close this old feature request.

keikawa:
Hi @Vinc0110, Thank you for the reminder. I will send a pull request later this week.

PR #943, opened 2023-08-20
Stability circles

keikawa:
This pull request adds stability circle calculation methods related to #362.


skrf/network.py
```diff
@@ -4267,6 +4272,177 @@ def vswr_active(self, a: npy.ndarray) -> npy.ndarray:
...
+            raise ValueError("Stability circle is only defined for two ports")
+        
+        if npoints <= 0:
+            raise ValueError("npoints must be a positive integer")
+
+        # Calculate the determinant of the scattering matrix
+        D = self.s[:, 0, 0] * self.s[:, 1, 1] - self.s[:, 0, 1] * self.s[:, 1, 0]
+
+        # Calculate the center and radius of the source stability circle
+        ssc_center = (self.s[:, 0, 0] - self.s[:, 1, 1].conjugate() * D).conjugate() / (npy.abs(self.s[:, 0, 0]) ** 2 - npy.abs(D) ** 2)
```
FranzForstmayr, line 4432: This two lines are the only differences between `load` and `source`  circles, isn't it?
Would it be better to make a generic `stability_circle` method where one can choose `source` or `load` via an attribute?

keikawa: Thank you for your suggestion. I have combined these two methods into one.

jhillairet (approved):
Nice addition, thank you. 

Do you think you could write an associated notebook for the tutorial part of the documentation, to explain what are stability circles and how to use it?

keikawa:
That sounds good. 
I'm planning to start by implementing the gain circle calculation methods and then move on to creating a notebook, because stability circles are typically used with gain circles, as requested in #362.
But, unfortunately, I'm currently busy this month due to preparations for conferences, so I'd prefer to send other pull requests for these tasks.

Issue #947, opened 2023-08-23
network_2_dataframe does not work for more than 10 ports


FranzForstmayr:
Thank you for reporting a bug! To help us understanding the issue and how to reproduce it, please fill the following sections as best as you can.

**Describe the bug**
`network_2_dataframe` does not work for more than 10 ports

**To Reproduce**

```python
import skrf as rf

s = np.random.standard_normal((1, 11, 11))
f = [1]
netw = rf.Network(s=s, f=f)

s2 = netw.to_dataframe().values
assert s.size == s2.size
```

**Expected behavior**
`network_2_dataframe` should not drop data silently.

Issue behind is the column name creation in `skrf.io.general.network_2_dataframe` which result in the same column name S 111 for S (1,11) and S(11,1). Maybe we should add commas in the column name too, or use a different approach.

cafeclimber:
Alot of SCPI stuff I've looked at uses underscores? E.g. S2_1, S11_1, S1_11, etc.

PR #955, opened 2023-09-07
Fix network_2_dataframe for more than 10 ports

FranzForstmayr:
Currently the columns in the exported dataframe look like `s_db 11`, `s_db 21` ...
There's a name clash in case of more than 11 ports as s_db 1_11 and s_db 11_1 look the same if the separator is removed.

Now a port separator can be specified manually.
If not specified, `` is used for networks with a maximum of 10 ports, a `_` otherwise. This allows to keep backward compability.


FranzForstmayr:
Will close #947

jhillairet (approved):

PR #958, opened 2023-09-15
Add timeout argument to VNA constructor

cafeclimber:
Very simple. I was running into issues where measurements were taking too long. Without being able to set timeout, I had to set it directly on the internal `_resource` variable which is...gross.


FranzForstmayr:
@cafeclimber please merge master into this branch, as the notebook issue with matplotlib is fixed in master branch. #959

cafeclimber:
Should be good!

jhillairet (approved):

PR #970, opened 2023-11-02
Remove `html_theme_path` setting from docs config to fix search & other JavaScript functions

miek:
I've noticed that search and other JavaScript functions (like opening the version information box in the bottom left) fail on the readthedocs page, which I think is due to an exception thrown over missing jQuery.

I found some advice in the sphinx_rtd_theme issues ([1](https://github.com/readthedocs/sphinx_rtd_theme/issues/1434#issuecomment-1472671651), [2](https://github.com/readthedocs/sphinx_rtd_theme/issues/1452#issuecomment-1575476244)) suggesting this can be caused by setting `html_theme_path`, which was previously recommended but is now out of date.

This updates the config in-line with current advice from sphinx_rtd_theme adding it to `extensions` and removing the `html_theme_path` setting.

~~I wasn't able to reproduce the problem offline to test, but I'm hoping this PR kicks off an RTD build to test it out~~
Edit: the build worked, and it seems to be fixed! https://scikit-rf--970.org.readthedocs.build/en/970/

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -12,7 +12,6 @@ author = 'scikit-rf team'
 
 import sys
 import os
-import sphinx_rtd_theme
 import warnings
 warnings.filterwarnings('ignore')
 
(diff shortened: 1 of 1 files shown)


Vinc0110 (approved):
Great, thanks for reporting and fixing this!
I have not idea about the details, but I can confirm the search now works again in your test build of the docs.

jhillairet (approved):

jhillairet:
Thanks a lot !

PR #971, opened 2023-11-04
Fix de-embedding when 2 networks are provided

atavella:
The network operator "//" used for de-embed 1 or 2 networks is not working for exactly 2 networks.  

ntwk1 // ntwk2 works
ntwk1 // (ntwk2) works
ntwk1 // (ntwk2, ntwk3) does not work

it fixes that

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -618,10 +618,7 @@ class Network:
 
         if isinstance(other, (list, tuple)):
             if len(other) >= 3:
-                warnings.warn(
-                    "Number of networks greater than 2. Truncating!",
-                    RuntimeWarning
-                )
+                raise ValueError('Incorrect number of networks.')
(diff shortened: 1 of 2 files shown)


FranzForstmayr:
I think we should even be more strict and raise a Exception (RuntimeError?) if more than two elements are passed for other.

atavella:
@FranzForstmayr that makes sense to me. But, right now is raising a RuntimeWarning if more than two elements are passed and the flow continues with just the first two elements.

FranzForstmayr:
Yes but warnings are likely to get overseen, and there's no reason to accept more than two for others.
If some existing code triggers the warning, there's probably another issue in the code which is simply not found yet.

jhillairet (approved):

FranzForstmayr:
Thanks @atavella

PR #983, opened 2023-12-08
Fixed SOLT calibration

floschl:
Hi there,

I have noticed that the SOLT calibration does not work as desired.
I have not prepared a minimal working example, but this is my use case:
I have
- Three one-port reflectance standards, both measured and idealised
- known

After running the SOLT calibration and measuring the known through, I would expect the through to be measured as known. However, the S21, for example, is centred around 0 dB.

The reason for this is that the wrong equation is used within the run() calibration method to calculate the transmission tracking (e10*e32). I assume that the eq from [1], p. 18 was simply used. However, this eq assumes that the connection between port 1 and port 2 is a zero thru (i.e. direct connection of both ports: S21=S12=1, S11=S22=0).
As this is not the case here, because we have given the SOLT the known thru, we have to consider the actual s-params for the thru when calculating the transmission tracking (e10*e32).

**The solution is quite simple: just take the S21M eq from [1] p.16 and solve for e10*e32.**.
The known thru will then be properly taken into account in the transmission retracking and the measurement will output the actual s-params after calibration.

Sorry for the lack of plots. I think the reason is quite obvious.

Thank you very much for your work and for considering this request!

Best regards,
Florian

PS: In addition, I generalised the paths of two of the (vectorfitting) tests that failed because I called the test from the PyCharm GUI, which caused the paths to fail - they are expected to be called from the repo's base path.

[1] Rytting, D. (1996) Network Analyzer Error Models and Calibration Methods. RF 8: Microwave Measurements for Wireless Applications (ARFTG/NIST Short Course Notes)

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -4,6 +4,7 @@ import skrf
 import numpy as np
 import tempfile
 import os
+from pathlib import Path
 import sys
 
 
@@ -44,14 +45,16 @@ class VectorFittingTestCase(unittest.TestCase):
 
     def test_190ghz_measured(self):
         # perform the fit without proportional term
-        nw = skrf.network.Network('./../../doc/source/examples/vectorfitting/190ghz_tx_measured.S2P')
+        s2p_file = Path(__file__).parent.parent.parent / 'doc/source/examples/vectorfitting/190ghz_tx_measured.S2P'
+        nw = skrf.network.Network(s2p_file)
         vf = skrf.vectorFitting.VectorFitting(nw)
         vf.vector_fit(n_poles_real=4, n_poles_cmplx=4, fit_proportional=False, fit_constant=True)
         self.assertLess(vf.get_rms_error(), 0.02)
 
     def test_no_convergence(self):
+        s2p_file = Path(__file__).parent.parent.parent / 'doc/source/examples/vectorfitting/190ghz_tx_measured.S2P'
         # perform a bad fit that does not converge and check if a RuntimeWarning is given
-        nw = skrf.network.Network('./../../doc/source/examples/vectorfitting/190ghz_tx_measured.S2P')
+        nw = skrf.network.Network(s2p_file)
         vf = skrf.vectorFitting.VectorFitting(nw)
 
         with pytest.warns(RuntimeWarning) as record:
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -44,14 +44,14 @@ class VectorFittingTestCase(unittest.TestCase):
 
     def test_190ghz_measured(self):
         # perform the fit without proportional term
-        nw = skrf.network.Network('./doc/source/examples/vectorfitting/190ghz_tx_measured.S2P')
+        nw = skrf.network.Network('./../../doc/source/examples/vectorfitting/190ghz_tx_measured.S2P')
         vf = skrf.vectorFitting.VectorFitting(nw)
         vf.vector_fit(n_poles_real=4, n_poles_cmplx=4, fit_proportional=False, fit_constant=True)
         self.assertLess(vf.get_rms_error(), 0.02)
 
     def test_no_convergence(self):
         # perform a bad fit that does not converge and check if a RuntimeWarning is given
-        nw = skrf.network.Network('./doc/source/examples/vectorfitting/190ghz_tx_measured.S2P')
+        nw = skrf.network.Network('./../../doc/source/examples/vectorfitting/190ghz_tx_measured.S2P')
         vf = skrf.vectorFitting.VectorFitting(nw)
 
         with pytest.warns(RuntimeWarning) as record:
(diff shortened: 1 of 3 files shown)


arsenovic:
excellent! great catch.  would it be possible for you to  modify the test_calibration to include this feature?

floschl:
I now see that some tests are failing, specifically 
SOLTTest.
test_accuracy_of_dut_correction
test_convert_12term_2_8term_correction_accuracy
test_embed_equal_measure
test_reverse_transmission_tracking_accuracy

but have not yet fully understood the full logic behind it and also why it succeeded before the change. It is a bit odd that the failing tests succeed for the other calibration methods, i.e. that either sth broke now or the definition of the tests is wrong in general.

`
    def test_forward_transmission_tracking_accuracy(self):
        self.assertEqual(
            self.Xf.s21*self.Yf.s21 ,
            self.cal.coefs_ntwks['forward transmission tracking'])
` 

does this self.Xf.s21*self.Yf.s21 correspond to the coefficients e10*e32 from [3]?

floschl:
Just found that also the computation of the load match is wrong for the same reason as for the transmission tracking (solve eq S11M in [3] p. 16 for e22). 
I will post an update the next days.

arsenovic:
aahh yes, ttl is right on this.   thanks ttl.

floschl:
Thanks for the info, I missed the docstring comment of the SOLT class.
After having a short look into the code of the `TwelveTerm` class, I see the that the formulas are the same as the ones that I am after. That's logical because the reference [2] mentioned in the docstring of the `TwelveTerm` class points to the same as [1]. 

Nevertheless, I do not see any reason why SOLT should not work with non-flush thrus, it simply a matter of using the right formula, as described above. Surely this should not break backward compatibility to also work with the flush-thru.

[2] "Calibration Process of Automatic Network Analyzer Systems"  by Stig Rehnmark

floschl:
Ok, so also fixed the calculus of the load matches. The `SOLT` class works now also with a non-flush thru. 
The `MultiportSOLTTest` now also passes with an arbitrary thru.

All calibration tests in `test_calibration.py` do pass.

It would be great to merge the pull request into the main branch. As I said, backward-compatibility is maintained, it is just less confusing for users that simply want to use the SOLT class and miss the comment to not use the SOLT class in case of a non-flush standard.

arsenovic:
so now that we were reminded that twelveterm exists what is the reason not to use twelve term?
fixing SOLT to be like twelve term is a redundant algo , no?

floschl:
indeed there is not much reason why both exists. The only feature that the twelve term class has in addition to the SOLT class is that categorize the thru standards.
In my opinion why to leave the SOLT class in this state. SOLT is a very common method with a common understanding what it can do, there is no reason that should be capable of only working with flush thrus.
My suggestion: include the fix and leave both SOLT and TwelveTerm classes for reasons of backward compatibility.

Ttl:
The code in `SOLT` is essentially just duplicate of what `TwelveTerm` already has. Neither class can be removed because of backwards compatibility, but ideally we wouldn't have duplicate code either. For example see how `TwoPortOnePath` class `TwelveTerm` calibration.

Note that `VectorFittingTestCase.test_dc` fails in CI. PR labeler also has issues but it isn't related to this PR.

floschl:
> Note that `VectorFittingTestCase.test_dc` fails in CI. PR labeler also has issues but it isn't related to this PR.

Did you pull the latest changes?

I just ran all tests: VectorFittingTestCase.test_dc passes
The only ones that fail are:
NetworkTestCase.
test_max_gain
test_max_stable_gain
test_unilateral_gain

None of them are related to the change in the SOLT class. All are related to a check whether a RuntimeWarning is thrown in case of a division by zero. 

I think in any case it would be good to fix the CI and the tests (including the PR labeler).

arsenovic:
ttl i believe you are suggesting that SOLT inherit from TwelveTerm?

FranzForstmayr:
You can ignore PR labeler for now, there's an upstream issue in the action itself.

skrf/tests/test_vectorfitting.py
```diff
@@ -61,7 +64,7 @@ def test_no_convergence(self):
 
     def test_dc(self):
         # perform the fit on data including a dc sample (0 Hz)
-        nw = skrf.Network('./skrf/tests/cst_example_4ports.s4p')
+        nw = skrf.Network('./cst_example_4ports.s4p')
```
FranzForstmayr, line 67: This change causes the test_dc to fail.

Ttl:
You can see the CI checks in the list at the bottom. For example here is a link to Python 3.11 tests which fails at check_dc with FileNotFoundError: https://github.com/scikit-rf/scikit-rf/actions/runs/7178376697/job/19555281436?pr=983

SOLT already inherits from TwelveTerm. It could just call its `run` function too.

floschl:
1. Ah yes, the test_dc tests passes on my side because I run the test via the PyCharm IDE which runs pytest from the local dir.
I just got the test_dc failing and pushed a fix of the test.

2. Forget about the failing tests (NetworkTestCase), this was caused by uncommenting `error` on my side in pyproject.toml (for another reason):

filterwarnings = [
    "error",

3. I just tried to comment out the run method from the SOLT class, all test_calibration tests including the SOLTTest pass. I would prefer such a solution as code duplication is avoided, backward compatibility is maintained and also an arbitrary flush works for the SOLT class.

arsenovic:
that seems like the easiest and most coherent solution. i apologize for not pointing this out earlier. 
i think thei PR can be left alone, and you could make a new one with the solution of simply removing SOLT.run?   any reason to keep this around  @Ttl for old times sake?  like SOLT.__old_run()?

Ttl:
I don't see any reason why `run` can't be completely removed from `SOLT`. Docstring should be also updated and there should be at least one test for `SOLT` with non-flush thru.

arsenovic:
ok i will close this one then. @floschl  hope you patience to make a new one. thank you for your effort!

floschl:
@arsenovic 
No problem, will do that on coming Tuesday.

PR #990, opened 2023-12-19
Make SOLT class inherit run method from TwelveTermModel to allow arbitrary thrus

floschl:
This is a follow-up PR from this PR #983 

To date, SOLT class does only accept flush thrus. This a modification to make SOLT class inherit run method from TwelveTermModel to allow arbitrary thrus, as discussed in #983 .

Docstrings were also adjusted.

### DIFF
--- a/skrf/calibration/tests/test_calibration.py
+++ b/skrf/calibration/tests/test_calibration.py
@@ -1229,16 +1229,10 @@ class SOLTTest(TwelveTermTest):
             wg.short(nports=2, name='short'),
             wg.open(nports=2, name='open'),
             wg.match(nports=2, name='load'),
-            None,
-            ]
-        actuals = [
-            wg.short(nports=2, name='short'),
-            wg.open(nports=2, name='open'),
-            wg.match(nports=2, name='load'),
-            wg.thru(),
+            wg.random(n_ports=2),
             ]
 
-        measured = [ self.measure(k) for k in actuals]
+        measured = [ self.measure(k) for k in ideals]
 
         self.cal = SOLT(
             ideals = ideals,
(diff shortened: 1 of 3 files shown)


skrf/calibration/tests/test_calibration.py
```diff
@@ -1229,16 +1229,10 @@ def setUp(self):
...
-        actuals = [
-            wg.short(nports=2, name='short'),
-            wg.open(nports=2, name='open'),
-            wg.match(nports=2, name='load'),
-            wg.thru(),
+            wg.random(n_ports=2),
             ]
 
-        measured = [ self.measure(k) for k in actuals]
+        measured = [ self.measure(k) for k in ideals]
```
Ttl, line 1235: This test previously tested that without specifying an ideal thru a flush thru is assumed. There isn't now a test for it and it's now the only difference between SOLT and TwelveTerm. I think this could be reverted. In `MultiportSOLTTest` it should be possible to replace thru with`wg.line` though.

floschl: done.

`SOLTTest `now tests again a flush thru if no ideal thru is given. `MultiportSOLTTest `tests against non-flush thru mismatch**line**mismatch

Ttl (approved):

jhillairet (approved):

jhillairet:
Thank you @floschl !

floschl:
You are welcome! Keep your great work up!

skrf/calibration/tests/test_calibration.py
arsenovic: great!

PR #991, opened 2023-12-20
Allow printing of Networks with empty frequency range

eendebakpt:
* Improve exception message for missing attribute
* Allow printing of Networks with empty frequency range
* Misc changes added by the ruff formatter

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -946,7 +946,7 @@ class Network:
             ntwk.s = self.s[:, t0, t1]
             ntwk.z0 = self.z0[:, t0]
             return ntwk
-        raise AttributeError(f'{self} does not have attribute {name}')
+        raise AttributeError(f'object does not have attribute {name}')
 
     def __dir__(self):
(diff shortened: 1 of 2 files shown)


skrf/network.py
```diff
@@ -907,7 +904,11 @@ def __str__(self) -> str:
         if _z0.ndim < 2:
             z0 = _z0
         else:
-            z0 = self.z0[0, :]
+            if _z0.size > 0:
+                z0 = _z0[0, :]
+            else:
+                # empty frequency range
+                z0 = float('nan')
```
jhillairet, line 911: Did the previous version was an issue in some situation? If yes, could you add a dedicated test?

eendebakpt: @jhillairet I added a test. The problem occurs in case one creates a network with an empty frequency range. That should not happen often in practice, but it can happen and the representation method of the Network object should work in that case.

jhillairet (approved):

Issue #354, opened 2020-05-13
HFSS touchstone import leads to port with 0 ohms => issue with rf circuits


freakyuser-osm:
I've encountered an issue with HFSS touchstone files and there port impedance

Please find the touchstone file from the HFSS Spiral Inductor Example here:
[Spiral_Inductor_Microstrip_Spiral.zip](https://github.com/scikit-rf/scikit-rf/files/4621151/Spiral_Inductor_Microstrip_Spiral.zip)

When the imported network is directly plotted, everything works fine and gives the same results as in HFSS. But it seems like only one port has a 50 ohm impedance, the other one is 0 ohms. Is this an export issue of HFSS or an import issue from scikit-rf?

The wrong port impedance leads to issues, when the rf circuit feature is used, and gives thus wrong results. (Python gives some divided by zero error messages as well).

I've added some example code and the output graphs:
![Circuit_plot](https://user-images.githubusercontent.com/23432012/81799263-603fd480-9511-11ea-9314-982dda49bed9.png)
![S_params](https://user-images.githubusercontent.com/23432012/81799285-66ce4c00-9511-11ea-9bae-9b1320774725.png)

```
import skrf as rf
import matplotlib.pyplot as plt

fig= plt.figure(figsize=(16,6))
ax1 = fig.add_subplot(211)
ax3 = fig.add_subplot(212)

readnet = rf.Network('C:\\PATH\\Spiral_Inductor_Microstrip_Spiral.s2p')
print(readnet)
readnet.plot_s_db(ax=ax1)

freq = readnet.frequency
port1 = rf.Circuit.Port(freq, name='port1', z0=50)
port2 = rf.Circuit.Port(freq, name='port2', z0=50)

connexions = [
        [(port1, 0 ), (readnet,0)],
        [(readnet,1), (port2,0)],
]

resulting_circuit = rf.Circuit(connexions)

resulting_circuit.plot_graph(network_labels=True, network_fontsize=15,
               port_labels=True, port_fontsize=15,
              edge_labels=True, edge_fontsize=10)

resulting_network = resulting_circuit.network

resulting_network.plot_s_db(ax=ax3)
```

jhillairet:
Indeed I confirm. 

But the Touchstone file is weird: it's a 2 ports, but there are 4 values of Gamma and z0...

jhillairet:
[...]

Issue #753, opened 2022-09-11
Touchstone cannot parse Example 5 from Touchstone spec


NAThompson:
The [Touchstone spec](http://www.ibis.org/touchstone_ver2.0/touchstone_ver2_0.pdf) gives a few examples of certain representative touchstone files. Consider example 5, on page 11:

```
! 4-port S-parameter data
! Default impedance is overridden by the [Reference] keyword arguments
! Data cannot be represented using 1.0 syntax
[Version] 2.0
# GHz S MA R 50
[Number of Ports] 4
[Number of Frequencies] 1
[Reference] 50 75 0.01 0.01
[Matrix Format] Full
[Network Data]
5.00000 0.60 161.24 0.40 -42.20 0.42 -66.58 0.53 -79.34 !row 1
0.40 -42.20 0.60 161.20 0.53 -79.34 0.42 -66.58 !row 2
0.42 -66.58 0.53 -79.34 0.60 161.24 0.40 -42.20 !row 3
0.53 -79.34 0.42 -66.58 0.40 -42.20 0.60 161.24 !row 4
```

Now attempt to parse this:

```python
>>> from skrf.io.touchstone import Touchstone
>>> ts = Touchstone('example_5.ts')
Traceback (most recent call last):
  File "~/lib/python3.9/site-packages/skrf/io/touchstone.py", line 135, in __init__
    self.load_file(fid)
  File "~/lib/python3.9/site-packages/skrf/io/touchstone.py", line 280, in load_file
    values.extend([ float(v) for v in line.split() ])
  File "~/lib/python3.9/site-packages/skrf/io/touchstone.py", line 280, in <listcomp>
    values.extend([ float(v) for v in line.split() ])
ValueError: could not convert string to float: '[matrix'
```

jhillairet:
That's right, Touchstone format V2 is not yet supported. It's a good reminder that a support of the V2 should be great before releasing a version 1.0

Vinc0110:
Another user on the email list was stumbling across the missing touchstone v2 support: https://groups.google.com/g/scikit-rf/c/79ZOyQV8pxc

I think a UserWarning would be a good idea in such cases, so users know what the problem is. Of course, we should also start working on adding the v2 support.

@jhillairet Do you know if anyone has already started working on it?

jhillairet:
I'd like to support it before releasing v1.0.

As far as I know, nobody is working on it. It's in my to-do list, but if
someone wants to tackle it ;)

[...]

PR #993, opened 2024-01-03
Start touchstone 2 support

### LINKED ISSUE #400: Touchstone parser issues with Ansys touchstone 2.0 files
Hello,

There are some issues parsing Touchstone 2.0 files generated by HFSS/Network Explorer.  
An example of such file is:
[ansys_test.zip](https://github.com/scikit-rf/scikit-rf/files/5362864/ansys_test.zip)

I've noticed three problems.  One is that the parser doesn't support `[matrix format]` keyword that is used by Ansys (although it only seems to generate full matrices). 

The parser is sensitive to location of  `[number of ports]` keyword when there are HFSS port names in the comments.  The comments precede the keyword. The parser seems to check the port number before the `rank` variable is assigned. And  `rank` is not know up front because this is a .ts file.

This is a mixed port reference case where  `[reference]` spans multiple lines -- and has in-line comment -- but the parser doesn't handle this situation.

LR

FranzForstmayr:
This should close #400.


FranzForstmayr:
The guys who invented touchstone format should write their parsers theirself....

FranzForstmayr:
Does anyone use the `Touchstone` class itself without `Network`? Thinking about backward compatibilty...

FranzForstmayr:
#753 should also work now.

FranzForstmayr:
Mixed mode is not implemented yet.

FranzForstmayr:
Should fix #354 as well now.

FranzForstmayr:
from #354 

This file is marked as "Terminal data exported" but has only 1xN Port impedances....
https://github.com/scikit-rf/scikit-rf/blob/master/doc/source/examples/circuit/circuit_vi_Coaxial_T.s3p

This is why the notebooks are failing currently.

jhillairet (approved):
Amazing work, the new parser is very impressive! 

Some parts of it looks like magic to me, I'm discovering some Python syntax (like `self._parse_dict: dict[str, Callable[[str], None]] = {` etc). I wonder if you could add some comments here and there for future readers.

FranzForstmayr:
> Amazing work, the new parser is very impressive!

> Some parts of it looks like magic to me, I'm discovering some Python syntax (like self._parse_dict: dict[str, Callable[[str], None]] = { etc). I wonder if you could add some comments here and there for future readers.

Will do of course!

@jhillairet 
Any thoughts about this file?
https://github.com/scikit-rf/scikit-rf/blob/master/doc/source/examples/circuit/circuit_vi_Coaxial_T.s3p

jhillairet:
> @jhillairet Any thoughts about this file? https://github.com/scikit-rf/scikit-rf/blob/master/doc/source/examples/circuit/circuit_vi_Coaxial_T.s3p

Well, It's a very standard 2port solved in modal setup. Here is the same Touchstone exported with 2022.2. Any change with this one?

[circuit_vi_Coaxial_T.zip](https://github.com/scikit-rf/scikit-rf/files/13891478/circuit_vi_Coaxial_T.zip)

FranzForstmayr:
@jhillairet does it make any difference if you disable gamma export?
https://github.com/scikit-rf/scikit-rf/issues/354#issuecomment-627930213 we can see here NxN impedances but no gamma.

jhillairet:
> @jhillairet does it make any difference if you disable gamma export? 
Here are the different exports (also the recent "non standard Touchstone extensions for mixed reference impedances" option) 
[circuit_vi_Coaxial_T_2.zip](https://github.com/scikit-rf/scikit-rf/files/13893843/circuit_vi_Coaxial_T_2.zip)

jhillairet (approved):

PR #935, opened 2023-07-21
Using SI Units

jhillairet:
- use Hz instead of GHz as new default
- use second instead of nanosecond as new default

PR #1008, opened 2024-01-22
Set default f_unit to Hz

FranzForstmayr:
This is a follow up PR to #935.

The default unit meant to be changed from `ghz` to `hz`, but this was missed obviously.


jhillairet (approved):
Ho dear... My bad.

PR #1009, opened 2024-01-23
Finalize ts2 support

FranzForstmayr:
Including Parameters different from S


jhillairet (approved):
Should we make a minor release including this asap? Seems logic to me.

FranzForstmayr:
yes we could indeed. 
I just thought about deprecate the `Touchstone` for public use, so that users should only use the `Network` class. We could do some more cleanup, if we do not need to care about backward compatibility.

skrf/io/touchstone.py
```diff
@@ -48,6 +48,7 @@ def remove_prefix(text: str, prefix: str) -> str:
         return text[len(prefix) :]
     return text
 
+FREQ_UNITS = {"hz": 1.0, "khz": 1e3, "mhz": 1e6, "ghz": 1e9}
```
jhillairet, line 51: you can add THz as it used by some users. (we have margin for PHz, EHz and ZHz...)

jhillairet (approved):

PR #1019, opened 2024-01-29
Remove white-space from directory name

mchristen-astranis:
`skrf_qtapps/skrf_qtwidgets/analyzers/driver development` contains a space in it which causes issues with installing `skrf` via https://github.com/bazelbuild/rules_python.

Moreover this is the only directory in the repo that contains a space, and as long as there is no particular reason to have a space here, it seems fine to rename in a way that removes the space.


FranzForstmayr:
I don't know Bazel, but it seems like a build tool. The qtapps are probably not needed in your application, maybe you can exclude them from the build completely?

mchristen-astranis:
> I don't know Bazel, but it seems like a build tool. The qtapps are probably not needed in your application, maybe you can exclude them from the build completely?

Thanks for the response, yep, Bazel is a build tool, https://bazel.build/. I'm working on an upstream patch for `rules_python` to also exclude directories with whitespace, but even if that does land, it seems like having a directory with whitespace isn't intentional and this would be a good change for scikit-rf? Or would y'all rather not update it?

jhillairet (approved):
OK, I have nothing against

mchristen-astranis:
It looks like I require approval for running the remaining CI checks?

mchristen-astranis:
Lovely thanks! This should be good to merge whenever works for y'all.

jhillairet (approved):

Issue #1022, opened 2024-02-03
S-parameter calculation incorrect for 3-port circuit


Fabdef:
**Bug**
S-parameter calculation is incorrect for a 3-port circuit

**Procedure to reproduce the bug**
Take the Wilkinson circuit available on scikit-rf website:
https://scikit-rf.readthedocs.io/en/latest/examples/circuit/Wilkinson%20Power%20Splitter.html

Copy the code and remove the resistor, which gives a Tee circuit with impedance matching:

```python
# standard imports
import numpy as np
import matplotlib.pyplot as plt
import skrf as rf
rf.stylely()

# frequency band
freq = rf.Frequency(start=0, stop=2, npoints=501, unit='GHz')

# characteristic impedance of the ports
z0_ports = 50

# resistor
R = 100
line_resistor = rf.media.DefinedGammaZ0(frequency=freq, z0=R)
resistor = line_resistor.resistor(R, name='resistor')

# branches
z0_branches = np.sqrt(2)*z0_ports
beta = freq.w/rf.c
line_branches = rf.media.DefinedGammaZ0(frequency=freq, z0=z0_branches, gamma=0+beta*1j)

d = line_branches.theta_2_d(90, deg=True)  # @ 90°(lambda/4)@ 1 GHz is ~ 75 mm
branch1 = line_branches.line(d, unit='m', name='branch1')
branch2 = line_branches.line(d, unit='m', name='branch2')

# ports
port1 = rf.Circuit.Port(freq, name='port1', z0=50)
port2 = rf.Circuit.Port(freq, name='port2', z0=50)
port3 = rf.Circuit.Port(freq, name='port3', z0=50)

# Connection setup
#┬Note that the order of appearance of the port in the setup is important
connections = [
           [(port1, 0), (branch1, 0), (branch2, 0)],
           [(port2, 0), (branch1, 1)],
           [(port3, 0), (branch2, 1)]
        ]

# Building the circuit
C = rf.Circuit(connections)

fig, (ax1,ax2) = plt.subplots(2, 1, sharex=True)
C.network.plot_s_db(ax=ax1, m=0, n=0,  lw=2)  # S11
C.network.plot_s_db(ax=ax1, m=1, n=1,  lw=2)  # S22
ax1.set_ylim(-90, 0)
C.network.plot_s_db(ax=ax2, m=1, n=0,  lw=2)  # S21
[...]

PR #1028, opened 2024-02-17
fix n-port mismatch junction issue in Circuit

mhuser:
Fix issue #1022
It appears that formula (3) in P. Hallbjörner, Microw. Opt. Technol. Lett. 38, 99 (2003). does not work for mismatched intersections.
This was temporarily fixed for 2-connection intersections and is now generalized to n-connection intersections with mismatched port impedances.
- [x] Fix circuit
- [x] Optimize
- [x] Update test_circuit 
- [x] Add mismatched splitter to media and a test case

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -711,8 +711,8 @@ class Circuit:
         Xs : :class:`numpy.ndarray`
             shape `f x n x n`
         """
-        Xnn = self._Xnn_k(cnx_k)  # shape: (nb_freq, nb_n)
-        Xmn = self._Xmn_k(cnx_k)  # shape: (nb_freq, nb_n)
+        # Xnn = self._Xnn_k(cnx_k)  # shape: (nb_freq, nb_n)
+        # Xmn = self._Xmn_k(cnx_k)  # shape: (nb_freq, nb_n)
         # # for loop version
         # Xs = []
         # for (_Xnn, _Xmn) in zip(Xnn, Xmn):  # for all frequencies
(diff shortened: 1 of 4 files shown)


skrf/circuit.py
```diff
@@ -710,30 +628,20 @@ def _Xk(self, cnx_k: list[tuple]) -> np.ndarray:
...
-        # return Xs.s
+        
+        y0s = np.array([1/ntw.z0[:,ntw_port] for (ntw, ntw_port) in cnx_k]).T
+        y_k = y0s.sum(axis=1)
+
+        Xs = np.zeros((len(self.frequency), len(cnx_k), len(cnx_k)), dtype='complex')
+        
+        Xs = 2 *np.sqrt(np.einsum('ij,ik->ijk', y0s, y0s)) / y_k[:, None, None]       
+        np.einsum('kii->ki', Xs)[:] -= 1  # Sii
+        return Xs
```
jhillairet, line 644: Since you've implemented `splitter`, eventually `_Xk` could use directly it. However, maybe a price of a loss of performances... Up to you.

mhuser:
I wrote the equations in [`Media.splitter`](https://scikit-rf--1028.org.readthedocs.build/en/1028/api/media/generated/skrf.media.Media.splitter.html#skrf.media.Media.splitter) doc as a reminder.

jhillairet (approved):
Great work @mhuser !

jhillairet:
It's no more relevant, but for info here is another implementation of the S-matrix renormalization from a given Z0 to Z0'. The numerical precision is better than the current implementation using `s2z(z2s())` but it only works for power waves.

```python
def renormalize_s2(
        s: npy.ndarray, z_old: NumberLike, z_new: NumberLike,
        ) -> npy.ndarray:
    """
    Renormalize a s-parameter matrix given old and new port impedances.
    
    Use equations from [#Reveyrand]_. 
    The numerical precision of this method is better than `renormalize_s`, 
    however, it only supports the power-wave definition.
   
    Parameters
    ----------
    s : complex array of shape `fxnxn`
        s-parameter matrix

    z_old : complex array of shape `fxnxn` or a scalar
        old (original) port impedances

    z_new : complex array of shape `fxnxn` or a scalar
        new port impedances

    Returns
    -------
    :class:`numpy.ndarray`
        renormalized s-parameter matrix (shape `fxnxn`)

    See Also
    --------
    Network.renormalize : method of Network to renormalize s
    renormalize_s
    fix_z0_shape
    s2s

    References
    ----------
    .. [#Reveyrand] T. Reveyrand, "Multiport conversions between S, Z, Y, h, 
        ABCD, and T parameters’, in 2018 International Workshop on Integrated 
        Nonlinear Microwave and Millimetre-wave Circuits (INMMIC)"",
        Brive La Gaillarde: IEEE, Jul. 2018, pp. 1–3.
        doi: 10.1109/INMMIC.2018.8430023.

    """
    nfreqs, nports, nports = s.shape
    z_old = fix_z0_shape(z_old, nfreqs, nports)
    z_new = fix_z0_shape(z_new, nfreqs, nports)
   
    Id = npy.zeros_like(s, dtype='complex')
    R = npy.zeros_like(s, dtype='complex')
    Gold = npy.zeros_like(s, dtype='complex')
    Gnew = npy.zeros_like(s, dtype='complex')
    
    npy.einsum('ijj->ij', Id)[...] = 1
    npy.einsum('ijj->ij', R)[...] = (z_new - z_old)/(z_new + z_old.conj())
    npy.einsum('ijj->ij', Gold)[...] = 1/npy.sqrt(npy.abs(z_old.real))
    npy.einsum('ijj->ij', Gnew)[...] = 1/npy.sqrt(npy.abs(z_new.real))

    A = npy.linalg.inv(Gnew) @ Gold @ (Id - R.conj())

    return npy.linalg.inv(A) @ (s - R.conj()) @ npy.linalg.inv(Id - R@s) @ A.conj()
```

skrf/circuit.py
mhuser: Indeed, the use of a splitter makes very clear the connection purpose. But I think here it is however interesting to spare the `media` and `splitter` instances. I also thought of a `splitter_s` helper function used by both `media` and `circuit`, but finally it is also interesting to see directly in each class the computation that is used.

mhuser:
> It's no more relevant, but for info here is another implementation of the S-matrix renormalization from a given Z0 to Z0'. The numerical precision is better than the current implementation using `s2z(z2s())` but it only works for power waves.

Very interesting. Maybe we could use this in `renormalize_s` and fallback to `z2s(s2z()` if `s_def` is not power-wave?
We know that current `renormalize_s` is weak when transformation between $S$ and $Z$ would yeld the inverse of a singular matrix.

jhillairet:
> > It's no more relevant, but for info here is another implementation of the S-matrix renormalization from a given Z0 to Z0'. The numerical precision is better than the current implementation using `s2z(z2s())` but it only works for power waves.
> 
> Very interesting. Maybe we could use this in `renormalize_s` and fallback to `z2s(s2z()` if `s_def` is not power-wave? We know that current `renormalize_s` is weak when transformation between S and Z would yeld the inverse of a singular matrix.

Eventually... Let's see that in another PR.

PR #1030, opened 2024-02-19
Remove some unused code

FranzForstmayr:


skrf/calibration/calibration.py
```diff
@@ -6429,14 +6421,12 @@ def convert_12term_2_8term(coefs_12term, redundant_k = False):
     Erf = coefs_12term['forward reflection tracking']
     Etf = coefs_12term['forward transmission tracking']
     Elf = coefs_12term['forward load match']
-    Eif = coefs_12term.get('forward isolation',0)
```
jhillairet, line 6432: I guess this one and Eir should be kept for reference.

skrf/network.py
```diff
@@ -845,7 +845,6 @@ def __getitem__(self, key: str | int | slice | Sized) -> Network:
         >>> d = b.s12
         """
 
-        a = self.z0  # HACK: to force getter for z0 to re-shape it (z0 getter has side effects...)
```
jhillairet, line 848: keep it?

FranzForstmayr: we removed the side effects in z0's getter, so it's not needed anymore.

jhillairet (approved):

PR #1033, opened 2024-02-26
Automatic vector fitting using pole adding and skimming algorithm

Vinc0110:
It took a while, now I finally managed to get it working: a fully automatic version of `VectorFitting.vector_fit()`, called `VectorFitting.autofit()`.

The implementation is based on the "vector fitting with adding and skimming" algorithm from this paper: 

> S. Grivet-Talocia and M. Bandinu, "Improving the convergence of vector fitting for equivalent circuit extraction from noisy frequency responses," in IEEE Transactions on Electromagnetic Compatibility, vol. 48, no. 1, pp. 104-120, Feb. 2006, DOI: https://doi.org/10.1109/TEMC.2006.870814

Advantages over regular vector fitting:
- No need to provide the number of poles, which can sometimes be very challenging. So this is very user friendly.
- The automatically fitted model can be very close to the optimum in terms of model order (lowest number of poles) and model error.
- Huge improvement in the fitting of noisy data, both in terms of model order and rate of convergence.
- Large number of optional parameters controlling the algorithm, so users can still have a lot of control and optimize for their data.

There is still a lot of testing to do, but here is a first comparison of the `Agilent_E5071B.s4p`:
- `vector_fit(1, 26): model order = 53, rms error = 0.0090`
- `autofit(): model order = 59, rms error = 0.0059`

agilent_vectorfit_1-26

agilent_autofit


Vinc0110:
Another comparison with `190ghz_tx_measured.S2P`:
- vector_fit(4, 4): model order = 12, rms error = 0.0147
- autofit(): model order = 15, rms error = 0.0140

190ghz_vectorfit_4-4

190ghz_autofit

Vinc0110:
I renamed it to `auto_fit()` to be more consist with `vector_fit()`. 

@jhillairet The code is ready now. I might add a dedicated unit test later, but feel free to merge this PR if you want to proceed with the release of v.0.32.0. Reviews are obviously welcome, but I realize the details are somewhat tricky.

Vinc0110:
What do you guys think about the return values from `auto_fit()`? Is it a good thing or is it annoying? 
The `vector_fit()` does not return anything and keeps those values stored in the respective class attributes. The are not fully compatible, but I can of course add some attributes to be filled only by `auto_fit()`...

jhillairet (approved):
LGTM. Nice documentation addition.

jhillairet:
> What do you guys think about the return values from `auto_fit()`? Is it a good thing or is it annoying? The `vector_fit()` does not return anything and keeps those values stored in the respective class attributes. The are not fully compatible, but I can of course add some attributes to be filled only by `auto_fit()`...

Well, ideally it seems better to me if the behaviour of these two functions are similar.

Vinc0110:
Ok, thanks! I decided to remove the return values altogether, as I'm not even sure they are really that useful.
So, I think I'm finished here for now.

jhillairet (approved):

PR #1037, opened 2024-03-03
8510 Fixes

jjoonathan:
Hi!

A few years back, I contributed an 8510 class for doing compound sweeps, but it happened at a bad time right before a refactor of the VNA classes. Life got in the way and I disappeared before I finished updating the code. Apologies for disappearing on you like that, but now I've got a new project with the 8510 and with it renewed time/motivation to work on this.

I am currently trying to figure out what VNA interface I should adapt my code to in order to maximize longevity. In a few months I will no longer have access to the 8510, so I would like to future-proof as much as practical. I started with the hp8510c.py I currently see in the project, but it has a number of basic problems that suggest it was written without the benefit of an 8510 to test against, so perhaps it is incomplete in other ways and not the best reference. I'd appreciate guidance on what to use as a "compatibility checklist/reference" for updating my code.

These patches suffice to  fix HP8510C get_snp_network([1]) and get_snp_network([1,2]), but they do spread some 8510-isms into places that might not want the complexity. For example, number of points is a string like '  2.01000000000E+02     \n', and this patchset contains updates to VNA.command to deal with that but I tend to suspect one-offing it in the 8510 class might be better from the view of isolating complexity. Is VNA.command important for metaprogramming somewhere or could I just define my own properties where convenient?

### DIFF
--- a/skrf/vi/validators.py
+++ b/skrf/vi/validators.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any, Sequence
+    from typing import Any, Sequence, Callable
 
 import re
 from abc import ABC, abstractmethod
@@ -116,7 +116,7 @@ class FreqValidator(Validator):
 
         else:
             try:
-                return int(arg)
+                return int(float(arg))
             except ValueError as err:
                 raise ValidationError("Could not convert {arg} to an int") from err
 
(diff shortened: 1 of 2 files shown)


jjoonathan:
Whoops, I forgot to fix up the mocks -- if we decide this is the right approach after all I'll go back and do that.

jhillairet:
Thank you @jjoonathan 

@cafeclimber, could you have a look and help?

jjoonathan:
Closing this -- I've been using a modernized version of my old class instead, I'll probably open a new pull request to contribute it this weekend.

PR #1040, opened 2024-03-09
Updated 8510 class with bugfixes and compound sweep support

jjoonathan:
Ok, I ported my old but working HP8510 code into the shell of the new, not-working HP8510 class, fixed the bugs, added some tests, and moved the compound sweep class into the hp subfolder where it ought to be.

There are fundamental problems with freq_start/freq_stop API on the 8510; it doesn't register the updated values immediately, so anything that explicitly or implicitly reads back freq_start/freq_stop without sleeping will incorrectly receive old limits. My examples steer people towards set_frequency_sweep which does not suffer from this problem, but I'd almost rather get rid of the old API or add implicit sleeps to make it correct but slow. How important is it to support freq_start/freq_stop?


jjoonathan:
Hmm, odd that 3.9 would fail but 3.7, 3.8, 3.10, 3.11, and 3.12 would succeed -- and with an import error in vna.py, which I used in a test but didn't otherwise touch:

`  skrf/vi/vna/vna.py:23: in <module>
      import pyvisa
  E   ModuleNotFoundError: No module named 'pyvisa'`
  
  any ideas what that could be about?

jjoonathan:
Huh. The pyvisa dependencies look the same between working/broken/working environments according to `Test the code, tutorials and examples` > `python -m pip freeze --all`

`py38:PyVISA==1.14.1,PyVISA-py==0.7.2`

`py39: PyVISA==1.14.1,PyVISA-py==0.7.2`

`py310: PyVISA==1.14.1,PyVISA-py==0.7.2`

jjoonathan:
I wonder if some kind of PATH is overflowing somewhere. I still don't understand why it's 3.9 only though

```
 ==================================== ERRORS ====================================
  ________ ERROR collecting skrf/vi/vna/hp/tests/test_8510c_sweep_plan.py ________
  ImportError while importing test module '/home/runner/work/scikit-rf/scikit-rf/skrf/vi/vna/hp/tests/test_8510c_sweep_plan.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  /opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/importlib/__init__.py:127: in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
  skrf/vi/vna/hp/tests/test_8510c_sweep_plan.py:3: in <module>
      from skrf.vi.vna.hp.hp8510c_sweep_plan import SweepSection, SweepPlan
  skrf/vi/vna/__init__.py:180: in <module>
      from .vna import VNA, Channel, ValuesFormat # isort: skip
  skrf/vi/vna/vna.py:23: in <module>
      import pyvisa
  E   ModuleNotFoundError: No module named 'pyvisa'
  ```

**Comment by jjoonathan** [comment]
Here's a minor documentation fix with the real intent of making the tests run again to see if the problem magically fixes itself lol

**Comment by jjoonathan** [comment]
So much for that, but I think I see the real problem now: there is some kind of dependency minimization that runs in the 3.9 test environment but not 3.8/3.10/etc. Looking into it.

**Comment by jjoonathan** [comment]
Ok looks like that was it but I feel like the correct solution might be to make the test compatible with minimal-dependencies rather than removing minimal-dependencies from the 3.9 environment to make it like the 3.7/3.8/3.10/3.11/3.12 environments.

**Comment by jjoonathan** [comment]
Ok that was a bit painful but now I understand how to not provoke the minimal-dependencies test environment :)

**Comment by jhillairet** [comment]
@cafeclimber could you have a look?

**Comment by jjoonathan** [comment]
Yeah I'd appreciate the code review. As for testing, I have been using this code to take a bunch of measurements that lean heavily on its ability to stitch multiple sweeps. The results have been good -- comparable to but lower noise than the FieldFox, even with bandwidth at 1kHz: https://github.com/jjoonathan/cal-std-8510-fieldfox/blob/main/quick_check.ipynb

**Comment by cafeclimber** [comment]
Hey sorry for the delay. Wrapping up PhD work is taking a lot of my focus atm. Taking a look now!

File: skrf/vi/vna/hp/hp8510c.py [hunk 1]
```diff
@@ -1,187 +1,419 @@
...
+Advanced example.
+>>> vna = skrf.vi.vna.hp.HP8510C('TCPIP::ad007-right.lan::gpib0,16::INSTR')
+>>> freq_block_1 = np.linspace(1e9,2e9,801)
+>>> freq_block_2 = [10e9,11e9,12e9]
+>>> freqs = np.concatenate((freq_block_1, freq_block_2))
+>>> vna.frequency = skrf.Frequency.from_f(freqs)
+>>> vna.get_snp_network(ports=(1,2))
+"""
+
+import numpy as npy
```
cafeclimber, line 39: @jhillairet I just noticed, it seems some parts of skrf use `npy` and some parts use `np`. Worth it to standardize on one or the other?

skrf/vi/vna/hp/hp8510c_sweep_plan.py
```diff
@@ -0,0 +1,248 @@
+import dataclasses
+import numpy as np
+from typing import List, Union
+from abc import ABC, abstractmethod
+
+
+class SweepSection(ABC):
+    @abstractmethod
+    def get_hz(self):
```
cafeclimber, line 9: Again if you could add type hints :)

skrf/vi/vna/hp/hp8510c_sweep_plan.py
```diff
@@ -0,0 +1,248 @@
...
+    def apply_8510(self, hp8510c):
+        hp8510c._set_instrument_step_state(self.hz_min, self.hz_max, self.n_points)
+
+
+@dataclasses.dataclass
+class LinearMaskedSweepSection(SweepSection):
+    hz_min: float
+    hz_max: float
+    n_points: int
+    mask: object
```
cafeclimber, line 43: Is there a more specific type than `object`?

cafeclimber (changes requested):
I think this looks great! Very thoroughly written and tested. Just a few nit-picky things about types and some formatting.

skrf/vi/vna/hp/hp8510c.py
jhillairet: The `npy` notation is historic. I guess we should now favor `np`. What do you think @FranzForstmayr ? We could homogeneize.

FranzForstmayr: I think we should go for `np`, but this could be a batch PR for all files.

jjoonathan: I personally prefer `np` and was only using `npy` to rhyme with other code I saw, so I would be happy to change the PR -- and consider this a vote for global `np` too :)

skrf/vi/vna/hp/hp8510c.py
```diff
@@ -1,187 +1,419 @@
...
+
+class HP8510C(VNA):
+    '''
+    8510 driver that is capable of compound sweeps.
+    '''
+    min_hz = None
+    max_hz = None
+    compound_sweep_plan = None
+
+    def __init__(self, address="TCPIP::ad007-right.lan::gpib0,16::INSTR", backend='@py', **kwargs):
```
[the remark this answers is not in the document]
jjoonathan: Yep, I got rid of the setup-specific default and made it rhyme with super().__init__

skrf/vi/vna/hp/hp8510c.py
```diff
@@ -1,187 +1,419 @@
...
+        super(HP8510C, self).__init__(address, backend=backend, **kwargs)
```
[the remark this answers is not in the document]
jjoonathan: Sounds good, updated.

skrf/vi/vna/hp/hp8510c.py
```diff
@@ -1,187 +1,419 @@
...
+        # Tested 2024-03-09:
+        #     HP8510C.07.14
+        #     AD007 (a VXI11-complaint GPIB-Ethernet adapter)
+        #     address="TCPIP::ad007-right.lan::gpib0,16::INSTR", backend="@py"
+
+        # 8510s are slow. This check ensures we won't wait 60s for connection error.
+        self._resource.timeout = 2_000 #set_visa_attribute(pyvisa.constants.ResourceAttribute.timeout_value, 2000)
```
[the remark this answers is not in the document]
jjoonathan: Yeah that was just the first thing I tried that didn't work, it's past time I removed the stale code. Done.

skrf/vi/vna/hp/hp8510c.py
```diff
@@ -1,187 +1,419 @@
...
+            if trace.upper()=='S11':
+                trace_out = self.s11
+            elif trace.upper()=='S22':
+                trace_out = self.s22
+            elif trace.upper()=='S12':
+                trace_out = self.s12
+            elif trace.upper()=='S21':
+                trace_out = self.s21
+            else:
+                raise(ValueError(trace+" is not a valid trace. Options: "+' '.join(self.get_list_of_traces())))
```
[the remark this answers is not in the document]
jjoonathan: Yep, happy to update it. I vaguely remembered that in 2022 maybe some of the old unit tests ran on a really old python that didn't support f-strings or something, but even if that's still the case it's worth a try to update. I'd like to use f-strings if available.

skrf/vi/vna/hp/hp8510c.py
jjoonathan: PR updated (but just for my code, I didn't rope in the scope of a global search and replace)

skrf/vi/vna/hp/hp8510c_sweep_plan.py
jjoonathan: Yep, no prob

skrf/vi/vna/hp/hp8510c_sweep_plan.py
jjoonathan: Whoops I forgot that one on my first commit, second commit incoming.

jjoonathan:
Hmm I can't see the HP8510C class in the docs, do I need to add it somewhere to make that happen?

jhillairet:
> Hmm I can't see the HP8510C class in the docs, do I need to add it somewhere to make that happen?

Indeed, you may have to update https://github.com/scikit-rf/scikit-rf/blob/master/skrf/vi/vna/__init__.py

skrf/vi/vna/hp/__init__.py
```diff
@@ -1 +1,10 @@
+"""
+.. module:: skrf.vi.vna.hp
+===========================================
+HP VNAs (:mod:`skrf.vi.vna.hp`)
+===========================================
+
+From the age when HP made expensive test equipment rather than expensive printer ink.
```
jhillairet, line 7: The comment is OK to me ;  I let others say something eventually ;)

Vinc0110: I love it!

FranzForstmayr:
> Hmm, odd that 3.9 would fail but 3.7, 3.8, 3.10, 3.11, and 3.12 would succeed -- and with an import error in vna.py, which I used in a test but didn't otherwise touch:
> 
> `  skrf/vi/vna/vna.py:23: in <module>
>       import pyvisa
>   E   ModuleNotFoundError: No module named 'pyvisa'`
>   
>   any ideas what that could be about?

Unfortunately I didn't have a look earlier, but I can explain now. 
We want to enable the the majority of scikit-rf's features to users who don't have pyvisa or matplotlib installed. That's why unit test without instrument interface or plotting capability run in a specific minimal unit test as well.

skrf/vi/vna/hp/__init__.py
jjoonathan: Well I had to find something to say about HP in general, lol. I'm afraid this will be hidden for now -- it looks like Sphinx doesn't generate module docs unless they are linked elsewhere and given that we have only 1.3 VNAs per manufacturer at the moment I think it's probably best to keep the skrf.vi.vna page as a list of classes right now. One day, though, there will be enough VNAs to break out the categories, and on that day I hope to contribute to the archaeological record about HP :)

jjoonathan:
Thanks for explaining, that makes good sense.

jjoonathan:
Ok I'm pretty happy with the state of both the documentation and code here. Does anyone have final thoughts or is it time to think about merging?

jhillairet (approved):
Fine to me.

skrf/vi/vna/hp/hp8510c_sweep_plan.py
```diff
@@ -0,0 +1,245 @@
+import dataclasses
+import numpy as np
+import skrf
+from typing import List, Union
```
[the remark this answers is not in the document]
jjoonathan: Ruff is pretty slick, good tip :)

jjoonathan:
Ok, I made a bunch of lint fixes and tested against the physical VNA again. Still works :+1:

jhillairet (approved):

FranzForstmayr (approved):

Vinc0110 (approved):

PR #1041, opened 2024-03-09
Advanced Smith Chart examples with markers or background

mhuser:
Provides two advanced Smith Chart examples:

# Advanced Smith Chart example with markers

The Smith Chart is a convenient plot that shows complex numbers with real and imaginary parts from zero to infinity with a zoom on a real reference in the center region. It is, however, not straightforward to figure out what point is related to which frequency because there is no frequency axis. A common workaround is to provide some frequency markers related to a table.

Example1

# Advanced Smith Chart example with background

The Smith Chart can be drawn above a high-resolution Smith Chart background (or another image upon your fantasy).

Example2


mhuser:
[Documentation preview](https://scikit-rf--1041.org.readthedocs.build/en/1041/tutorials/Plotting.html#Advanced-Smith-Chart-with-markers)

jhillairet:
Cool, nice to add the solution you have provided in #1039 to the doc!

jhillairet (approved):

PR #1044, opened 2024-03-12
Run CI against nightly as well

FranzForstmayr:
As `numpy` 2.0 is nearly I decided to run some tests against this new version.


jhillairet (approved):

jhillairet:
[The latest run failed ](https://github.com/scikit-rf/scikit-rf/actions/runs/8684979110/job/23813526669?pr=1053) with the nightly version:
`
  ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
  pyarrow 15.0.2 requires numpy<2,>=1.16.6, but you have numpy 2.1.0.dev0 which is incompatible.
`

FranzForstmayr:
Will check end of this week

FranzForstmayr:
I just reran the tests and everything worked...
The deviation in result log is really small and indicates a numeric isssue, but I'm not sure as we're doing basically the same operation twice.

PR #1045, opened 2024-03-17
CI speedup

jhillairet:
Uses [pytest-xdist](https://pytest-xdist.readthedocs.io/en/stable/), a [pytest](https://docs.pytest.org/en/8.0.x/) extension that allows tests to be run in parallel. Tests are run on the available processors with `pytest -n auto`. GitHub workflow currently uses up to two processes, providing a significant improvement. Locally the speedup is even more  significant.

I've tried to use [uv](https://github.com/astral-sh/uv), a recent replacement for pip, pip-tool, and virtualenv using the [tox-uv](https://github.com/tox-dev/tox-uv) extension. However the latest requires Python >= 3.8, which exclude Python 3.7 currently supported by scikit-rf.


skrf/mathFunctions.py
```diff
@@ -267,7 +267,7 @@ def magnitude_2_db(z: NumberLike, zero_nan: bool = True):
     """
     out = 20 * npy.log10(z)
     if zero_nan:
-        return npy.nan_to_num(out, nan=LOG_OF_NEG, neginf=-npy.inf)
+        return npy.nan_to_num(out, copy=False, nan=LOG_OF_NEG, neginf=-npy.inf)
```
FranzForstmayr, line 270: You need to remove `copy=False` for numpy 2 compatibility :)

jhillairet: Yeah, I failed my merging...

jhillairet: OK now. The on hp8510c.py comes from pre-commit/ruff

PR #1055, opened 2024-04-20
Move apps to a seprate repo

jhillairet:
Move qt-apps and traitsui-apps to the new repository [skrf-apps](https://github.com/scikit-rf/skrf-apps)


FranzForstmayr:
CI Failed again once (before restarting), I think I need to have a look again tomorrow.

PR #1061, opened 2024-05-02
Change all occurences of npy to np

cafeclimber:
Per the discussion in #1040 , convert all instances of npy to np

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -96,7 +96,7 @@ from numbers import Number
 from textwrap import dedent
 from warnings import warn
 
-import numpy as npy
+import numpy as np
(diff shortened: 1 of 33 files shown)


jhillairet (approved):
Right, good idea, thanks. 

The py12 linter fails for some reason (?), but to me the PR ok.

cafeclimber:
Yeah I'm not sure. All the tests pass so hm...

FranzForstmayr:
ruff reports some errors when executing locally, that's why the test fails. However it's confusing that the errors are not shown properly. I'll check in the next few days locally.

cafeclimber:
Seems like those changes addressed the linting issues! (Just waiting for the rtd task to finish)

PR #1065, opened 2024-05-06
Make most modules lazy load

### LINKED ISSUE #1057: Importing scikit-rf imports many modules and is slow
Issue created following [this post on reddit](https://www.reddit.com/r/rfelectronics/comments/1cf9zu2/comment/l1suprm/).

The number of modules imported in the namespace when importing scikit-rf is quite large and can take quite some times. 

Just importing `skrf.constants` imports PIL, scipy, matplotlib, pandas, and all of the skrf modules, and generally takes about 1 second to complete (where numpy takes around 150 ms). 

Here's an example showing the difference between numpy import and skrf imports. First, here's numpy as a baseline.

# numpy
```python
from time import perf_counter
import sys

tic = perf_counter()
from numpy import array

print ("numpy import: ", perf_counter() - tic)
print (list(sorted(sys.modules.keys())))
```
numpy import: 0.13497460004873574

[List of imported modules](https://pastebin.com/jCeXk5V7).

# scikit-rf

```python
from time import perf_counter
import sys

tic = perf_counter()
from skrf.constants import c

print ("SKRF import: ", perf_counter() - tic)

print (list(sorted(sys.modules.keys())))
```

With the result:

SKRF import: 1.014745100052096

[List of imported modules](https://pastebin.com/hfM2MK0e)

cafeclimber:
Addresses #1057. On my machine, import times were reduced from over 1 second to about 0.3 seconds.


jhillairet (approved):

cafeclimber:
Ok I think I super broke this PR trying to pull in the most recent changes. Happy to fix just lmk!

FranzForstmayr:
would be great if possible :)

cafeclimber:
> would be great if possible :)

Okay I think that did it lol. Just gonna wait for the CI

FranzForstmayr:
I think most of the linter issues can be fixed with `ruff check . --fix`

cafeclimber:
Hmmmm...I'm not sure why only 3.9 is failing. Specifically something about the matplotlib import

FranzForstmayr:
I think all the calibration Test should run without matpotlib installed but it's imported somewhere.

cafeclimber:
Ah it was me being a dingus and not understanding how lazy_loader deals with missing imports. Seems ok now?

cafeclimber:
Ah dang this seems like a deep problem. Matplotlib is easily the most time consuming import. At worst a fix might be to just move plotting imports to those specific functions that use them to avoid issues associated with using lazy_loader with mpl

skrf/__init__.py
```diff
@@ -3,10 +3,9 @@
 implemented in Python.
 """
 
-__version__ = '1.0.0'
+__version__ = "1.0.0"
```
Vinc0110, line 6: Out of curiosity: why the change from ' to " for the strings? I always thought Python does accept both notations. Is it personal taste or a new general rule?

cafeclimber: That's almost certainly ruff's auto formatting. I try to avoid noisy commits that are primarily changes due to auto formatting but sometimes stuff slips by 😮‍💨

cafeclimber: That's almost certainly ruff's auto formatting. I try to avoid noisy commits that are primarily changes due to auto formatting but sometimes stuff slips by 😮‍💨

FranzForstmayr:
Seems like the root cause:
https://github.com/scientific-python/lazy_loader/issues/55

`matplotlib` is currently not compatible with `lazy_load`

cafeclimber:
Yeah I think the solution is to try to move all mpl-related imports into the methods that use them so they get deferred until then. Will be a bit clunky but the speed up is significant. I'll try to push something today

jhillairet:
How about this PR ? Would be nice to include it in the next version, to be released soon.

cafeclimber:
Hey sorry got caught up with dissertation work. Will make edits tonight to remove the lazy load module and move mpl imports into the respective plotting functions. Think that's the best we can do

jhillairet:
Ok. Should we merge it like that?

luar123:
This would be a great improvement. Is there something blocking from merging?

jhillairet:
> This would be a great improvement. Is there something blocking from merging?

some conflicts to address to start with... Otherwise I think it was OK.

luar123:
I tried to merge with current master, but the improvement is limited for two reasons:
1. matplotlib is always imported when installed: https://github.com/scikit-rf/scikit-rf/blob/28b032ee26dbf5b7e274f250b0e0081a94208b99/skrf/__init__.py#L102
And btw. `plotting_available `is always `None`
2. Many modules in media and constants import scipy.constants which has a lot of dependencies.

Importing skrf on my windows pc:
1. current master: 3.1s
2. updated PR: 2.5s
3. 2 + disable matplotlib: 1.5s
4. 3 + move imports from scipy.constants into functions (these are a lot): 0.6s

(numbers are difficult as there seems to be some caching)

Not sure how to proceed here. Applying everything definitely helps a lot, but would change the interface. Users would need to call `setup_plotting()` manually and `from skrf.constants import c` would not work anymore (not sure if this is a use case).

jhillairet:
> Not sure how to proceed here. Applying everything definitely helps a lot, but would change the interface. Users would need to call `setup_plotting()` manually

I'm not a super fan of this. Many users use scikit-rf for easy plotting, so adding one more step is not a good idea in my opinion. 

> and `from skrf.constants import c` would not work anymore (not sure if this is a use case).

For sure, it is a use case, so I'd prefer to keep it.

luar123:
> I'm not a super fan of this. Many users use scikit-rf for easy plotting, so adding one more step is not a good idea in my opinion.

I guess there could be a solution by calling `setup_plotting()` implicitly with the first use of a plotting function. And of course if matplotlib is not needed and not installed it is also no issue.

> > and `from skrf.constants import c` would not work anymore (not sure if this is a use case).
> 
> For sure, it is a use case, so I'd prefer to keep it.

I see, also it can be found in the api documentation. I looked into this a little bit more and I think we could lazy-load the skrf.constants module (potentially also other subpackages/modules) but only if this part is removed from __init__.py:
https://github.com/scikit-rf/scikit-rf/blob/fd203766ade139573f36674e726aff8bdf2c5f72/skrf/__init__.py#L30
This would still allow to do `from skrf.constants import c` but you can't do `import skrf as rf; rf.c` anymore.
In general, I see this whole part as problematic:
https://github.com/scikit-rf/scikit-rf/blob/fd203766ade139573f36674e726aff8bdf2c5f72/skrf/__init__.py#L28-L43
After `import skrf as rf`, the `rf` namespace contains all kind of stuff, you can do `rf.sys`, `rf.pd`, `rf.plt`, `rf.skrf_plt`, .... and often it is not clear where it comes from. For example you cannot import the skrf.io subpackage with `import skrf.io` or `from skrf import io` because it is shadowed by pythons io module imported in network.py.
My suggestion would be to remove that part entirely and only import stuff explicitly that is needed and documented like Network, Circuit, Frequency,.... As an alternative we could use `__all__`. This change might break some code that relies on undocumented features, though, but with the current approach every module level change (new function, imports) in any module could potentially break something. Once this is done lazy-loading could be used for subpackages and modules.

jhillairet:
> I guess there could be a solution by calling `setup_plotting()` implicitly with the first use of a plotting function. And of course if matplotlib is not needed and not installed it is also no issue.

Indeed, eventually. 
 
> In general, I see this whole part as problematic:
> 
> https://github.com/scikit-rf/scikit-rf/blob/fd203766ade139573f36674e726aff8bdf2c5f72/skrf/__init__.py#L28-L43
> 
> After `import skrf as rf`, the `rf` namespace contains all kind of stuff, you can do `rf.sys`, `rf.pd`, `rf.plt`, `rf.skrf_plt`, .... and often it is not clear where it comes from. For example you cannot import the skrf.io subpackage with `import skrf.io` or `from skrf import io` because it is shadowed by pythons io module imported in network.py.

You're right, various additions have been made during the life of the project without "great plan", resulting in this ;)

> My suggestion would be to remove that part entirely and only import stuff explicitly that is needed and documented like Network, Circuit, Frequency,.... As an alternative we could use `__all__`. This change might break some code that relies on undocumented features, though, but with the current approach every module level change (new function, imports) in any module could potentially break something. Once this is done lazy-loading could be used for subpackages and modules.

OK, why not. If you want to tackle this?

PR #1073, opened 2024-05-21
Add valid case to convert_8term_2_12term function where switch_terms …

floschl:
…are not available

The original motivation for this is to perform a 8-term-to-12-term conversion for e.g. the UnknownThru calibration in the case that no switch terms are provided. In this case gamma_f and gamma_r are zero, leading to a division by zero and thus Etf/Etr being NaN.

This extension covers the implementation as defined in [1], eqs. (36)-(39), where the conversion is conducted without the switch terms being provided.

The current test cases do not cover this, as random switch terms are always provided. The test coverage could extended to another set of tests with no switch terms provided. This would then cause the convert_8term_2_12term to return zero Etf/Etr, which is wrong.

Up to you, if we should also extend the tests in this regard.

Best regards
Florian

[1] Marks, Roger B. 1997. ‘Formulations of the Basic Vector Network Analyzer Error Model Including Switch-Terms’. In 50th ARFTG Conference Digest, 32:115–26. https://doi.org/10.1109/ARFTG.1997.327265.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -6472,24 +6472,35 @@ def convert_8term_2_12term(coefs_8term):
     k_first = coefs_8term.get('k first', k)
     k_second = coefs_8term.get('k second', k)
 
-    # taken from eq (36)-(39) in the Roger Marks paper given in the
-    # docstring
-    Elf  = Esr + (Err*gamma_f)/(1. - Edr * gamma_f)
-    Elr = Esf  + (Erf *gamma_r)/(1. - Edf  * gamma_r)
-    Etf  = ((Elf  - Esr)/gamma_f) * k_first
-    Etr = ((Elr - Esf )/gamma_r) * 1./k_second
-
     coefs_12term = {}
-    for l in ['forward directivity','forward source match',
-        'forward reflection tracking','reverse directivity',
-        'reverse reflection tracking','reverse source match',
-        'forward isolation', 'reverse isolation']:
-        coefs_12term[l] = coefs_8term[l].copy()
+
+    if True or (not (np.allclose(gamma_f, np.zeros_like(gamma_f))\
+            and np.allclose(gamma_r, np.zeros_like(gamma_r)))):
+        # taken from eq (36)-(39) in the Roger Marks paper given in the
+        # docstring
+        Elf  = Esr + (Err*gamma_f)/(1. - Edr * gamma_f)
+        Elr = Esf  + (Erf *gamma_r)/(1. - Edf  * gamma_r)
+        Etf  = ((Elf  - Esr)/gamma_f) * k_first
+        Etr = ((Elr - Esf )/gamma_r) * 1./k_second
+    else:
+        # taken from eq (40)-(44) in the Roger Marks paper given in the
+        # docstring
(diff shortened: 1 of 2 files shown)


skrf/calibration/calibration.py
```diff
@@ -6472,24 +6472,35 @@ def convert_8term_2_12term(coefs_8term):
...
-    Etr = ((Elr - Esf )/gamma_r) * 1./k_second
-
     coefs_12term = {}
-    for l in ['forward directivity','forward source match',
-        'forward reflection tracking','reverse directivity',
-        'reverse reflection tracking','reverse source match',
-        'forward isolation', 'reverse isolation']:
-        coefs_12term[l] = coefs_8term[l].copy()
+
+    if True or (not (np.allclose(gamma_f, np.zeros_like(gamma_f))\
```
FranzForstmayr, line 6477: This is testing code I assume?

floschl: yes, sorry, it was a remnant of some testing. Removed and pushed it.

Ttl:
This should have a test case. For example, generate random error network with zero switch terms, convert it to 12-term and back and check for equality to the original error network.

Code would be slightly cleaner if `not` is removed and `if` statement branches swapped. There's also still possibility of division by zero if only some of the switch terms are zero but I'm not sure how significant that case is.

Ttl:
I don't see any changes. Did you forget to push?

floschl:
> I don't see any changes. Did you forget to push?

It is pushed now.

skrf/calibration/calibration.py
```diff
@@ -6472,24 +6472,35 @@ def convert_8term_2_12term(coefs_8term):
...
-
     coefs_12term = {}
-    for l in ['forward directivity','forward source match',
-        'forward reflection tracking','reverse directivity',
-        'reverse reflection tracking','reverse source match',
-        'forward isolation', 'reverse isolation']:
-        coefs_12term[l] = coefs_8term[l].copy()
+
+    if np.allclose(gamma_f, np.zeros_like(gamma_f))\
+            or np.allclose(gamma_r, np.zeros_like(gamma_r)):
```
Ttl, line 6478: If only one switch term is all zeros then the other is also treated as if it was zero resulting in incorrect result if that's not the case. Division by zero also remains if only some elements in switch terms are zero.

Ttl (changes requested):

skrf/calibration/calibration.py
floschl: Good morning,
I just pushed another suggestion: gamma_r and gamma_f are now checked independently and in the case of being all-zero the reduced equations are taken. That means that none of the coeffs (forward/reverse load match, transmission tracking) will be nan after the conversion from 8- to 12-terms.

Ttl: I don't see any new commits.

floschl: Sorry, again my SSH agent has removed my SSH key. 
I pushed it now and moreover added tests that cover all cases: Both switch terms are there, none, and either of the two).

Ttl (approved):

jhillairet (approved):

PR #1077, opened 2024-05-23
Impulse Response 

FranzForstmayr:
This PR aims to make skrf's impulse response calculation comparable to Metas VNA Tools from [here](https://www.metas.ch/metas/en/home/fabe/hochfrequenz/vna-tools.html).

The first test matches the band pass impulse response, remaining transformations are still open.


skrf/network.py
```diff
@@ -4070,33 +4078,30 @@ def impulse_response(self, window: str = 'hamming', n: int = None, pad: int = 10
...
-            center_to_dc = None
         if window is not None:
-            w = self.windowed(window=window, normalize=False, center_to_dc=center_to_dc)
+            w = self.windowed(window=window, normalize=False, center_to_dc=not bandpass)
         else:
             w = self
 
-        ir = mf.irfft(w.s, n=n)
+        if bandpass:
+            ir = np.fft.fftshift(np.fft.ifft(w.s, axis=0, n=n), axes=0)
```
mhuser, line 4093: I was wondering why it is `ifft` and not `irfft` here, but this is coherent with https://github.com/scikit-rf/scikit-rf/pull/900#issuecomment-1526207822

In bandpass mode, there is no complex conjugate assumption like in lowpass and the center frequency of the band is just treated like the 0Hz, leading to a complex result. This `ifft` for bandpass and `irfft for lowpass`.

skrf/frequency.py
```diff
@@ -580,14 +580,34 @@ def copy(self) -> Frequency:
         freq.unit = self.unit
         return freq
 
+    def _t_padded(self, *, pad: int = 0, n: int | None = None, bandpass: bool | None = None) -> np.ndarray:
```
mhuser, line 583: This is probably more clear than the `fftfreq` trick. Also it is a good idea to have all the time vector made in frequency.py so we are sure it is the same as for time gating.

mhuser (changes requested):
Great PR, still need a test for bandpass mode.

skrf/network.py
FranzForstmayr: Yes I remember the discussion back then, but I think the current implementation is not 100% correct, or at least my understanding back then :) 

See the last note entry here:
https://numpy.org/doc/stable/reference/generated/numpy.fft.irfft.html#numpy.fft.irfft

It's stated that the highest frequency of the `irfft` input has to be purely real, which is not the case in our application.
Usually we didn't recognise, as we use lot of padding by default. 

Metas seems to to this right, that's why I wanted to take them as reference.

mhuser: Yes, I think that the implementation you propose is correct. Also it is good to have the tests and reference data.

mhuser (approved):
Great

FranzForstmayr:
Will revert latest commit, autoformatter was on....

skrf/network.py
```diff
@@ -4193,7 +4194,7 @@ def step_response(
             )
 
         t, y = self.impulse_response(window=window, n=n, pad=pad, bandpass=False, squeeze=squeeze)
-        return t, np.cumsum(y, axis=0)
+        return t, cumulative_trapezoid(y, initial=0, axis=0)
```
mhuser, line 4289: Interesting. Does it gives a better agreement?

skrf/tests/test_network.py
```diff
@@ -162,6 +162,17 @@ def test_lpi(self):
             np.testing.assert_allclose(ref[:,0], t * 1e12, rtol=2e-5)
             np.testing.assert_allclose(ref[:,1], y, rtol=5e-5)
 
+    def test_lps(self):
+        path = Path(self.test_dir) / "metas_tdr"
+
+        for fname in ["short_10ps_dc_50g", "short_10ps_dc_40g"]:
```
mhuser, line 169: How where the reference.csv files generated? Are they directly supplied from metas or the TDR is made from .snp and a reference tool?

mhuser (approved):
Quality PR. I did goes trough the IEEEP370 exemples of TDR in the doc. The shape of the curves changes a bit but it looks good.

You can also try `window = None` instead of `window = 'boxcar'` to see the raw output (both methods should gives the same results but `None` will skip the windowing step and `boxcar` will multiply by ones but the windows is then explicit).

skrf/tests/test_network.py
FranzForstmayr: I generated the csv files with Metas VNA Tools.
https://www.metas.ch/metas/en/home/fabe/hochfrequenz/vna-tools.html

skrf/network.py
FranzForstmayr: So in fact I asked the guy behind Metas VNA Tools how they calculate the step response, now we have exact agreement.
In general using trapezoid integration is most of the time not worse than simple rectangle integration.

FranzForstmayr: Good point about the window, I checked manually that boxcar and None behave the same, but a test case would be fine as well

skrf/tests/test_network.py
mhuser: Sounds good! Should we explain it in a comment?

Issue #1010, opened 2024-01-23
add statics function i.e. violin plots for a number of sample measurements


yeney:
let's say we have 10 RF cables manufactured and measured 10 s2p files.
instead of plotting just 10 S21 curves in one plot (which is also very useful). we can use some statics plots to study if the samples follows gaussian distribution.
one of method draws my attention is Violin plot from either seaborn.violinplot or matplotlib.pyplot.violinplot.
the code will involve a resampling of the frequency (if too many frequency points in the original snp files) and rearrange values of the individual frequency to be the 10 "sample" 
if so then we could do different format of the study i.e dB, phase angle. Rm Im etc.
would this be of interest to add as a feature?
![1D Plot for All Data](https://github.com/scikit-rf/scikit-rf/assets/7814069/f361f26c-ee5e-4281-a4cb-f06b3dc0f939)
![Violin Plot](https://github.com/scikit-rf/scikit-rf/assets/7814069/08d78c5c-64c9-4ac2-a8ee-b9af268fb48a)

jhillairet:
Sure, this could be a great new feature! Of interest for many users I suppose.

PR #1083, opened 2024-06-04
Add Violin Plots

ericwrice:
Adds violin plots to NetworkSet as requested in #1010. It dynamically adds the methods in the same way as the existing plots. It also adds an interpolate_frequency method to NetworkSet to provide an easy way of reducing the number of frequency points, as the plot can get cluttered very quickly. I'm working on adding functionality for the time domain parameters as those are a special case. 

Here's the section added to the NetworkSet tutorial notebook:

ViolinPlotTutorial


jhillairet (approved):
LGTM.

skrf/networkSet.py
```diff
@@ -209,6 +210,8 @@ def __init__(self, ntwk_set: list | dict = None, name: str = None):
             self.__add_a_element_wise_method('plot_'+network_property_name)
             self.__add_a_element_wise_method('plot_s_db')
             self.__add_a_element_wise_method('plot_s_db_time')
+            self.__add_a_plot_violin(network_property_name)
```
FranzForstmayr, line 213: To be honest I don't like the autogenerated method stuff, although I'm aware that we used this way several times in this repository. It's hard to remove existing methods as we would break most existing code, but this isn't the case here.

These methods were not recogniced by any type checker and won't have autocompletion therefore.

I'd prefer to have _one_ method called `plot_method(self, ...)` with a similar attribute like this one?
https://github.com/scikit-rf/scikit-rf/blob/8c5729dd304a3d668c2b4058330cd8e85ac99fe1/skrf/network.py#L4661

@jhillairet what do you think?

skrf/plotting.py
```diff
@@ -1448,6 +1448,80 @@ def plot_minmax_bounds_component(self: NetworkSet, attribute: str, m: int = 0, n
...
+
+    freq = self.ntwk_set[0].f
+
+    # default widths to 3/4 distance between frequencies
+    if not widths and len(freq) > 1:
+        widths = (freq[1]-freq[0])*0.75
+    elif not widths:
+        widths = 0.5
+
+    data = np.array([p.__getattribute__(attribute)[:,m,n] for p in self.ntwk_set])
```
FranzForstmayr, line 1516:
```suggestion
    data = np.array([getattr(p, attribute)[:,m,n] for p in self.ntwk_set])
```

skrf/plotting.py
```diff
@@ -1448,6 +1448,80 @@ def plot_minmax_bounds_component(self: NetworkSet, attribute: str, m: int = 0, n
     scale_frequency_ticks(ax, ntwk_mean.frequency.unit)
     ax.axis('tight')
 
+@axes_kwarg
+def plot_violin_component(
+        self: NetworkSet,
+        attribute: str,
+        m: int = 0,
+        n: int = 0,
```
FranzForstmayr, line 1456:
```suggestion
        n: int = 0,
        *, 
```

The single star forces the users to pass all the following arguments keyword only, which is probably more self explaing.

ericwrice: I like this suggestion, going to apply it to the similar uncertainty bounds and minmax bounds methods as well. Also helped find a bug in all these methods where they couldn't accept any positional arguments because they would get passed into the "attribute" argument.

skrf/networkSet.py
ericwrice: I agree, the lack of autocomplete is inconvenient and it is difficult to follow for development. It's also not clear what all the possible methods are in the api documentation. For now, I just decided to use the existing convention for consistency but I'm definitely open to other ideas.

With the single method approach, wouldn't the user have to pass in the property and function argument and need to know what the possible options are? Maybe something like enums would be a good way to implement that?

jhillairet: > I'd prefer to have _one_ method called `plot_method(self, ...)` with a similar signature like the one in plotting.py
> 
> @jhillairet what do you think? If you are fine, I will not block :)

Sure, the lack of autocompletion is indeed inconvenient.What would you suggest for @ericwrice as a replacement?

ericwrice: I would recommend leaving it as is for now to match the existing convention and then add in new methods for all plotting features that are type checker friendly. Eventually, the auto generated methods could then be deprecated if desired. 

Maybe something along the lines of this pseudo-code:

```python
from enum import StrEnum, auto


class PROPERTIES(StrEnum):
    s = auto()
    z = auto()
    y = auto()
    a = auto()
    h = auto()
    t = auto()

class FUNCTIONS(StrEnum):
    re = auto()
    im= auto()
    mag = auto()
    db = auto()
    db10 = auto()
    rad = auto()
    deg = auto()
    arcl = auto()
    rad_unwrap = auto()
    deg_unwrap = auto()
    arcl_unwrap = auto()
    vswr = auto()
    time = auto()
    time_db = auto()
    time_mag = auto()
    time_impulse = auto()
    time_step = auto()

# There would be similar implementations for other plots such as plot_smith(), plot_violin(), etc.
def plot_rect(self: Network, attribute: PROPERTIES, conversion: FUNCTIONS, *args, **kwargs):
    self.plot_attribute(attribute, conversion, *args, **kwargs)

# These would be used instead of plot_s_db()
plot_rect(PROPERTIES.s, FUNCTIONS.db)  # Autocomplete and type checkers should work
plot_rect('s', 'db')  # Strings could be used instead, though some type checkers will complain
```

FranzForstmayr: Even better than using `Enum` is the `Literal` type in this case. 
For the user it's used exactly like a string (no import needed) but type checkers recognice if a allowed string is used.
https://peps.python.org/pep-0586/#shortening-unions-of-literals

However this is only available from 3.8 up, but I think it's to move on in the next time anyway.

FranzForstmayr: This done for example in numpy's `fft` code, where you can choose `forward`, `backward` or `ortho` norm.

https://numpy.org/doc/stable/reference/generated/numpy.fft.fft.html#numpy.fft.fft
https://github.com/numpy/numpy/blob/v2.0.0/numpy/fft/_pocketfft.pyi#L15

skrf/plotting.py
```diff
@@ -1449,6 +1450,71 @@ def plot_minmax_bounds_component(self: NetworkSet, attribute: str, m: int = 0, n
...
+
+    freq = self.ntwk_set[0].f
+
+    # default widths to 3/4 distance between frequencies
+    if not widths and len(freq) > 1:
+        widths = (freq[1]-freq[0])*0.75
+    elif not widths:
+        widths = 0.5
+
+    data = np.array([getattr(p, attribute)[:,m,n] for p in self.ntwk_set])
```
FranzForstmayr, line 1509: this `np.array` cast is not needed I think.

FranzForstmayr:
So I think the currently implemented `plot_violin_component` is already fine, although I would rename it to `plot_violin` directly.

I don't see any point to release a version with the autogenerated functions, so I would remove this autogeneration.
One can call `networkset.plot_violin('s_db')` which is fine I think.

Using Literals would become a follow up PR once Python >= 3.8 is used.

FranzForstmayr (changes requested):
I think the actual implementation of `plot_violin` is fine, although I would rename it to `plot_violin` directly.

I don't see any point releasing a version with autogenerated functions, if we can straight move on to the new improved approach. One can call this method simply like `networkset.plot_violin('s_db')` which is not harder than `networkset_plot_violin_s_db`. This is basically already possible with your current implementation, just remove the `__add_a_plot_violin` method. And I support your idea of using this approach for `plot_rect` as well.

The conversion from plain strings to Literals can be done, once we remove python3.7 from our supported python versions. This can be done in the next or the few releases I think.

jhillairet:
Thanks @FranzForstmayr for the review. I suggest making the minimal changes for now, to include this in the next release V1.1. 

The Python 3.7 can be drop in later release after. OK?

skrf/networkSet.py
ericwrice: Yep, I like this suggestion. The 3.7 support was my concern with proposing this. I'll keep this on the list for future versions.

skrf/plotting.py
ericwrice: It ensures that the shape is correct for the matplotlib violinplot method. Doesn't work as a list of arrays.

ericwrice:
Thanks for all the feedback! I'll make the quick changes for V1.1 and look to support the proposed plotting methods in future versions.

skrf/plotting.py
FranzForstmayr: oh yes, my fault.

jhillairet (approved):

jhillairet:
Thank you @ericwrice

PR #1085, opened 2024-06-06
Fixed the issue that the Network copy method lost some attributes.

Asachoo:
This PR addresses the issue where the Network class's copy method fails to retain certain attributes. For example:
```python
>>> import skrf as rf
>>> frequency = rf.Frequency(1, 10, 101, 'ghz')
>>> P1 = rf.Circuit.Port(frequency=frequency, name='p1')
>>> P2 = P1.copy()
>>> P1._is_circuit_port
True
>>> P2._is_circuit_port
  File "<stdin>", line 1, in <module>
  File "/home/skrf/network.py", line 951, in __getattr__
    raise AttributeError
AttributeError
```
To ensure safety, the copy method only copies attributes and skips methods.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2029,6 +2029,12 @@ class Network:
           ntwk.noise = self.noise.copy()
           ntwk.noise_freq = self.noise_freq.copy()
 
+        # copy special attributes (such as _is_circuit_port) but skip methods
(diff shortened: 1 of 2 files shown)


jhillairet (approved):
LGTM

PR #1071, opened 2024-05-13
Network innerconnect

Asachoo:
This PR aims to improve the performance of `innerconnect_s` method during Network connection. And solved the issue that the circuit calculation slows down when using auto_reduce=True for small-scale circuits in https://github.com/scikit-rf/scikit-rf/pull/1069. Related to https://github.com/scikit-rf/scikit-rf/issues/574 https://github.com/scikit-rf/scikit-rf/pull/1069

This PR is to achieve acceleration by avoiding repeated matrix calculations and caching intermediate calculation results. In my benchmark, the `innerconnect_s` method has a **5~6 times** performance improvement in MKL/OpenBLAS backend.

Your feedback and suggestions on this optimization are welcome!

Benchmark Code
```python
import numpy as np
from time import time, sleep
from matplotlib.pyplot import plot, show, legend, scatter

def innerconnect_s(A: np.ndarray, k: int, l: int) -> np.ndarray:
    """
    Performance optimized innerconnect_s function.
    """

    if k > A.shape[-1] - 1 or l > A.shape[-1] - 1:
        raise (ValueError("port indices are out of range"))

    nA = A.shape[1]  # num of ports on input s-matrix

    # external ports index
    ext_i = [i for i in range(nA) if i not in (k, l)]

    # Indexing sub-matrices of internal ports (only k, l)
    Akl = 1.0 - A[:, k, l]
    Alk = 1.0 - A[:, l, k]
    Akk = A[:, k, k]
    All = A[:, l, l]

    # Indexing sub-matrices of other external ports
    Ake = A[:, k, ext_i].T
    Ale = A[:, l, ext_i].T
    Aek = A[:, ext_i, k].T
    Ael = A[:, ext_i, l].T

    # Create an suit-sized s-matrix, to store the result
    x, y = np.meshgrid(ext_i, ext_i)
    C = A[:, y, x]

    # create temporary matrices for calculation
    det = (Akl * Alk - Akk * All)
    tmp_a = (Ael * Alk + Aek * All) / det
    tmp_b = (Ael * Akk + Aek * Akl) / det

    # loop through ports and calculates resultant s-parameters
    for i in range(nA - 2):
        for j in range(nA - 2):
            C[:, i, j] += Ake[j] * tmp_a[i] + Ale[j] * tmp_b[i]

    return C

def stand_rst(A: np.ndarray, k: int, l: int) -> np.ndarray:
    """
    Innerconnect_s function in master.
    """
[...]

PR #1094, opened 2024-06-14
Optimize the performance of innerconnect_s method.

Asachoo:
This PR aims to improve the performance of innerconnect_s method during Network connection. Related to https://github.com/scikit-rf/scikit-rf/issues/574 https://github.com/scikit-rf/scikit-rf/pull/1071

This PR is to achieve acceleration by reducing the number of for-loops and optimizing the location where `det` is calculated. These optimizations have demonstrated a nearly 20% performance improvement in my benchmarks.

Benchmark Code
```python3
from timeit import timeit

import numpy as np


def comman_code(A: np.ndarray, k: int, l: int):
    if k > A.shape[-1] - 1 or l > A.shape[-1] - 1:
        raise (ValueError("port indices are out of range"))

    nA = A.shape[1]  # num of ports on input s-matrix

    # external ports index
    ext_i = [i for i in range(nA) if i not in (k, l)]

    # Indexing sub-matrices of internal ports (only k, l)
    Akl = 1.0 - A[:, k, l]
    Alk = 1.0 - A[:, l, k]
    Akk = A[:, k, k]
    All = A[:, l, l]

    # Indexing sub-matrices of other external ports
    Ake = A[:, k, ext_i].T
    Ale = A[:, l, ext_i].T
    Aek = A[:, ext_i, k].T
    Ael = A[:, ext_i, l].T

    # Create an suit-sized s-matrix, to store the result
    x, y = np.meshgrid(ext_i, ext_i)
    C = A[:, y, x]

    # create temporary matrices for calculation
    det = Akl * Alk - Akk * All

    return C, det, Ake, Ale, Aek, Ael, Akl, Alk, Akk, All, nA


def init_mehod(A: np.ndarray, k: int, l: int) -> np.ndarray:
    # This is the extracted public code
    C, det, Ake, Ale, Aek, Ael, Akl, Alk, Akk, All, nA = comman_code(A, k, l)

    tmp_a = (Ael * Alk + Aek * All) / det
    tmp_b = (Ael * Akk + Aek * Akl) / det

    # loop through ports and calculates resultant s-parameters
    for i in range(nA - 2):
        for j in range(nA - 2):
            C[:, i, j] += Ake[j] * tmp_a[i] + Ale[j] * tmp_b[i]

    return C


def modi_mehod(A: np.ndarray, k: int, l: int) -> np.ndarray:
    # This is the extracted public code
    C, det, Ake, Ale, Aek, Ael, Akl, Alk, Akk, All, nA = comman_code(A, k, l)

    tmp_a = Ael * (Alk / det) + Aek * (All / det)
    tmp_b = Ael * (Akk / det) + Aek * (Akl / det)

    # loop through ports and calculates resultant s-parameters
    for i in range(nA - 2):
        C[:, i, :] += (Ake * tmp_a[i] + Ale * tmp_b[i]).T

    return C


if __name__ == "__main__":
    nfreq, nports, ncalls = 2001, 16, 2000
    np.random.seed(0)
    A = np.random.rand(nfreq, nports, nports)
    k, l = 2, 3
    print(np.allclose(init_mehod(A.copy(), k, l), modi_mehod(A.copy(), k, l)))

    t1 = timeit(lambda: modi_mehod(A, k, l), number=ncalls) / ncalls * 1000.0
    print(f"Modified nethod: {t1:.4f}ms")

    t2 = timeit(lambda: init_mehod(A, k, l), number=ncalls) / ncalls * 1000.0
    print(f"Initial method: {t2:.4f}ms")

```
And the output is as follows,
```
True
Modified nethod: 1.3667ms
Initial method: 1.8151ms
```

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -5910,8 +5910,8 @@ def connect_s(A: np.ndarray, k: int, B: np.ndarray, l: int) -> np.ndarray:
 
     # create composite matrix, appending each sub-matrix diagonally
     C = np.zeros((nf, nC, nC), dtype='complex')
-    C[:, :nA, :nA] = A.copy()
-    C[:, nA:, nA:] = B.copy()
+    C[:, :nA, :nA] = A
+    C[:, nA:, nA:] = B
 
     # call innerconnect_s() on composit matrix C
     return innerconnect_s(C, k, nA + l)
@@ -5999,13 +5999,12 @@ def innerconnect_s(A: np.ndarray, k: int, l: int) -> np.ndarray:
 
     # create temporary matrices for calculation
     det = (Akl * Alk - Akk * All)
-    tmp_a = (Ael * Alk + Aek * All) / det
-    tmp_b = (Ael * Akk + Aek * Akl) / det
+    tmp_a = Ael * (Alk / det) + Aek * (All / det)
+    tmp_b = Ael * (Akk / det) + Aek * (Akl / det)
 
     # loop through ports and calculates resultant s-parameters
     for i in range(nA - 2):
-        for j in range(nA - 2):
-            C[:, i, j] += Ake[j] * tmp_a[i] + Ale[j] * tmp_b[i]
+        C[:, i, :] += (Ake * tmp_a[i] + Ale * tmp_b[i]).T
 
     return C
 


skrf/network.py
```diff
@@ -5999,13 +5999,12 @@ def innerconnect_s(A: np.ndarray, k: int, l: int) -> np.ndarray:
 
     # create temporary matrices for calculation
     det = (Akl * Alk - Akk * All)
-    tmp_a = (Ael * Alk + Aek * All) / det
-    tmp_b = (Ael * Akk + Aek * Akl) / det
+    tmp_a = Ael * (Alk / det) + Aek * (All / det)
+    tmp_b = Ael * (Akk / det) + Aek * (Akl / det)
```
jhillairet, line 6003: Why is this more optimized than the previous implementation?

Asachoo: The shape of ndarrays (`Alk`, `All`, `Akk`, `Akl`, and `det`) are (n_freq,), but the shape of ndarrays (`Ael`, `Aek`) are (n_freq, nA-2).
So I believe that performing `Ael * (Alk / det) + Aek * (All / det)` is faster than `(Ael * Alk + Aek * All) / det`, because it is appropriate to use two small-scale computations to replace the large-scale computation performed via broadcasting.

Asachoo: A demo may be able to describe my idea more clearly,
```python
>>> import numpy as np
>>> from timeit import timeit
>>> 
>>> a=np.random.rand(4,101)
>>> b=np.random.rand(4,101)
>>> c=np.random.rand(101)
>>> d=np.random.rand(101)
>>> e=np.random.rand(101)
>>>
>>> timeit(lambda: (a*c)/e) # Perform large scale computation (a * c) / e at first
2.020706125003926
>>> timeit(lambda: a*(c/e)) # Perform small scale computation a * (c / e) at first
1.373241916000552
>>> timeit(lambda: (a*c+b*d)/e)
3.5458894160037744
>>> timeit(lambda: a*(c/e)+b*(d/e))
3.1978541249991395
```

jhillairet: Interesting! Are you using some tools to find these improvements?

jhillairet (approved):

skrf/network.py
Asachoo: Usually, I use tools like `cProfile` and `VizTracer` to profile my code. However, optimizations such as swapping division orders come from extensive testing and experimentation on my part. 😂

PR #1095, opened 2024-06-14
Optimize the performance of connect method.

Asachoo:
This PR aims to improve the performance of connect method during Network connections that need to maintain order. Related to  https://github.com/scikit-rf/scikit-rf/pull/1071

This PR aims to speed up networks that need to maintain order by pre-sorting the scattering matrix instead of doing the expensive renumber operation after the calculation. These optimizations have shown nearly 15% performance improvement in my benchmarks.

Benchmark Code
```
from time import time

import numpy as np

import skrf as rf

if __name__ == "__main__":
    np.random.seed(0)

    fcalls = 500
    ts = []
    ports = range(2, 10)

    freqn, Bports = 501, 2
    freq = rf.Frequency(start=1, stop=10, npoints=freqn, unit="ghz")
    ntwkB = rf.Network(freqquency=freq)
    ntwkB.s = np.random.rand(freqn, Bports, Bports) + 1j * np.random.rand(
        freqn, Bports, Bports
    )

    for n in ports:
        np.random.seed(0)
        ntwkA = rf.Network(freqquency=freq)

        ntwkA.s = np.random.rand(freqn, n, n) + 1j * np.random.rand(freqn, n, n)

        tmp = []
        for _ in range(fcalls):
            s = -time()
            rf.connect(ntwkA, n - 1, ntwkB, 1)
            s += time()
            tmp.append(s * 1000)
        ts.append(np.min(tmp))

    plt.plot(ports, ts)
    plt.xlabel("Number of ports")
    plt.ylabel("Time taken (ms)")
    plt.ylim([0.2, 1])
    plt.show()
```
master branch benchmark:​​
Master
this branch benchmark:
AvoidRenumbering
Speed up:
SpeedUp


skrf/network.py
```diff
@@ -3230,6 +3230,10 @@ def renumber(self, from_ports: Sequence[int], to_ports: Sequence[int]) -> None:
             List of port indices to change. Size between 1 and N_ports.
         to_ports : list-like
             List of desired port indices. Size between 1 and N_ports.
+        only_z0 : bool
+            If true only z0 is renumbered, s-parameters are not affected.
+            This should only be used after executing the "connect_s" method
+            which keeps the port index where you expect it to be.
```
jhillairet, line 3236: Could add an example using the only_z0 in the example section of the docstring?

Asachoo: I have added a new case only_z0=True to the docstring.
Additionally, I attempted to resolve the LC Filter image error on the [Circuit tutorial page](https://scikit-rf.readthedocs.io/en/latest/tutorials/Circuit.html).  I apologize for inadvertently introducing this issue, I suspect the issue arose because both the Circuit reduction images and the original LC Filter images share the same name, `![low pass filter]`. This discrepancy does not occur in my local notebook environment.

jhillairet (approved):
Great addition, thank you!

PR #1096, opened 2024-06-17
Optimize the performance of Network copy mehod

Asachoo:
This PR aims to improve the performance of copy method. The `Network.copy()` method is widely called in the skrf package, such as `rf.connect()`, `Network.interpolate()` and other time-consuming methos.

This PR improves efficiency in two aspects:
1. **Direct Assignment to `ntwk._s`**: Since `self.s` is already broadcast-safety, directly assign values ​​to `ntwk._s` is efficient, instead of passing the s-parameters through the `ntwk.__init__()` method and using `s.setter` with `fix_param_shape()`.
2. In order to solve the issue https://github.com/scikit-rf/scikit-rf/pull/1085, `set(self.dir()) - set(ntwk.dir())` is used to identify extended attributes. Profiling revealed that `dir()` is quite slow, so I've added a `_ext_attrs` list to record extended attributes. **Note:** This approach requires manually adding the attribute's name to `_ext_attrs` list when setting the extended attributes. While this approach is efficient, it may not a robust solution (maybe we should create a `Network.set_ext_attr()` method to auto add attribute's name to `_ext_attrs` and set the value?).

Rather than directly comparing the performance of `Network.copy()`, it might be more meaningful to focus on more commonly used scenarios such as `rf.connect()`. These optimizations have shown nearly 35% performance improvement in my benchmarks.

Benchmark Code:
```python
from time import time

import numpy as np

import skrf as rf

if __name__ == "__main__":
    np.random.seed(0)

    fcalls = 500
    ts = []
    ports = range(2, 10)

    freqn, Bports = 501, 2
    freq = rf.Frequency(start=1, stop=10, npoints=freqn, unit="ghz")
    ntwkB = rf.Network(freqquency=freq)
    ntwkB.s = np.random.rand(freqn, Bports, Bports) + 1j * np.random.rand(
        freqn, Bports, Bports
    )

    for n in ports:
        np.random.seed(0)
        ntwkA = rf.Network(freqquency=freq)

        ntwkA.s = np.random.rand(freqn, n, n) + 1j * np.random.rand(freqn, n, n)

        tmp = []
        for _ in range(fcalls):
            s = -time()
            rf.connect(ntwkA, n - 1, ntwkB, 1)
            s += time()
            tmp.append(s * 1000)
        ts.append(np.min(tmp))

    plt.plot(ports, ts)
    plt.xlabel("Number of ports")
    plt.ylabel("Time taken (ms)")
    plt.ylim([0.1, 0.8])
    plt.show()
```
master branch benchmark:
Master
this branch benchmark:
SpeedUp
Speed up:
1363a3e0-b3b7-48e2-963f-e0e623b49c13

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -251,18 +251,18 @@ class Circuit:
             return True
 
     @classmethod
-    def _is_port(cls, ntw):
+    def _is_port(cls, ntw: Network):
         """
         Return True is the network is a port, False otherwise
         """
-        return getattr(ntw, "_is_circuit_port", False)
+        return ntw._ext_attrs.get("_is_circuit_port", False)
 
     @classmethod
-    def _is_ground(cls, ntw):
+    def _is_ground(cls, ntw: Network):
         """
         Return True is the network is a ground, False otherwise
         """
-        return getattr(ntw, "_is_circuit_ground", False)
+        return ntw._ext_attrs.get("_is_circuit_ground", False)
 
     @classmethod
     def Port(cls, frequency: Frequency, name: str, z0: float = 50) -> Network:
@@ -296,8 +296,7 @@ class Circuit:
         """
         _media = media.DefinedGammaZ0(frequency, z0=z0)
         port = _media.match(name=name)
-        port._is_circuit_port = True
-        port._ext_attrs.append('_is_circuit_port')
+        port._ext_attrs['_is_circuit_port'] = True
         return port
 
     @classmethod
@@ -422,8 +421,7 @@ class Circuit:
 
         """
         ground = cls.ShuntAdmittance(frequency, Y=INF, name=name)
-        ground._is_circuit_ground = True
-        ground._ext_attrs.append('_is_circuit_ground')
+        ground._ext_attrs['_is_circuit_ground'] = True
         return ground
 
     @classmethod
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -296,7 +296,7 @@ class Circuit:
         """
         _media = media.DefinedGammaZ0(frequency, z0=z0)
         port = _media.match(name=name)
-        setattr(port, '_is_circuit_port', True)
+        port._is_circuit_port = True
         port._ext_attrs.append('_is_circuit_port')
         return port
 
@@ -422,7 +422,7 @@ class Circuit:
 
         """
         ground = cls.ShuntAdmittance(frequency, Y=INF, name=name)
-        setattr(ground, '_is_circuit_ground', True)
+        ground._is_circuit_ground = True
         ground._ext_attrs.append('_is_circuit_ground')
         return ground
 
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -296,7 +296,8 @@ class Circuit:
         """
         _media = media.DefinedGammaZ0(frequency, z0=z0)
         port = _media.match(name=name)
-        port._is_circuit_port = True
+        setattr(port, '_is_circuit_port', True)
+        port._ext_attrs.append('_is_circuit_port')
         return port
 
     @classmethod
@@ -421,7 +422,8 @@ class Circuit:
 
         """
         ground = cls.ShuntAdmittance(frequency, Y=INF, name=name)
-        ground._is_circuit_ground = True
+        setattr(ground, '_is_circuit_ground', True)
+        ground._ext_attrs.append('_is_circuit_ground')
         return ground
 
     @classmethod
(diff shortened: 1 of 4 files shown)


skrf/network.py
```diff
@@ -2014,11 +2015,11 @@ def copy(self) -> Network:
...
         """
-        ntwk = Network(s=self.s,
-                       frequency=self.frequency,
-                       z0=self.z0, s_def=self.s_def,
-                       comments=self.comments
-                       )
+        ntwk = Network(z0=self.z0, s_def=self.s_def, comments=self.comments)
+
+        ntwk._s = self.s.copy()
+        ntwk.frequency = self.frequency
```
FranzForstmayr, line 2021: I think `ntwk.frequency` is not immutable yet, however I think it was decided once, that it should be immutable.
But for now it's probably best to call `self.frequency.copy()`

skrf/circuit.py
```diff
@@ -297,6 +297,7 @@ def Port(cls, frequency: Frequency, name: str, z0: float = 50) -> Network:
         _media = media.DefinedGammaZ0(frequency, z0=z0)
         port = _media.match(name=name)
         port._is_circuit_port = True
+        port._ext_attrs.append('_is_circuit_port')
```
FranzForstmayr, line 300: I'm not a fan of manually appending a list after setting an internal value...

What about storing arbitrary information like `_is_circuit_port` in an ordinary dictionary or even an predefined dataclass?

FranzForstmayr (changes requested):

FranzForstmayr:
The speed gain from the direct assignment to `_s` is impressive, thank you for that great improvement :)

skrf/network.py
Asachoo: Yes, I completely agree that `ntwk.frequency` should remain immutable.

Before submitting this PR, I verified whether `ntwk.frequency` is just a shallow copy or a view of `self.frequency`. It turns out to be safe: `ntwk.frequency` has a different ID than `self.frequency`, and modifying `self.frequency` does not affect the copied `ntwk.frequency`.

```python
>>> import skrf as rf
>>> import numpy as np
>>> 
>>> frequency = rf.Frequency(1,10,101,'ghz')
>>> s = np.random.rand(101, 2, 2) + np.random.rand(101, 2, 2) * 1.j
>>> 
>>> ntwk = rf.Network(s=s, frequency=frequency)
>>> ntwk2 = ntwk.copy()
>>> 
>>> # Check frequency id 
>>> id(ntwk.frequency)
138540906669776
>>> id(ntwk2.frequency)
138540906742544
>>> # Check frequency points
>>> ntwk.frequency == ntwk2.frequency
True
>>> # Check frequency attribute independence/immutability
>>> ntwk.frequency = rf.Frequency.from_f(ntwk.frequency.f[:50])
>>> ntwk.frequency
1000000000.0-5410000000.0 Hz, 50 pts
>>> ntwk2.frequency
1.0-10.0 GHz, 101 pts
```

I think the `new_frequency.copy()` operation automatically performed in the `@frequency.setter` method ensures the immutability of `frequency` even when `self.frequency` is used instead of `self.frequency.copy()`.

https://github.com/scikit-rf/scikit-rf/blob/3ba8bf84aa07f9dab0ab70992c70dbc178f7929e/skrf/network.py#L1372-L1383

Therefore, I think calling `self.frequency` directly can provide higher performance while maintaining the safety and integrity of the result.

I look forward to hearing your thoughts on this and discussing it further.

skrf/circuit.py
Asachoo: Creating an ordinary dictionary for `Network._ext_attrs` is indeed a simple and effective approach. I will update the relevant code accordingly.

Asachoo:
Additionally, the issue where some `Network` objects generated by Media had empty `port_modes` array has been fixed. The specific errors are as follows:

Operations on the master branch:
```python
>>> import skrf as rf
>>> freq = rf.Frequency(1,10,101,'ghz')
>>> line = rf.media.DefinedGammaZ0(frequency=freq)
>>> cap = line.capacitor(1)
>>> cap.port_modes
array([], dtype=float64)
```

Operations on this branch:
```python
>>> import skrf as rf
>>> freq = rf.Frequency(1,10,101,'ghz')
>>> line = rf.media.DefinedGammaZ0(frequency=freq)
>>> cap = line.capacitor(1)
>>> cap.port_modes
array(['S', 'S'], dtype='<U1'
```

skrf/network.py
FranzForstmayr: great

FranzForstmayr (approved):

jhillairet (approved):

jhillairet:
ok to be merged?

Asachoo:
Yes, I consider it finished.

Issue #1099, opened 2024-06-18
network_2_spreadsheet doesn't scale frequency with units


soxand16:
Thank you for reporting a bug! To help us understanding the issue and how to reproduce it, please fill the following sections as best as you can.

**Describe the bug**

**To Reproduce**
```
import skrf as rf

ring_slot = rf.Network('data/ring slot.s2p')
ring_slot.frequency.unit = 'MHz'
ring_slot.write_spreadsheet('data/ringslot_spreadsheet',file_type='csv')
```

First two lines of saved csv:
```
Freq(MHz),S11 Log Mag(dB),S11 Phase(deg),S21 Log Mag(dB),S21 Phase(deg),S12 Log Mag(dB),S12 Phase(deg),S22 Log Mag(dB),S22 Phase(deg)
75000000000.0,-3.3408248302499572,137.7316273455639,-2.9169962710163078,30.874885544911944,-2.9169962710163078,30.874885544911944,-3.3708001371849146,107.1105380340549
...
```

**Expected behavior**
After changing the frequency units to MHz, we'd expect for the frequency value to be in MHz instead of Hz (e.g. 75000.0 instead of the 75000000000.0). Issue persists for kHz, GHz, etc.

**Data**
Used `ring slot.s2p` from the data directory in the repository.

**System**
 - OS: Windows 11, macOS Sonoma 14.5
 - scikit-rf version 1.0.0

PR #1100, opened 2024-06-18
scaled frequency units

soxand16:
Resolves Issue #1099.

### DIFF
--- a/skrf/io/general.py
+++ b/skrf/io/general.py
@@ -686,7 +686,7 @@ def network_2_spreadsheet(ntwk: Network, file_name: str = None,
         file_name = ntwk.name + '.'+file_extns[file_type]
 
     d = {}
-    index =ntwk.frequency.f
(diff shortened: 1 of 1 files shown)


jhillairet:
Is the reading of a file with corrected frequencies has the correct frequencies?

soxand16:
@jhillairet Importing data from a csv saved using `network_2_spreadsheet` isn't part of my workflow, but I don't see any function designed to read in that particular csv format. 

There is `read_pna_csv` and functions like it that return a data frame and network, but they are specifically tailored for Agilient PNA output and do not work on a csv outputted by `network_2_spreadsheet` due to formatting differences.

jhillairet:
OK. Just wondering the impact on other users. But indeed, the label of the frequency column was supposed to be scaled to the unit, so your PR is fully justified.

jhillairet (approved):

jhillairet:
Thank you @soxand16

PR #1102, opened 2024-06-24
Drop python 3.7 to use literals and walrusses :)

FranzForstmayr:


jhillairet (approved):

jhillairet:
Ok, LGTM. Bye bye python 3.7...

PR #1104, opened 2024-06-25
Add MDIF functionality for improved interaction with ADS

wsapia:
network.py
- Add a space in the SKRF comment

networkSet.py
- Two more booleans for the write_mdif method:
    - One to change how the variables are written which allows ADS to read the file properly
    - One to allow the user to remove the SKRF comment

io/mdif.py
- Pass through the same two booleans from the networkSet.py write_mdif method

### DIFF
--- a/skrf/io/mdif.py
+++ b/skrf/io/mdif.py
@@ -365,9 +365,10 @@ class Mdif:
               filename : str,
               values: dict | None = None,
               data_types: dict | None = None,
-              ads_compatible: bool = True,
+              *,
               comments: str | None = None,
-              skrf_comment: bool = True):
+              skrf_comment: bool = True,
+              ads_compatible: bool = True):
         """
         Write a MDIF file from a NetworkSet.
 
@@ -385,14 +386,14 @@ class Mdif:
         data_types: dictionary or None. Default is None.
             The keys are MDIF variables and the value are datatypes
             specified by the following strings: "int", "double", and "string"
-        ads_compatible: bool. Default is True.
-            Indicates whether to write the file in a format that
-            ADS will read properly.
         comments: list of strings
             Comments to add to output_file.
             Each list items is a separate comment line
         skrf_comment : bool, optional
             write `created by skrf` comment
+        ads_compatible: bool. Default is True.
+            Indicates whether to write the file in a format that
+            ADS will read properly.
 
         See Also
         --------
--- a/skrf/io/mdif.py
+++ b/skrf/io/mdif.py
@@ -444,11 +444,11 @@ class Mdif:
                     if p not in data_types:
                         data_types[p] = "double"
 
-                    var_type = "" if ads_compatible else f" {p}({dict_types[data_types[p]]})"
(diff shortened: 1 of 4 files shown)


skrf/io/mdif.py
```diff
@@ -365,7 +365,9 @@ def write(ns : NetworkSet,
               filename : str,
               values: dict | None = None,
               data_types: dict | None = None,
-              comments = None):
+              ads_compatible: bool = True,
+              comments: str | None = None,
+              skrf_comment: bool = True):
```
FranzForstmayr, line 370: please keep the order and do not insert `ads_compatible` before `comments`.
Also use `*` to seperate boolean comments from the rest and make them keyword only for clarification.

https://peps.python.org/pep-3102/

Same below.

FranzForstmayr, line 370: please keep the order and do not insert ads_compatible before comments.
Also use * to seperate boolean comments from the rest and make them keyword only for clarification.

https://peps.python.org/pep-3102/

Same below.

FranzForstmayr (changes requested):

skrf/io/mdif.py
wsapia: Sure thing!

skrf/io/mdif.py
```diff
@@ -365,7 +365,10 @@ def write(ns : NetworkSet,
               filename : str,
               values: dict | None = None,
               data_types: dict | None = None,
-              comments = None):
+              *,
+              comments: str | None = None,
+              skrf_comment: bool = True,
+              ads_compatible: bool = True):
```
FranzForstmayr, line 371: I thought about:

```python
@staticmethod
    def write(ns : NetworkSet,
              filename : str,
              values: dict | None = None,
              data_types: dict | None = None,
              comments: str | None = None,
              *,
              skrf_comment: bool = True,
              ads_compatible: bool = True):
```

because of comments already exists. Sorry for not beeing clear before.

skrf/networkSet.py
```diff
@@ -948,7 +948,9 @@ def write_mdif(self,
                    filename: str,
                    values: dict | None = None,
                    data_types: dict | None = None,
-                   comments = None):
+                   ads_compatible: bool = True,
+                   comments: str | None = None,
+                   skrf_comment: bool = True):
```
FranzForstmayr, line 953: Same here.

FranzForstmayr (changes requested):

wsapia:
Alright, changes made

FranzForstmayr:
Great, thanks for your work!

could you add one or two unit tests to make sure everything stays as is should in the future?

wsapia:
Tests added

jhillairet (approved):

FranzForstmayr:
LGTM

We can merge this one although CI is failing due to a new ruff check.
The ruff error is legit, because `np.trapz` was removed, but we  didn't detect that earlier due to missing coverage.
https://coveralls.io/builds/68353193/source?filename=skrf%2FvectorFitting.py#L140
We will fix this in a separate PR, as a new unit test for `vectorfitting` is required as well.

Issue #1090, opened 2024-06-07
Error with NanoVNA if frequency npoints = 1


mxxmxn:
**Describe the bug**
When I try to set the frequency to one specific point, I get an error.

**To Reproduce**
Run the following example with a NanoVNA:
```
instr.frequency = Frequency(start=2.4, stop=2.4, npoints=1, unit='GHz')
```

**Expected behavior**
I should be able to set the frequency.

**System**
 - OS: Linux
 - scikit-rf version: 1.0.0

FranzForstmayr:
Can you post the exact error message? I don't have access to a NanoVNA.

mxxmxn:
The issue can be reproduced without a NanoVNA, the device string only needs to be a valid serial interface:
```
import skrf
from skrf.vi.vna.nanovna import NanoVNA

instr = NanoVNA("ASRL/dev/tty1::INSTR")
freq = skrf.Frequency(start=2.4, stop=2.4, unit='GHz', npoints=1)
instr.frequency = freq
```

I get the following Error:
```
skrf/frequency.py:412: RuntimeWarning: invalid value encountered in scalar divide
  return self.span/(self.npoints-1.)
Traceback (most recent call last):
  File "scratches/scratch.py", line 6, in <module>
    instr.frequency = freq
    ^^^^^^^^^^^^^^^
  File "skrf/vi/vna/nanovna/nanovna.py", line 155, in frequency
    self.write(OP.WRITE8, REG_ADDR.SWEEP_STEP, 8, f.step)
  File "skrf/vi/vna/nanovna/nanovna.py", line 74, in write
    arg = int(arg).to_bytes(nbytes, byteorder="little", signed=False)
          ^^^^^^^^
ValueError: cannot convert float NaN to integer
```

PR #1108, opened 2024-06-27
fix frequency for npoints=1

mxxmxn:
Fixes error with frequency span for npoints = 1. See #1090

### DIFF
--- a/skrf/frequency.py
+++ b/skrf/frequency.py
@@ -409,7 +409,10 @@ class Frequency:
         --------
         df : for general case
         """
-        return self.span/(self.npoints-1.)
+        if self.span == 0:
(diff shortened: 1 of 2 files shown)


jhillairet:
Thank you.

Seems the linter is complaining for another reason in

```
skrf/vectorFitting.py:140:25: NPY201 `np.trapz` will be removed in NumPy 2.0. Use `numpy.trapezoid` on NumPy 2.0, or ignore this warning on earlier versions.
```

mxxmxn:
I fixed it (hopefully), even though it has nothing to do with the original issue.

jhillairet (approved):

Issue #1062, opened 2024-05-03
No good results with IEEEP370_MM_ZC_2xThru


Hasko7:
### **Description of the issue**
Hi guys, I am using the mixed mode IEEEP370 de-embedding implementation from scikit-rf, but I am not getting good results with MM IEEE 370 de-embedding with impedance correction ZC. 

I simulated a Beatty structure as a DUT in CST and then with the fixtures as a composite structure Fixture1-DUT-Fixture2. I also simulated fixtures without the DUT as 2x-Thru. After the simulation, I produced it as a PCB and measured it.  For validation, I applied the mixed-mode de-embedding from scikit-rf to it and compared it with the simulated DUT from CST. 

I think that fixture 2 is returned incorrectly with ZC, because after de-embedding there is always a ripple on the right side in the time domain. This is also visible in the frequency domain. 

### **Code**
All the utilized MM S-Parameters are in odd/even format.  Therefore i used at the beginning the renumber-methode: 
 .renumber([0,2,1,3], [0, 1, 2, 3])

**This is how I realized the ZC deembedding:
simulated data:** 
```
z0 = 50
dm_mmzc_bend  = IEEEP370_MM_ZC_2xThru(dummy_2xthru = se_2xthru_bend, dummy_fix_dut_fix = se_fdf_bend,
                         z0 = z0, bandwidth_limit = 67e9, pullback1 = 0, pullback2 = 0,
                         leadin = 1, NRP_enable = True, name = 'mmzc sharp bends simulated')
se_d_dut_bend_zc = dm_mmzc_bend.deembed(se_fdf_bend)
se_zc_fix1 = dm_mmzc_bend.se_side1.copy()
se_zc_fix1.name = 'zc_FIX-1 sharp bends simulated'
se_zc_fix2 = dm_mmzc_bend.se_side2.copy()
se_zc_fix2.name = 'zc_FIX-2 sharp bends simulated'

mm_d_dut_bend_zc = se_d_dut_bend_zc.copy()
mm_d_dut_bend_zc.se2gmm(p = 2)
mm_zc_fix1_bend = se_zc_fix1.copy()
mm_zc_fix1_bend.se2gmm(p = 2)
mm_zc_fix2_bend = se_zc_fix2.copy()
mm_zc_fix2_bend.se2gmm(p = 2)

mm_d_dut_bend_zc_dc = mm_d_dut_bend_zc.extrapolate_to_dc(kind='cubic')
[...]

PR #698, opened 2022-06-11
IEEEP370 Flip FIX-2 and minor improvements

mhuser:
Make IEEEP370 deembedding method use the labelling and numbering scheme recommended by the IEEEP370 standard.
The deembedding equation should be like this:
`DUT = FIX-1.inv ** FIX-DUT-FIX ** FIX-2.flipped().inv`

Also add minor improvements:

- Add a verbose mode that plot NZC  length and midpoint impedance determination, useful for diagnostic purpose
- Add the possibility to use z-transform instead fft for very short structure (based on the Matlab code)

PR #1114, opened 2024-07-07
Enhance IEEEP370 deembedding documentation, plots, code readability and verbose output 

mhuser:
Currently, if single-ended deembedding `IEEEP370_SE_ZC_2xThru` or its mixed-modes counter-part IEEEP370_MM_ZC_2xThru are called with `verbose = True`, the resulting plots showing the peeling effects are interesting but difficult to read.
image

With this PR, a plot showing the determination of gamma with alpha and beta is added as a sanity check. It can also help to tune the bandwidth in case it is required to fit alpha with a 2nd-order polynomial. The plot showing the superposed peeling state shows instead the start with FIX-DUT-FIX superposed with the end with deembeded DUT and modeled fixture, which are visually more simple to check.
image

image

This PR add the following assets:

- Update notebook to use a more realistic asymmetric fixture with connector (and explain the discrepency with aicc deembedding tool)
- Introduce helpers plots methods for fixture electrical requirements and consistency tests
- Use static methods and an abstract class `IEEEP370` to avoid code duplication between `IEEEP370_SE_NZC_2xThru` and `IEEEP370_SE_ZC_2xThru`
- Add `IEEEP370.extrapolate_to_dc()` method to be able to plot fixtures, deembedded dut and inputs networks with the same DC point used by the deembedding algorithm.
- Add more information to verbose plots and add debug variables to make internal intermediate results such as the length of the fixture available to advanced users
- Fix extrapolation to DC to match network's `z0` in case it is not 50 ohm.
- Add docstrings


skrf/calibration/deembedding.py
```diff
@@ -1097,7 +1097,58 @@ def deembed(self, ntwk):
 
         return s_side1.inv ** ntwk ** s_side2.flipped().inv
 
-    def dc_interp(self, s, f):
+    def extrapolate_to_dc(ntwk):
```
jhillairet, line 1100: shouldn't it defined as @staticmethod ?

jhillairet: same for other methods below where `self` has been suppressed?

skrf/calibration/deembedding.py
```diff
@@ -2572,7 +2705,11 @@ def split2xthru(self, se_2xthru, se_fdf):
                                   NRP_enable = self.NRP_enable,
                                   leadin = self.leadin,
                                   verbose = self.verbose)
-
+        # debug outpus
+        self.gamma_cc = dm_cc.gamma
+        self.x_end_cc = dm_cc.x_end
+        self.z_side1_cc = dm_cc.z_side1
+        self.z_side2_cc = dm_cc.z_side2
```
jhillairet, line 2712: still needed? (same above)

skrf/calibration/deembedding.py
mhuser: Indeed. They should probably need a docstring too.

skrf/calibration/deembedding.py
mhuser: I think it is interesting for users who either want to check what happened internally or want to do a port-extension of their fixture. Should the need arise like in #1075  they have the gamma and the index to read the last impedance of the fixture. They can build from this a transmission line for common and differential modes and convert back to single-ended. We could include a dedicated method for that if there is more demand, but it is outstide the scope of this PR.

mhuser (commented):
I wonder if class `IEEEP370_MM_ZC_2xThru(Deembedding)` should inherit of class IEEEP370_SE_ZC_2xThru(Deembedding).  Only init, `deembed` and `split2xthru` changes slightly. Idem for its NZC counterpart.

mhuser:
Example notebook: https://scikit-rf--1114.org.readthedocs.build/en/1114/examples/networktheory/IEEEP370%20Deembedding.html
Effect of the docstrings: https://scikit-rf--1114.org.readthedocs.build/en/1114/api/calibration/generated/skrf.calibration.deembedding.IEEEP370_SE_ZC_2xThru.html#skrf.calibration.deembedding.IEEEP370_SE_ZC_2xThru

jhillairet (approved):
Huge work, thank you!

Ttl:
Some nits about the notebook:
* Some plots have bad labels, such as Z11 three times in the first time step plot. It's hard to say which trace corresponds to which network.
* Some magnitude and phase error plots are missing titles. It's hard to say which one is which.

mhuser:
> * Some plots have bad labels, such as Z11 three times in the first time step plot. It's hard to say which trace corresponds to which network.

Oh, you are right! The DC extrapolation method did not pass the `name` to the output network anymore!

I will also go trough all the plots and fix titles.

mhuser:
> Example notebook: https://scikit-rf--1114.org.readthedocs.build/en/1114/examples/networktheory/IEEEP370%20Deembedding.html
> Effect of the docstrings: https://scikit-rf--1114.org.readthedocs.build/en/1114/api/calibration/generated/skrf.calibration.deembedding.IEEEP370_SE_ZC_2xThru.html#skrf.calibration.deembedding.IEEEP370_SE_ZC_2xThru

Ok, I have it. The notebook and documentation should be available in some minutes after readthedocs build.

mhuser:
I know it would be good to add type hinting in function, more test cases (e.g. asymetrical fix to detect wrong flip) and maybe an "advanced IEEE 370" notebook showing the effect of different options, but maybe it is better to leave this for another PR

mhuser:
Waiting for #1124 to ease plots syntax.

skrf/calibration/deembedding.py
```diff
@@ -1384,7 +2144,108 @@ def split2xthru(self, s2xthru):
...
+                       linestyle = 'dashed', color = 'r')
+        ax[1].legend(loc = 'upper right')
+        fig.tight_layout()
+
+        return (fig, ax)
+
+    def plot_check_impedance(self, fix_dut_fix: Network = None, ax: Axes = None,
+                             window: str = 'hamming') -> (Figure, Axes):
+        # if dc point already exists, it will be replaced
+        s2xthru = IEEEP370.extrapolate_to_dc(self.s2xthru)
```
jhillairet, line 2186: remark/question of Oriented Object programmers: as the IEEEP370 class should not be used (since it's an Abstract) shouldn't you use `self.` instead of `IEEEP370.` (and the method not defined as static) ?

skrf/util.py
```diff
@@ -117,6 +117,20 @@ def wrapper(func: Callable) -> Callable:
         return func
     return wrapper
 
+def figure(*args, **kwargs) -> Figure:
```
jhillairet, line 120: shouldn't it in `plotting.py` instead than in `util.py` ?

jhillairet (commented):
Great work!
Somme comments/questions below. Also, it's better to clean the output of the notebooks, to reduce the amount of data to store in the git repo

mhuser:
> Also, it's better to clean the output of the notebooks, to reduce the amount of data to store in the git repo

Oh no, I committed an accident! I remove the notebook outputs ASAP.

skrf/calibration/deembedding.py
mhuser: Thank you for this excellent question. Because the static function is inherited, I could maybe use `self` here 🤔 
Maybe I can give a short description of the programming challenge and gather your opinion about it:

- The user may want to extrapolate using IEEEP370 or test the fixtures electrical requirements without instantiating one of the deembedding classes this is why I used static initially.
- Currently, if one of the IEEEP370 classes is instantiated, the fixture is split in init which is time-consuming and requires suitable networks (so it cannot be instantiated to use the helpers functions only)
- Because of the split in init, the mixed modes flavours cannot call the init of the single-ended one, because the fixture split would be done twice.

I feel like if the split should have been performed in a `run(self)` method outside init but it would a major break the API.

mhuser: > Because the static function is inherited, I could maybe use self here 🤔

Well I cannot use `self` of course, but the class name, which is of little ehancement

skrf/util.py
mhuser: I think it is related to d832ef6 with matplotlib being optionnal. Because `suplots` were already there I added `figure`.
It is also true both concern plotting.

mhuser: Afterthought: the optional matplotlib dependency is pertinent for a method that shall work without matplotlib (e.g. if it is only an optional verbose output). Plotting functions cannot be performed without it, because their sole purpose is to plot.

mhuser:
Hello @denzchoe, @jhillairet 
I am quite excited to share this new result regarding the comparison of AICC tool and scikit-rf for IEEE 370 performance:

There was always a small discrepancy remaining between both implementation for NZC algorithm.
Using the single-ended example of the notebook, but with connectors to make the fixtures non-symmetrical, it happens that the AICC tool has a wrong FIX-2 flip (black dashed on plot below). This is evident because the results are not the same on both sides for a symmetrical DUT.
Embedding this error into scikit-rf for comparison sake gives the mighty agreement (red curve), while scikit-rf gives the best result (green curve).

image

This was found thanks to @Asachoo works #1062  and is also the case for mixed-modes (scikit-rf gives the most correct results).

mhuser:
The single-ended case is ready at https://scikit-rf--1114.org.readthedocs.build/en/1114/examples/networktheory/IEEEP370%20Deembedding.html

I will update mixed-modes too.

jhillairet (approved):
Great job!! A new pro case for open-source software!

denzchoe:
> Hello @denzchoe, @jhillairet I am quite excited to share this new result regarding the comparison of AICC tool and scikit-rf for IEEE 370 performance:
> 
> There was always a small discrepancy remaining between both implementation for NZC algorithm. Using the single-ended example of the notebook, but with connectors to make the fixtures non-symmetrical, it happens that the AICC tool has a wrong FIX-2 flip (black dashed on plot below). This is evident because the results are not the same on both sides for a symmetrical DUT. Embedding this error into scikit-rf for comparison sake gives the mighty agreement (red curve), while scikit-rf gives the best result (green curve).
> 
> 
> This was found thanks to @Asachoo works #1062 and is also the case for mixed-modes (scikit-rf gives the most correct results).


<br>
<br>

@mhuser , been following #1062 for awhile now.. I got lost midway in the discussion (very cerebral. But very engaging!). 
Side question though; didn't you already corrected the Fixture (head-tail) direction and proven this in #698, June 2022? 
I got lost myself at reading the code how many times it `flip` the fixture .. so I cannot verify myself :D

mhuser:
Hello @denzchoe 

> didn't you already corrected the Fixture (head-tail) direction and proven this

Thank you for your insights. I think the NZC fixtures have been correct in `scikit-rf` for a while now. However, I did not - or do not remember which is also possible - realize the Amphenol tool had the wrong flip for FIX-2 with NZC. Probably they got confused too! Another pitfall is that testing with symmetrical fixtures like a single simulated transmission line section does not shows the flip issues.

mhuser:
@denzchoe Could I have your opinion on the updated notebook?

denzchoe:
> @denzchoe Could I have your opinion on the updated notebook?

alrighty, let me look at it in a glance later today

mhuser:
Some random thoughts:

I stumbled on this [MATLAB example](https://mathworks.com/help/rf/ug/applications-of-ieee-370.html) that use a `IEEEP3702xThru_ModifiedTimeGate` method using vector fitting to do the return loss time gating of `NZC` using [this paper](https://www.mathworks.com/content/dam/mathworks/tag-team/Objects/t/time-domain-gating-of-microwave-component-responses-using-analog-techniques.pdf). However there is no source and the method is not on the IEEE repository.

On the IEEE repository [some additional code has appeared](https://opensource.ieee.org/elec-char/ieee-370/-/blob/master/TG3/qualityCheckFrequencyDomain.m?ref_type=heads) to assess for causality, passivity, and reciprocity of the measurement with methods in time and frequency domains, which are also covered in the standard.

denzchoe:
> On the IEEE repository [some additional code has appeared](https://opensource.ieee.org/elec-char/ieee-370/-/blob/master/TG3/qualityCheckFrequencyDomain.m?ref_type=heads) to assess for causality, passivity, and reciprocity of the measurement with methods in time and frequency domains, which are also covered in the standard.

Yes this is a much needed feature!! Maybe in another PR. The amount of commits here in this PR is already a deal of work! 
So much have changed since the last time I use this `Class`. Wow! 
<br>

For the notebook: I wish we could add some visual aid of how the Transmission Lines as per your description looks like (KiCad's Layout or something that is embedded inside QucsStudio). But these are all Cosmetic level of improvement that can be attempted later. The Content now is Solid! 

<br>

Just a minor observation of your PR. I think we miss something. See https://github.com/mhuser/scikit-rf/issues/7

mhuser:
Hi @denzchoe,

> For the notebook: I wish we could add some visual aid of how the Transmission Lines as per your description looks like (KiCad's Layout or something that is embedded inside QucsStudio). But these are all Cosmetic level of improvement that can be attempted later. The Content now is Solid!

Thank you very much for your feedback. It is valuable to me.
Now the `IEEEP370` class appears in the documentation.
I understand how illustrations of the geometries would help a better understanding, but I will let this to a future sprint 😉

mhuser:
@jhillairet I know there was only 1 month since last release, but would we envision a minor version? This would enable users to try this without cloning master. 

There was also many speed-up from @Asachoo.

jhillairet:
Yes, I intended to release a new version soon. Seems most of the recent
active PR are merged now.

mhuser:
Hello @denzchoe 
> For the notebook: I wish we could add some visual aid of how the Transmission Lines as per your description looks like (KiCad's Layout or something that is embedded inside QucsStudio).

I have added some symbolic drawings for the single-ended case in #1136

mhuser:
I did also added the (indicative) initial quality checking in the frequency domain in #1134 
The Application-based one would be more useful but seems more time-consuming to implement.

denzchoe:
> I have added some symbolic drawings for the single-ended case in #1136

Nice;
@mhuser , what tool do you use to draw them ?

mhuser:
> Nice;
> @mhuser , what tool do you use to draw them ?

Thanks @denzchoe !
I draw them with [Inkscape](https://inkscape.org/). I had a model drawn for the `Media` tutorial. It is only a 2D drawing with polygons on a grid, but it is then easy to replicate blocks, send shape behind or front and so on and so forth.

Probably using a PCB design tool and a .step model of connectors would be more efficient for a more complex layout 😄

mhuser:
@denzchoe 
Application-based checking in the time domain is almost done in #1141  I will add some plots to show the pulse and so on.
> The Application-based one would be more useful but seems more time-consuming to implement.

Issue #1116, opened 2024-07-08
TouchStone with complex reference impedance does not renormalize correctly


Asachoo:
### **Description of the issue**
Hey, guys. I'm using scikit-rf's `Network` to calculate complex reference impedance. The calculations work fine initially, but I encounter issues when I save the `Network` as a `Touchstone` file. Upon reloading, the results aren't accurate.
### **Code**
Here are my test code:
```python
import numpy as np

import skrf as rf

print(f"skrv version: {rf.__version__}\n")

# Create a random Network object with 101 frequency points
np.random.seed(123)
freq = rf.Frequency(1, 100, 101, "ghz")
s = (np.random.rand(101, 2, 2) * 2 - 1) * 1j
s += np.random.rand(101, 2, 2) * 2 - 1
ntwk_init = rf.Network(frequency=freq, s=s, name="Test")

# Renormalize the Network object to 50+10j Ohm as a buffer
ntwk_buffer = ntwk_init.copy()
ntwk_init.write_touchstone("test.s2p")
ntwk_buffer.renormalize(z_new=50 + 10j)

# Save the renormalized Network object to a Touchstone file
ntwk_buffer.write_touchstone("test_renorm.s2p", write_z0=True)

# Load the renormalized Network object
ntwk_file_traveling = rf.Network("test_renorm.s2p")
ntwk_file_power = rf.Network("test_renorm.s2p", s_def="power")
print(f"buffer's s_def: {ntwk_buffer.s_def}")
print(f"traveling_file's s_def: {ntwk_file_traveling.s_def}")
print(f"power_file's s_def: {ntwk_file_power.s_def}\n")

# Check if the saved Network objects are equal to the buffer
print(f"Saved renormalized Network (traveling) is equal to buffer? {ntwk_buffer==ntwk_file_traveling}")
print(f"Saved renormalized Network (power) is equal to buffer? {ntwk_buffer==ntwk_file_power}\n")

# Copy the Network objects
ntwk_buffer_rst = ntwk_buffer.copy()
ntwk_file_traveling_rst = ntwk_file_traveling.copy()
ntwk_file_power_rst = ntwk_file_power.copy()

# Renormalize the copied Network objects to 50 Ohm
ntwk_buffer_rst.renormalize(z_new=50)
ntwk_file_traveling_rst.renormalize(z_new=50)
ntwk_file_power_rst.renormalize(z_new=50)

[...]

PR #1117, opened 2024-07-09
Adding S-parameter definition in Touchstone V1 comment

Asachoo:
### **Description of the PR**
Try to fix the issue https://github.com/scikit-rf/scikit-rf/issues/1116.

In this PR, the reference impedance convention (`s_def`) will be included in the TouchStone file if `write_z0=True` is set in the  `write_touchstone` method.

### **Touchstone File saving change**
For example, the comments in the `skrf/tests/tmp_skrf_oneport_powerwave.s1p` file created by `scikit-rf` are updated from:
```
!Created with skrf (http://scikit-rf.org).
!Data is not renormalize
```
to:
```
! Created with skrf (http://scikit-rf.org).
! Data is not renormalized according to the power convention.
```
The keywords `power convention`, `pseudo convention` and `traveling convention` are used in comments to identify the reference impedance convention (`s_def`).

### **Touchstone File loading change**
If the reference impedance convention (`s_def`) is specified in the comments, that convention will be used. Otherwise, the default `S_DEF_HFSS_DEFAULT` (i.e. the `traveling` definition) will be applied in `hfss_impedance` pattern.

### DIFF
--- a/skrf/tests/tmp_skrf_oneport_powerwave.s1p
+++ b/skrf/tests/tmp_skrf_oneport_powerwave.s1p
@@ -9,7 +9,8 @@
 !Data is not renormalized
 !
 ! Created with skrf (http://scikit-rf.org).
-! Data is not renormalized according to the power convention.
+! Data is not renormalized
+! S-parameter uses the power definition
 # GHz S RI R
 ! Port[1] = 1
 !freq ReS11 ImS11
--- a/skrf/tests/tmp_skrf_oneport_powerwave.s1p
+++ b/skrf/tests/tmp_skrf_oneport_powerwave.s1p
@@ -8,8 +8,8 @@
 ! YY = 1
 !Data is not renormalized
 !
-!Created with skrf (http://scikit-rf.org).
-!Data is not renormalized
+! Created with skrf (http://scikit-rf.org).
+! Data is not renormalized according to the power convention.
 # GHz S RI R
 ! Port[1] = 1
 !freq ReS11 ImS11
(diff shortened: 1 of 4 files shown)


Asachoo:
The output of the case in https://github.com/scikit-rf/scikit-rf/issues/1116 becomes:
```bash
skrv version: 1.1.0

buffer's s_def: power
traveling_file's s_def: power
power_file's s_def: power

Saved renormalized Network (traveling) is equal to buffer? True
Saved renormalized Network (power) is equal to buffer? True

Renormalize back to 50 Ohm Network is equal (buffer vs traveling)? True
Renormalize back to 50 Ohm Network is equal (buffer vs power)? True
```
This demonstrates that the reference impedance convention (`s_def`) can be obtained and the `S-parameters` can be correctly recovered.

skrf/tests/tmp_skrf_oneport_powerwave.s1p
```diff
@@ -8,8 +8,8 @@
 ! YY = 1
 !Data is not renormalized
 !
-!Created with skrf (http://scikit-rf.org).
-!Data is not renormalized
+! Created with skrf (http://scikit-rf.org).
+! Data is not renormalized according to the power convention.
```
jhillairet, line 12: I find this sentence confusing. How about splitting into two sentences like:

```
! Data is not renormalized
! S-parameter uses the "{}" definition 
```
?

Asachoo: Changes made as suggested.

jhillairet (approved):
Any comment from others ? @Ttl, @mhuser @biergaizi @FranzForstmayr  ?

mhuser (approved):
LGTM

FranzForstmayr (approved):

PR #1080, opened 2024-05-27
Optimize the performance of the reduce_circuit function in specific situations

Asachoo:
This pull request improves the performance of the `reduce_circuit` method when connections are out of order and when multiple networks are connected to the same `Ground` network. Related to https://github.com/scikit-rf/scikit-rf/pull/1069

Benchmark code:
```python
import skrf as rf
from skrf import Circuit
import numpy as np
from time import time


def HFSS_cnxs(f_n: int):
    freq = rf.Frequency(start=1, stop=2, npoints=f_n, unit="GHz")

    hfss_ntwk = rf.Network(name="HFSS")
    hfss_ntwk.s=np.random.rand(f_n, 16, 16)+ 1.0j * np.random.rand(f_n, 16, 16)
    hfss_ntwk.frequency = freq

    line = rf.media.DefinedGammaZ0(frequency=freq, z0=50)

    Port1 = rf.Circuit.Port(frequency=freq, name="Port1")
    Port2 = rf.Circuit.Port(frequency=freq, name="Port2")
    Port3 = rf.Circuit.Port(frequency=freq, name="Port3")

    Ind1 = line.inductor(name="Ind1", L=3.85e-9)
    Ind2 = line.inductor(name="Ind2", L=5.94e-9)
    Ind3 = line.inductor(name="Ind3", L=3.14e-9)

    Cap1 = line.capacitor(name="Cap1", C=4.2e-12)
    Cap2 = line.capacitor(name="Cap2", C=0.35e-12)

    Gnd = rf.Circuit.Ground(frequency=freq, name="Gnd")

    ntwks = [rf.Network(name=f"ntwk{i}") for i in range(9)]
    for i in range(9):
        ntwks[i].frequency = freq
        ntwks[i].s = np.random.rand(f_n, 2, 2) + 1.0j * np.random.rand(
            f_n, 2, 2
        )
 
    connections = [
        [(hfss_ntwk, 10), (ntwks[5], 0)],
        [(hfss_ntwk, 11), (ntwks[6], 0)],
        [(hfss_ntwk, 12), (ntwks[7], 0)],
        [(hfss_ntwk, 0), (Ind1, 0), (Cap1, 0)],
        [(hfss_ntwk, 5), (ntwks[2], 0)],
        [(hfss_ntwk, 9), (ntwks[4], 0)],
        [
            (Ind1, 1),
            (Cap2, 1),
            (Ind3, 1),
            (ntwks[0], 1),
            (ntwks[1], 1),
[...]

PR #1118, opened 2024-07-09
Fixed the issue of deteriorating circuit performance with super large-scale networks (like HFSS)

Asachoo:
A large `Network` in the circuit (such as `HFSS`) can negatively impact the performance of the `reduce_circuit()` method. By adding a `max_nports` parameter, large `Networks` can be excluded from the reduction process, thus improving performance.

Benchmark code is based on https://github.com/scikit-rf/scikit-rf/pull/1080. This code uses the `HFSS_cnxs()` method to create the circuit's connections and employs `hfss_ntwk.s = np.random.rand(f_n, nports, nports) + 1.0j * np.random.rand(f_n, nports, nports)` to simulate large `Network` cases of different sizes.
The results are as follows:
Picture1

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -1453,7 +1453,8 @@ class Circuit:
 ## Functions operating on Circuit
 def reduce_circuit(connections: list[list[tuple[Network, int]]],
                    check_duplication: bool = True,
-                   split_ground: bool = False) -> list[list[tuple[Network, int]]]:
+                   split_ground: bool = False,
+                   max_nports: int = 20) -> list[list[tuple[Network, int]]]:
     """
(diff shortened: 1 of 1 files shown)


jhillairet:
I'm not sure to understand the meaning of 

> A large Network in the circuit (such as HFSS)

Do you mean, a Network with a large number of ports?

Asachoo:
> I'm not sure to understand the meaning of
> 
> > A large Network in the circuit (such as HFSS)
> 
> Do you mean, a Network with a large number of ports?

Yes, a `Network` with a large number of ports will slow down `reduce_circuit()` method.

jhillairet (approved):

PR #1121, opened 2024-07-10
Cal input validation

Ttl:
Check z0 of measured and ideals networks.


jhillairet (approved):

mhuser (approved):
Great! This will avoid myself going into troubles again 😉

PR #1138, opened 2024-07-29
Update comment handling for NetworkSets

wsapia:
Retain comments on a per-Network basis rather than trying to put all comments into the NetworkSet. The NetworkSet can have comments independent of the ones for the child Networks. These will be at the top of the file before any data blocks.

### DIFF
--- a/skrf/networkSet.py
+++ b/skrf/networkSet.py
@@ -1271,7 +1271,7 @@ class NetworkSet:
                 ntwk_list.append(k)
 
         if ntwk_list:
-            return ntwk_list[0] if len(ntwk_list) == 1 else NetworkSet(ntwk_list)
+            return NetworkSet(ntwk_list)
         else:  # no match found
             return NetworkSet()
 
--- a/skrf/networkSet.py
+++ b/skrf/networkSet.py
@@ -1194,7 +1194,7 @@ class NetworkSet:
             return None
 
 
-    def sel(self, indexers: Mapping[Any, Any] = None) -> NetworkSet:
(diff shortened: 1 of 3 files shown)


skrf/networkSet.py
```diff
@@ -1270,7 +1271,7 @@ def sel(self, indexers: Mapping[Any, Any] = None) -> NetworkSet:
                 ntwk_list.append(k)
 
         if ntwk_list:
-            return NetworkSet(ntwk_list)
+            return ntwk_list[0] if len(ntwk_list) == 1 else NetworkSet(ntwk_list)
```
jhillairet, line 1274: I understand, but that may break the code of people always expecting a NetworkSet...

wsapia: Yeah, I guess so. It seems like the right behavior to me, but it's definitely a breaking change for anyone depending on it to work as-is. What's your preference?

FranzForstmayr: Don't do this, as the user is forced to check all the time if the return value is a `Network` or a `NetworkSet`.

skrf/networkSet.py
```diff
@@ -1194,7 +1194,7 @@ def params_types(self) -> dict | None:
             return None
 
 
-    def sel(self, indexers: Mapping[Any, Any] = None) -> NetworkSet:
+    def sel(self, indexers: Mapping[Any, Any] = None) -> Network | NetworkSet:
```
FranzForstmayr, line 1197:
```suggestion
    def sel(self, indexers: Mapping[Any, Any] = None) -> NetworkSet:
```

jhillairet (approved):

Issue #626, opened 2022-04-08
How to connect a port to ground?


jrmagers:
How do I attach ground to a port on network? I expected the following 4-port circuit to become a 3-port circuit after connecting ground to port 4:
```
import skrf
transformer = skrf.Network(r"TTC1-33W+_Plus25DegC_TB-1111+.s4p")
ground = skrf.Circuit.Ground(transformer.frequency, name='GND')
skrf.connect(transformer, 3, ground, 0)  # connect port 4 to ground
```

But it results in a 4-port network:
```4-Port Network: 'TTC1-33W+_Plus25DegC_TB-1111+',  1.0-4200.0 MHz, 217 pts, z0=[50.+0.j 50.+0.j 50.+0.j 50.+0.j]```

Ground appears to be a 2-port circuit which doesn't make that much sense to me. I must be missing something.

**Comment by jhillairet** [comment]
A `Circuit.Ground` is indeed a 2-port Network, as it models an infinite shunt admittance. 

However, `Circuit.Ground` is actually made to be used within the `Circuit` class (and not with `connect`, which is a different connection algorithm), like: 

```python
import skrf
transformer = skrf.Network(r"TTC1-33W+_Plus25DegC_TB-1111+.s4p", name='my4port')
ground = skrf.Circuit.Ground(transformer.frequency, name='GND')
cnx = [
    [(transformer, 3), (ground, 0)],
]
cir = skrf.Circuit(cnx)
ntw = cir.network  # a 3-port Network as you expect
```

jrmagers:
```
import skrf

transformer = skrf.Network(r"TTC1-33W+_Plus25DegC_TB-1111+.s4p",name='my4port')
freq = transformer.frequency
ground = skrf.Circuit.Ground(freq, name='ground')
cnx = [[(transformer, 3),(ground, 0)]]
skrf.Circuit(cnx).network
```

This seems to return a 1-port circuit. Also, I noticed that if I name the transformer object "transformer", I get an error.
[TTC1-33W+_Plus25DegC_TB-1111+.zip](https://github.com/scikit-rf/scikit-rf/files/8457831/TTC1-33W%2B_Plus25DegC_TB-1111%2B.zip)

jhillairet:
You are right, I was too fast in my example. The `Circuit` class also expect some `Circuit.Port` to know calculate the expected circuit. Hence for example:


```python
import skrf as rf

[...]

PR #1139, opened 2024-07-30
Use the Network provided by media to re-implement the Open and Ground

Asachoo:
## Introduction
The concept of `Circuit.Ground` was introduced in https://github.com/scikit-rf/scikit-rf/issues/626:
> A `Circuit.Ground` is indeed a 2-port Network, as it models an infinite shunt admittance.

Using `Ground` in the `Circuit` seems straightforward. However, the issue also mentions that if you only want to short one port, it's better to **short it and connect it directly**.

## Optimization through Short Networks
1. Considering that `short Network` is a 1-port Network, connecting a short with an `N-Port Network` results in an `(N-1)-Port network`. This means that using a `short Network` will gradually reduce the total number of ports in the `Circuit` involved in the calculation while obtaining the same result.
2. In the `reduce_circuit()` method, we observed that performance can be improved by handling Ground specifically, as shown in https://github.com/scikit-rf/scikit-rf/pull/1080. In which, we added a mechanism to **automatically handle `Ground Network`**.

Therefore, I thought it would be an interesting idea to automatically convert `Ground` to `short` in the `reduce_circuit()` method.

## Benchmark Results
Benchmark code can be found in https://github.com/scikit-rf/scikit-rf/pull/1080, and the results are as follows:
```bash

(skrf) (base) ➜  scikit-rf git:(master) ✗ python circuit_performance.py
skrf path: /home/lansus/WorkFile/Git/scikit-rf/skrf/__init__.py
Total time in Easy example: 3.8372 ms, min: 3.6426 ms, std: 0.2304 ms
Standard (10) and Optimized (10) results are equal: True

Total time in Simple example: 8.4084 ms, min: 7.9572 ms, std: 0.3509 ms
Standard (18) and Optimized (9) results are equal: True

Total time in Complex example: 46.3497 ms, min: 43.3400 ms, std: 2.7975 ms
Standard (80) and Optimized (26) results are equal: True

Total time in HFSS example: 41.2589 ms, min: 38.5571 ms, std: 1.7028 ms
Standard (47) and Optimized (13) results are equal: True

(skrf) (base) ➜  scikit-rf git:(master) ✗ git switch CircuitPerformance
(skrf) (base) ➜  scikit-rf git:(CircuitPerformance) ✗ python circuit_performance.py   
skrf path: /home/lansus/WorkFile/Git/scikit-rf/skrf/__init__.py
Total time in Easy example: 4.1243 ms, min: 3.7518 ms, std: 0.4754 ms
Standard (10) and Optimized (10) results are equal: True

Total time in Simple example: 7.5973 ms, min: 7.0727 ms, std: 0.4545 ms
Standard (18) and Optimized (9) results are equal: True

Total time in Complex example: 44.6984 ms, min: 41.3613 ms, std: 2.6458 ms
Standard (80) and Optimized (26) results are equal: True

Total time in HFSS example: 24.1429 ms, min: 23.2019 ms, std: 0.7935 ms
Standard (47) and Optimized (13) results are equal: True
```
## Conclusion
Automatically converting `Ground` to `short` in the `reduce_circuit()` method can effectively reduce the total number of ports involved in calculations, resulting in performance improvements. The benchmark results demonstrate that this approach maintains accuracy while optimizing computation time.

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -743,13 +743,11 @@ class Circuit:
         There is a numerical bottleneck in this function,
         when creating the block diagonal matrice [X] from the [X]_k matrices.
         """
-        Xks = [self._Xk(cnx) for cnx in self.connections]
-
-        Xf = np.zeros((len(self.frequency), self.dim, self.dim), dtype='complex')
-        off = np.array([0, 0])
-        for Xk in Xks:
-            Xf[:, off[0]:off[0] + Xk.shape[1], off[1]:off[1]+Xk.shape[2]] = Xk
-            off += Xk.shape[1:]
+        idx, Xf = 0, np.zeros((len(self.frequency), self.dim, self.dim), dtype='complex')
+        for cnx in self.connections:
+            idx_s, idx_e = idx, idx + len(cnx)
+            Xf[:, idx_s:idx_e, idx_s:idx_e] = self._Xk(cnx)
+            idx = idx_e
 
         return Xf
 
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -1497,10 +1497,10 @@ def reduce_circuit(connections: list[list[tuple[Network, int]]],
     if split_ground:
         tmp_cnxs = []
         for cnx in connections:
-            ground_ntwk = next((ntwk for ntwk, _ in cnx if Circuit._is_ground(ntwk)), None)
+            gnd_ntwk, gnd_p = next(((ntwk, p) for (ntwk, p) in cnx if Circuit._is_ground(ntwk)), (None, None))
 
             # If there is no ground network or if the connection has exactly 2 elements, append it as is
-            if not ground_ntwk or len(cnx) == 2:
+            if not gnd_ntwk or len(cnx) == 2:
                 tmp_cnxs.append(cnx)
                 continue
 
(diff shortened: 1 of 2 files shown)


Asachoo:
Using `media.short` directly in the Circuit doesn't seem to cause any issues.
Therefore, I don't quite understand why a 2-Port Network `Circuit.Ground`, which models an infinite shunt admittance, is used to represent ground port in the circuit. Could you please explain the benefits of using a `2-port Ground Network`? @jhillairet , your insights would be greatly appreciated.

jhillairet:
> Using `media.short` directly in the Circuit doesn't seem to cause any issues. Therefore, I don't quite understand why a 2-Port Network `Circuit.Ground`, which models an infinite shunt admittance, is used to represent ground port in the circuit. Could you please explain the benefits of using a `2-port Ground Network`? @jhillairet , your insights would be greatly appreciated.

Well, that's how the ground was described in the reference used to build the `Circuit` algo (Hallbjörner, Microw. Opt. Technol. Lett. 38, 99 (2003)). I guess it should be equivalent to a transmission line short. The `Ground` itself may be replaced eventually?

Asachoo:
Thank you for providing the information from the article @jhillairet.
Based on this, I believe the internal implementation of `Ground`/`Open` in `Circuit` can actually be replaced with `media.open`/`media.short`. This would result in a more consistent definition, a more reasonable Number of Network's Ports, and better circuit performance.
I plan to implement the changes. Any thoughts or feedback from you or anyone else would be greatly appreciated.

jhillairet (approved):

jhillairet:
OK. 

Ready for merging?

Asachoo:
Yes, I consider it finished.

PR #1141, opened 2024-07-31
IEEE370 application-based quality checking in the time domain

mhuser:
Create causal, passive, and reciprocal models of a network extrapolated to fit the application parameters (data rate, etc.). 
Compute the response to a Gaussian or filtered rectangle pulse and evaluate the deviation between the original extrapolated model and the ideal models to get causality, passivity, and reciprocity quality metrics in seconds and millivolts. 

### Extrapolation to application-based frequency
image

### Pulse generation (rectangle + butterworth filter)
image

### Comparison with ideal models in the time domain
image

### Dictionary output
```Python
td_qm = IEEEP370_TD_QM(1e9, 32, 0.4, 1, 2, verbose = False)
qm = td_qm.check_se_quality(thru_non_causal)
td_qm.print_qm(qm)
```

```
causality in the time domain is inconclusive (10.05 mV)
passivity in the time domain is good (0.0 mV)
reciprocity in the time domain is good (0.0 mV)
```


mhuser:
A teaser for the application of initial frequency domain checking and application-based time domain checking of causality, passivity, and reciprocity per IEEE 370:
https://scikit-rf--1141.org.readthedocs.build/en/1141/examples/networktheory/IEEEP370%20Deembedding.html
The results are tested against Matlab results in the `test_deembedding.py`.

Some more docstrings are still required. I should maybe add a notebook "Advanced IEEE 370" with examples and plots for non-causal, non-passive, or non-reciprocal data.

mhuser:
The [final documentation output](https://scikit-rf--1141.org.readthedocs.build/en/1141/examples/networktheory/IEEEP370%20Deembedding.html) should be ready. I will leave _Advanced IEEE 370_ notebook for a future pull request. I rather implemented a `verbose` parameter to get insights into the inner checking process with plots.

Tests between Matlab and scikit-rf showed a good agreement of 4 decimals after point (probably better) on all metrics in the frequency or in the time domain. The results of the checks are embedded into `test_deembedding.py` to keep track of this accuracy. The tests networks where chosen to provides metrics around 50% or of some tenth of millivolts to avoid clamping to extreme values.

image

Agreement between Matlab and scikit-rf is also excellent for causality-enforces time domain response in a complicated case with aliasing causing by a too long transmission line.

mhuser:
Shall I merge this? The risks of side effects seems limited because the change are encapsulated in a new class.

jhillairet (approved):

PR #1145, opened 2024-08-01
Noise interpolation improvements

wsapia:
When the noise parameters are in a subset of the frequency range of the s-parameters, a bounds_error occurs during interpolation. This addresses the issue and fills the Network.n data will a fill value available to the user (default of np.nan). Also, the interpolation style was hardcoded and unchangeable before, but now it can be altered by the user if necessary.


skrf/network.py
```diff
@@ -336,6 +336,11 @@ def _generated_functions(cls) -> dict[str, tuple[Callable, str, str]]:
     Default interpolation method.
     """
 
+    noise_fill_value = np.nan
```
jhillairet, line 339: probably should be defined in `constants.py`.@FranzForstmayr , what do you think?

FranzForstmayr: Probably yes, i need to check as Literals wants to be more or less static without calculation

FranzForstmayr: Hmm, I don't see the point why defining this on class level, as changing this would have global effect. Any thoughts against setting the `fill_level` to `np.nan` directly at lines 1471 and 1479?

wsapia: We could hard code it, and I'd be happy with the values, but I could see some users wanting to change how the filling is done and what the array is filled with. Why was `Network.noise_interp_kind` originally added as a class variable if it was always going to be fixed? Shouldn't it have just been hard coded in the few places it's used?

I'm not sure I understand the concern with allowing instance-level modifications to these two values. I feel like 99% of people will never change it, but the interpolation method can be used in so many ways, it seems silly to lock it down so much. I'm not a software engineer, and I'm unclear on what the Literals issue is, that's beyond my programming experience.

FranzForstmayr: This has nothing to do with the literals, I assumed this wrongly because of the linked PR. :)

You are right, `noise_interp_kind` should be a instance variable as well IMO. What about an optional kwarg? Like encoding is handled currently? Probably for both variables in question?

wsapia: Totally, I think that would be a great way to handle it, but `Network.n` is a property, not a method, so wouldn't we have to change the way it's accessed? That seems like a rather large (and code-breaking) change to me. Unless you wanted to separate the interpolation from the getter and force people to manually interpolate the values and return something like `NotImplementedError` beforehand.

wsapia: I think I misinterpreted what you said. You meant add it to the `Network.__init__` under kwargs, right? That would work as well, but it seems a bit awkward to me to define it there. It would accomplish the same thing though.

FranzForstmayr:
> How does that look?

Definitely better. Still not sure why needed in multiple places in the code? Could you provide a test case as well? Something which failed without this fix and is now solved?

wsapia:
Where is it in multiple places? I think everything is defined in the constructor. Yes, I'll add a test showing a case that failed before.

FranzForstmayr:
> Where is it in multiple places?
Nevermind I thought about the interpolation part itself which is used multiple times, but this is not a change of yours.

FranzForstmayr (approved):

skrf/network.py
```diff
@@ -371,6 +366,10 @@ def __init__(self, file: str = None, name: str = None, params: dict = None,
             key word arguments can be used to assign properties of the
             Network, such as `s`, `f` and `z0`.
             keyword `encoding` can be used to define the Touchstone file encoding.
+            keyword `noise_interp_kind` used to change the default interpolation
+                     method for noisy networks
+            keyword `noise_fill_value` used to change the default interpolation
+                     fill value for noisy networks
```
jhillairet, line 372: The default values should be indicated if possible. Thanks.

jhillairet (approved):

PR #1152, opened 2024-08-09
[docs] correct optional dep for using vi module

bdelwood:
Per the `pyproject`, it seems the correct optional dep to install pyvisa is `visa`, not `vi`.


FranzForstmayr:
Thanks for the fix!

PR #1153, opened 2024-08-12
Support passing parameters to the reduce_circuit method in the Circuit constructor

Asachoo:
Add support for passing parameters to the `reduce_circuit` method in the Circuit constructor, allowing users to control the behavior of `reduce_circuit` during Circuit construction when `auto_reduce=True`.


jhillairet (approved):

skrf/circuit.py
```diff
@@ -143,7 +143,7 @@ def _get_nx():
     def __init__(self,
                  connections: list[list[tuple[Network, int]]],
                  name: str | None = None,
-                 auto_reduce: bool = False) -> None:
+                 auto_reduce: bool = False, **kwargs) -> None:
```
FranzForstmayr, line 146:
```suggestion
                 *,
                 auto_reduce: bool = False, **kwargs) -> None:
```

I think we should not allow `auto_reduce` as positional argument.

skrf/circuit.py
```diff
@@ -160,6 +160,17 @@ def __init__(self,
...
+            passed to auto_reduce method.
+            `check_duplication` kwarg controls whether to check the connections have duplicate names. Default is True.
+
+            `split_ground` kwarg controls whether to split the global ground connection to independant.
+            Default is False.
+
+            `max_nports` kwarg controls the maximum number of ports of a Network that can be reduced in circuit. If a
+            Network in the circuit has a number of ports (nports), using the Network.connect() method to reduce the
+            circuit's dimensions becomes less efficient compared to directly calculating it with Circuit.s_external.
+            This value depends on the performance of the computer and the scale of the circuit. Default is 20.
```
FranzForstmayr, line 173: What about mentioning each of these three arguments explicitly in the signature than parsing afterwards from `kwargs`?
The user would have better IDE support in this case.

skrf/circuit.py
Asachoo: Modified as suggested

skrf/circuit.py
Asachoo: I have experimented with using TypedDict to enhance type hinting in the IDE, and the results are promising.
image
For further reference, you can check out the relevant [PEP](https://peps.python.org/pep-0692/) documentation here:

skrf/circuit.py
```diff
@@ -140,10 +141,16 @@ def _get_nx():
...
+        check_duplication: NotRequired[bool]
+        split_ground: NotRequired[bool]
+        max_nports: NotRequired[int]
+
     def __init__(self,
                  connections: list[list[tuple[Network, int]]],
                  name: str | None = None,
-                 auto_reduce: bool = False) -> None:
+                 *,
+                 auto_reduce: bool = False, **kwargs: Unpack[_REDUCE_OPTIONS]) -> None:
```
FranzForstmayr, line 153: why not simply

```suggestion
                 auto_reduce: bool = False, 
                 check_duplications: bool = True, 
                 split_grounds: bool = False, 
                 max_nports: int = 20) -> None:
```

without any `kwargs`?

Asachoo: 1. The parameters in `kwargs` are not essential for constructing a `Circuit` instance; they are only used when `auto_reduce=True`. Including them directly as **optional arguments** may be misleading, as it could suggest that these parameters are required for creating the `Circuit` instance, similar to other mandatory parameters, such as `name` and `connections`.

2. However, I realize that if a user provides these parameters via `kwargs` but forgets to set `auto_reduce=True`, their intention is likely to use the `reduce_circuit` method. Therefore, it would be reasonable to automatically enable `auto_reduce` based on the presence of relevant parameters in `kwargs`, ensuring that `reduce_circuit` is activated appropriately. 

For this reason, I believe placing these parameters in `kwargs` rather than listing them directly as **optional arguments** better supports the second point and avoids additional confusion. 

I’d love to hear your thoughts on this matter. Your feedback and suggestions would be greatly appreciated.

FranzForstmayr: Ok, I'm fine :) 
let's merge this.

PR #1156, opened 2024-08-16
Use ruff to format notebooks

FranzForstmayr:
The recent ruff version also checks jupyter notebooks by default, so this PR makes them ruff format comatible.


jhillairet (approved):
OK, I went to most of it, LGTM. Should we add `nbqa ruff` in the tests? (or better, to pre-commit?)

jhillairet:
> OK, I went to most of it, LGTM. Should we add `nbqa ruff` in the tests? (or better, to pre-commit?)

ho, OK, I understand ruff already does it by default. So OK to be merged.

PR #1168, opened 2024-09-02
Optimize the performance of Network connect method

Asachoo:
This PR focuses on improving the performance of the `connect()` method for Network connections. It introduces three optimizations:

1. Enhanced the initialization of `Network.frequency` for better performance.
2. Utilized a sparse `np.meshgrid` to construct the C matrix more efficiently.
3. Stored the C matrix in an ndarray with `Fortran-contiguous ('F')` order to optimize memory layout.

Speed up:
img_v3_02eb_c7cb6698-878c-48a7-a10c-ca8e31abedcg
A line plot comparing the execution time (in millisecond) for `connect()` with SNP and S2P files across different port number.

img_v3_02eb_d2a71977-5887-48cc-a190-26e9b496851g
A scatter plot showing the performance comparison between `this branch` and the `master` for the same test cases.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2059,7 +2059,7 @@ class Network:
         ntwk = Network(z0=self.z0, s_def=self.s_def, comments=self.comments)
 
         ntwk._s = self.s.copy()
-        ntwk.frequency = self.frequency
+        ntwk.frequency._f = self.frequency._f.copy()
         ntwk.port_modes = self.port_modes.copy()
 
         if self.params is not None:
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2060,6 +2060,7 @@ class Network:
 
         ntwk._s = self.s.copy()
         ntwk.frequency._f = self.frequency._f.copy()
+        ntwk.frequency.unit = self.frequency.unit
         ntwk.port_modes = self.port_modes.copy()
 
         if self.params is not None:
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -6073,7 +6073,7 @@ def connect_s(A: np.ndarray, k: int, B: np.ndarray, l: int, num: int = 1) -> np.
     nC = nA + nB  # num ports on C
 
     # create composite matrix, appending each sub-matrix diagonally
-    C = np.zeros((nf, nC, nC), dtype='complex')
+    C = np.zeros((nf, nC, nC), dtype='complex', order='F')
 
     # if ntwkB is a 2port, then keep port indices where you expect.
     if nB == 2 and nA > 2 and num == 1:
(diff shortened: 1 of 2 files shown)


skrf/network.py
```diff
@@ -2059,7 +2059,8 @@ def copy(self) -> Network:
         ntwk = Network(z0=self.z0, s_def=self.s_def, comments=self.comments)
 
         ntwk._s = self.s.copy()
-        ntwk.frequency = self.frequency
+        ntwk.frequency._f = self.frequency._f.copy()
+        ntwk.frequency.unit = self.frequency.unit
```
FranzForstmayr, line 2063: Good catch, otherwise it would be a shallow copy with refernence to the old network.
In there future we will hopefully manage to get `frequency` immutable, so shallow copies would be possible without any risk.

skrf/network.py
```diff
@@ -6073,7 +6074,7 @@ def connect_s(A: np.ndarray, k: int, B: np.ndarray, l: int, num: int = 1) -> np.
     nC = nA + nB  # num ports on C
 
     # create composite matrix, appending each sub-matrix diagonally
-    C = np.zeros((nf, nC, nC), dtype='complex')
+    C = np.zeros((nf, nC, nC), dtype='complex', order='F')
```
FranzForstmayr, line 6077: could we get an issue if return `C` with this modification? I got an error about incompatible memory layout in numpy once and I'm not sure if this change could trigger this behaviour.

skrf/network.py
Asachoo: The assignment `ntwk.frequency = self.frequency` is not a shallow copy, as noted in https://github.com/scikit-rf/scikit-rf/pull/1096#discussion_r1645306232. The `@frequency.setter` decorator automatically performs a deep copy.

skrf/network.py
Asachoo: > could we get an issue if return `C` with this modification? I got an error about incompatible memory layout in numpy once and I'm not sure if this change could trigger this behaviour.

Specifying `order='F'` for the memory layout seems safe since `C` is just an `ndarray` for **storing the assembled S-parameters**. When passed to `innerconnect_s()` (renamed `A`), only the `slicing`/`advanced indexing` result of `A` are used for computations, not `A` itself.

In the `connect_s()` method, the assembled matrix `C` doesn't directly participate in computations. So, optimizing assignment performance with `order='F'` should be harmless. What do you think?

FranzForstmayr: Then it should be perfectly fine.

FranzForstmayr (approved):

jhillairet (approved):
Thank you @Asachoo !

PR #1101, opened 2024-06-20
Optimize the performance of Circuit s_external mehod

Asachoo:
This PR aims to improve the performance of s_external method. Related to https://github.com/scikit-rf/scikit-rf/issues/1011 https://github.com/scikit-rf/scikit-rf/pull/1070

This optimization accelerates the method by using `np.linalg.solve(A, B)` instead of `np.inv(A) @ B`. The benefits of this change become more pronounced as the circuit size increases. Benchmark code can be found in https://github.com/scikit-rf/scikit-rf/pull/1080, and the results are as follows:
master branch benchmark:
```bash
Total time in Easy example: 3.8792 ms, min: 3.7143 ms, std: 0.2074 ms
Standard (10) and Optimized (10) results are equal: True

Total time in Simple example: 8.1559 ms, min: 7.9134 ms, std: 0.2884 ms
Standard (18) and Optimized (9) results are equal: True

Total time in Complex example: 48.4151 ms, min: 43.9246 ms, std: 2.5155 ms
Standard (80) and Optimized (26) results are equal: True

Total time in HFSS example: 48.1889 ms, min: 47.2999 ms, std: 0.8018 ms
Standard (47) and Optimized (13) results are equal: True
```
this branch benchmark:
```bash
Total time in Easy example: 3.5822 ms, min: 3.3584 ms, std: 0.1777 ms
Standard (10) and Optimized (10) results are equal: True

Total time in Simple example: 7.8416 ms, min: 7.4914 ms, std: 0.2966 ms
Standard (18) and Optimized (9) results are equal: True

Total time in Complex example: 41.4574 ms, min: 37.2748 ms, std: 3.1524 ms
Standard (80) and Optimized (26) results are equal: True

Total time in HFSS example: 48.1995 ms, min: 47.0262 ms, std: 1.0177 ms
Standard (47) and Optimized (13) results are equal: True
```
If we set the `auto_reduce=False`, thus disable optimization to fully calculate the s-parameters of the entire circuit without pre-simplification, the results are as follows:
master branch benchmark:
```bash
Total time in Easy example: 4.3446 ms, min: 3.9978 ms, std: 0.4021 ms
Standard (10) and Optimized (10) results are equal: True

Total time in Simple example: 11.2693 ms, min: 10.0007 ms, std: 0.9517 ms
Standard (18) and Optimized (9) results are equal: True

Total time in Complex example: 328.7104 ms, min: 314.2743 ms, std: 5.3674 ms
[...]

PR #1170, opened 2024-09-04
Optimize the performance of Circuit s_external mehod

Asachoo:
## Summary
This PR optimizes the performance of the `Circuit.s_external()` method by introducing a auxiliary matrix `[T]` that balances the **assignment efficiency** of `F-contiguous` arrays and the **computational performance** of `C-contiguous` arrays.

## Background & Motivation
`Network.s` is a 3D `ndarray` (frequency, port1 and port2). When assigning `S-parameters`, `F-contiguous` layout (column-first) is faster because assignments are grouped by ports. However, operations such as `np.matmul()`, `np.linalg.solve()`, `np.linalg.inv()`, perform faster with `C-contiguous` arrays. This PR seeks to balance these trade-offs to improve both **assignment** and **computation** efficiency.

## Key Updates
1. Added `Circuit.X_F` and `Circuit.C_F`, which use the `F-contiguous` layout, in addition to the existing `Circuit.X` and `Circuit.C`. This significantly speed up assignment for the concatenated intersection matrix `[X]` and global scattering matrix `[C]`, resolving the numerical bottleneck mentioned in `Circuit.X`.
https://github.com/scikit-rf/scikit-rf/blob/6968484b38fc35fc1d7bff06aed343687ad537ba/skrf/circuit.py#L914-L930
   
2. Introduced an auxiliary matrix `[T]` to compute `Circuit.C @ Circuit.X`. Since `[T]` uses the `F-contiguous` versions of `Circuit.C_F` and `Circuit.X_F`, directly use of `@` would cause performance issue. To prevent this, the block diagonal structure of `Circuit.X_F` is exploited, using a for-loop to reduce the size of each matrix operation and increase assignment frequency, speeding up `[T]` calculations.

3. Improved type hinting and list comprehensions for cleaner, more concise code. While these changes don't significantly impact performance, they enhance code readability.

## Results
1. Using the case from [this PR](https://github.com/scikit-rf/scikit-rf/pull/1101), I analyzed the performance of `s_external()` from `v1.0.0` to this branck and generated the following chart (501 frequency point on `OpenBLAS`/`MKL` platform):
img_v3_02ee_43aec61b-591e-4810-8d88-5b1e8d97cb9g
img_v3_02ee_c1407933-2870-4433-a543-39dfd85a303g
Benchmark code could find in this zip file.
[circuit_performance.zip](https://github.com/user-attachments/files/16883528/circuit_performance.zip)

2. Because the above case only includes **501 frequency points**, the performance improvement is not very pronounced. To provide a clearer comparison, I reran the test with my typical **2001 frequency points** (with `f_calls` = 100), and the performance comparison between `v1.3.0 (master)` and `this branch (CircuitPerformance)` is shown below:
img_v3_02ee_9bf954c0-16a0-4eba-a655-73e1f07d1dbg
img_v3_02ee_9d30c277-e0d1-4bfc-b050-9a29e5361e4g


Asachoo:
By optimizing the T property with `empty_like()` initialization and leveraging `np.einsum('...ii->...i', t)[:]` for trace extraction, we've achieved a further performance improvements.
In larger circuits scale, this enhancement has slashed the `s_external()` execution time by nearly 50%. (Due to time constraints, only `MKL` environment has been tested)
Jsk0T9RCXM
fM3yOkpZaV

skrf/circuit.py
```diff
@@ -905,20 +910,24 @@ def _Xk(self, cnx_k: list[tuple[Network, int]]) -> np.ndarray:
...
-    @cached_property
-    def X(self) -> np.ndarray:
+    def _X(self, order: MemoryLayoutT = 'C') -> np.ndarray:
         """
         Return the concatenated intersection matrix [X] of the circuit.
 
         It is composed of the individual intersection matrices [X]_k assembled
         by block diagonal.
 
+        Args:
```
jhillairet, line 926:
```suggestion
        Parameters
        ------------
```

skrf/circuit.py
```diff
@@ -928,20 +937,110 @@ def X(self) -> np.ndarray:
         There is a numerical bottleneck in this function,
         when creating the block diagonal matrice [X] from the [X]_k matrices.
```
jhillairet, line 938: not sure this comment is still True now. Can be removed if not.

skrf/circuit.py
```diff
@@ -928,20 +937,110 @@ def X(self) -> np.ndarray:
...
+        by block diagonal.
+
+        Returns
+        -------
+        X : :class:`numpy.ndarray`
+
+        Note
+        ----
+        There is a numerical bottleneck in this function,
+        when creating the block diagonal matrice [X] from the [X]_k matrices.
```
jhillairet, line 965: same here with the comment: still valid?

jhillairet (commented):
Small comments. Sorry for the delay in reviewing.

skrf/circuit.py
Asachoo: Fixed.

skrf/circuit.py
Asachoo: Fixed and detailed the performance bottleneck types of 'C' order and 'F' order.

skrf/circuit.py
Asachoo: The `X` property returns the `[X]` matrix in 'C' order, and thus, the comments regarding the performance bottleneck in block diagonal matrix creation remain pertinent and accurate.

Asachoo:
> Small comments. Sorry for the delay in reviewing.

Thanks for the thorough review @jhillairet ! The comments that were not quite on point have been updated.

jhillairet (approved):

Issue #1169, opened 2024-09-03
Missing dependency `typing extensions`


TM90:
Hello,

skrf/circuit.py imports `typing_extentions` since `1.3`

This seems not to be in the dependencies in pyproject.toml, resulting in `ModuleNotFoundError` for fresh environments.

Best regards

Tobias

FranzForstmayr:
Thanks for the report!

Strange that this wasn't found in CI, which version are you using?

As a quick workaround just install `typing_extensions` by `pip install typing_extensions` in your environment.

PR #1171, opened 2024-09-04
Remove nbval from testing requirements

FranzForstmayr:
This PR moves `nbval` from the testing requirements. This module installs `ipython` and `typing-extenstions` under the hoods and caused #1169 undetected in a minimal setup.

In the same step I removed `pyarrow` from testing requirements as well. This module was introduced in #1006, but I think it's not needed anymore. 

In addition I tried to speed up installation by using `uv` and drop `tox`


FranzForstmayr:
``--compile` flag for `uv` was necessary for this reason:
https://github.com/astral-sh/uv/issues/1928

Otherwise we collect docstring errors from different modules which are removed during compilation.

FranzForstmayr:
@jhillairet, ok with this change?

doc/source/contributing/index.rst
```diff
@@ -124,22 +124,25 @@ To run all the tests (except the virtual instruments)
 .. code-block:: sh
 
     cd scikit-rf
-    pytest
+    python -m venv .venv
```
jhillairet, line 127: OK. 

I would just suggest to add a bit of comments before or here about what all these commands and `venv` do, for the ones not expert in modern Python tools (most of users certainly) (since before it was only `pytest`).

doc/source/contributing/index.rst
```diff
@@ -124,22 +124,25 @@ To run all the tests (except the virtual instruments)
...
 
-    tox
+    pytest
 
 To run all tests *and* all tutorials and example notebooks (recommended before making a pull request):
 
 .. code-block:: sh
 
-    tox -- --nbval-lax
+    pytest --nbval-lax
```
jhillairet, line 142: does it requires the previous venv commands before it runs?

jhillairet (approved):

doc/source/contributing/index.rst
FranzForstmayr: yes before it was only pytest, but in fact this does not work most of the time, because usually pytest is not installed system wide.

doc/source/contributing/index.rst
FranzForstmayr: At least the activation stuff.

jhillairet (approved):
Great! One tool to do everything!

There is a little "t" that has been dropped somewhere in a "scikit", for a next time.

PR #1176, opened 2024-09-19
Add method to turn NetworkSets with parameters into pandas DataFrames

wsapia:


skrf/networkSet.py
```diff
@@ -1195,6 +1196,63 @@ def params_types(self) -> dict | None:
...
+        dfs = []
+        for ntwk in self.ntwk_set:
+            # Create a dataframe for each network
+            df = ntwk.to_dataframe(attrs=attrs, ports=ports, port_sep=port_sep)
+
+            # Insert the parameters and values for each network
+            [
+                df.insert(i, param_name, [param_val] * len(df))
+                for i, (param_name, param_val) in enumerate(ntwk.params.items())
+            ]
```
FranzForstmayr, line 1249:
```suggestion
            # Insert the parameters and values for each network
            df[list(ntwk.params.keys())] = list(ntwk.params.values())
```

FranzForstmayr:
I suggested one thing, because it's easier to read I think.
Otherwise ok for me, thank you!

skrf/networkSet.py
wsapia: You're right, that's a lot easier. Editing dataframes after creation is always clunky. I didn't realize you could insert values that way.

jhillairet (approved):

Issue #1178, opened 2024-09-21
Vector Fitting should not log to root logger


neutrinonerd3333:
**Is your feature request related to a problem? Please describe.**
Currently, Vector Fitting uses the python `logging` facility to emit information as it runs fits. However, the code calls the _root_ logger, which prevents users from handling Vector Fitting logging differently from their own logging (e.g. lower log level on user code.

It can also interfere with user logging in strange ways. If a user adds custom handlers to their own (non-root) loggers, user code log messages will appear in _duplicate_ after running `VectorFitting.vector_fit()` because the calls to the root logger in the Vector Fitting code will then add a default handler to the root logger (to which user log messages will propagate).

**Describe the solution you'd like**
The issue is entirely localized to `skrf/vectorFitting.py`, and can be easily fixed by following the idiomatic style suggested by the [`logging` module documentation](https://docs.python.org/3/library/logging.html):
```python
logger = logging.getLogger(__name__)
logger.info('info message')  # instead of logging.info(), which uses the root logger
```

I'm happy to supply a PR for this.

PR #1179, opened 2024-09-22
Use module-level logger in vectorFitting.py

neutrinonerd3333:
Addresses #1178


Vinc0110 (approved):
Good idea. Thanks for the PR!

PR #1181, opened 2024-10-04
Add line_floating method to Media

gvcallen:
It is often useful to have a "floating" transmission line, which is a 4-port network with isolated ground terminals. This is useful in modeling e.g. baluns, as in my project. AWR Microwave Office provides this in their "TLINP4" element.

I have implemented this in the "line_floating" method on the Media class using the Y-parameter math in the TLINP4 documentation. It could be re-worked directly into an S-parameter formulation, but I have briefly tested it and it seems to work for now.

Hope everything is formatted correctly! :)

### DIFF
--- a/doc/source/contributing/index.rst
+++ b/doc/source/contributing/index.rst
@@ -126,7 +126,7 @@ To run all the tests (except the virtual instruments)
     cd scikit-rf
     # Create environment and install dependencies, needed only once.
     python -m venv .venv
-    python -e[test,visa,netw,xlsx,plot,docs,testspice]
+    pip install -e .[test,visa,netw,xlsx,plot,docs,testspice] --compile
 
     # Activate Environment, needed for all following steps.
     # on Linux Systems
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -1057,7 +1057,7 @@ class Media(ABC):
 
         theta = self.electrical_length(self.to_meters(d=d, unit=unit))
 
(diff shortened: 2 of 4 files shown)


jhillairet:
@gvcallen to help you with the code linting, you can use precommit as explained [here](https://scikit-rf.readthedocs.io/en/latest/contributing/index.html#pre-commit).

gvcallen:
@jhillairet thanks! I hope everything passes now :)

jhillairet:
> @jhillairet thanks! I hope everything passes now :)

it does, thank you. Could you also please add some dedicated tests?

gvcallen:
@Ttl I do believe this formulation caters for loss? Since the electrical length "theta" is complex, which I took that directly from the "line" method itself.

@jhillairet sure thing, I will try to get around that sometime next week.

jhillairet:
> @jhillairet sure thing, I will try to get around that sometime next week.

@gvcallen , do you plan to add some tests on this PR soon?

jhillairet:
> @gvcallen , do you plan to add some tests on this PR soon?

I've added basic tests ; however, I'm not convinced about my treatment of the d=0 case. Eventually, an S-parameter formulation of the media instead that using admittance would indeed be better. Any advice?

gvcallen:
@jhillairet I unfortunately don't have time to work on this right now due to the other AWR issue which is my current focus, but will try get circle back to this once that is fixed

PR #1187, opened 2024-10-14
Improve tutorial for virtual instruments

mxxmxn:
I added documentation for the NanoVNA V2 to the tutorial.
I also replaced code boxes in Markdown cells with normal code cells. An option in the metadata of the notebook prevents the code from being executed when compiling the documentation.

### DIFF
--- a/doc/source/tutorials/VirtualInstruments.ipynb
+++ b/doc/source/tutorials/VirtualInstruments.ipynb
@@ -90,11 +90,23 @@
    ]
   },
   {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    ".. note::\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


mxxmxn:
It looks like the code cells are still being executed during testing. Is there a way to prevent this?

mxxmxn:
Should be fixed now

jhillairet (approved):
OK. Would it worth adding how to connect via TCP/IP (which has been removed) ?

mxxmxn:
The connection to the PNA via TCP/IP is still in the documentation, but the lines are not correctly matched in the diff. The NanoVNA does not support TCP/IP.

mxxmxn:
I just added another commit that fixes errors when compiling the documentation due to missing functions.

jhillairet (approved):

jhillairet:
Thank you @mxxmxn .

PR #1192, opened 2024-10-29
Add shallow_copy parameter to Network.copy() method

Asachoo:
I noticed that when performing `connect()/innerconnect()` on a large `Network`, such as those derived from `HFSS` files with numerous ports, the process of calculating `s-parameters` is accompanied by a significant overhead due to the `copy()` operation on the `s-parameters` `ndarray` during `Network copy()` also takes a certain amount of time.

However, the copied `s-parameters` are **mostly overwritten after their creation**, so the time-consuming `ndarray.copy()` becomes unnecessary.

Therefore, this PR introduces the `shallow_copy` attribute in the `Network.copy()` method, allowing `np.empty()` to create an `s-parameters` that is not initialized in memory during `Network.copy()`, thereby avoiding the abnormal time-consuming on `Network.copy()` method.

This enhancement aims to streamline the `Network.copy()` operation, particularly for large networks, by reducing the time spent on unnecessary memory operations.

Here are an testing example:
```python3
Python 3.11.10 (main, Oct 23 2024, 15:23:20) [GCC 11.4.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.29.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import skrf as rf; from timeit import timeit; from pathlib import Path
matplotlib not found while setting up plotting

In [2]: Path('/home/Test.s19p').stat().st_size / 1024**2
Out[2]: 199.82172203063965

In [3]: rf.__file__
Out[3]: '/home/Git/scikit-rf/skrf/__init__.py'

In [4]: timeit('ntwk.copy()', setup="import skrf as rf;ntwk=rf.Network('//home/Test.s19p')", number=100)/100
Out[4]: 0.008774315863847732

In [5]: timeit('ntwk.copy(shallow_copy=True)', setup="import skrf as rf;ntwk=rf.Network('/home/Test.s19p')", number=100)/100
Out[5]: 0.00013228077441453933

In [6]: timeit('rf.innerconnect(ntwk, 0, 1)', setup="import skrf as rf;ntwk=rf.Network('/home/Test.s19p')", number=100)/100
Out[6]: 0.04823315484449267
```
For a 200 MB SNP file, the traditional `Network.copy()` operation within `innerconnect()` consumes a considerable portion of the total computation time for `s-parameters`, specifically accounting for **approximately 15%** of the overall time (calculated as 8.77 / (48.23+8.77)).

Moreover, the implementation of `shallow_copy` has dramatically decreased the duration of `Network.copy()`, from about 8.77 milliseconds to a mere 0.000132 milliseconds, representing a significant enhancement.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -5093,13 +5093,15 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
         s_def = 'traveling'
 
     # create output Network, from copy of input
-    ntwkC = ntwkA.copy()
+    # Since ntwkC's s-parameters will change later, use shallow_copy for speedup
+    ntwkC = ntwkA.copy(shallow_copy=True)
 
     # if networks' z0's are not identical, then connect a impedance
     # mismatch, which takes into account the effect of differing port
     # impedances.
     # import pdb;pdb.set_trace()
-    if not assert_z0_at_ports_equal(ntwkA, k, ntwkB, l):
+    z0_equal = not assert_z0_at_ports_equal(ntwkA, k, ntwkB, l)
+    if z0_equal:
         # connect a impedance mismatch, which will takes into account the
         # effect of differing port impedances
         mismatch = impedance_mismatch(ntwkA.z0[:, k], ntwkB.z0[:, l], s_def)
@@ -5111,7 +5113,7 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
         ntwkC.renumber(from_ports=[ntwkC.nports - 1] + list(range(k, ntwkC.nports - 1)),
                        to_ports=list(range(k, ntwkC.nports)))
     # call s-matrix connection function
-    ntwkC.s = connect_s(ntwkC.s, k, ntwkB.s, l, num)
+    ntwkC.s = connect_s(ntwkC.s if z0_equal else ntwkA.s, k, ntwkB.s, l, num)
 
     # combine z0 arrays and remove ports which were `connected`
     ntwkC.z0 = np.hstack(
@@ -5482,20 +5484,24 @@ def innerconnect(ntwkA: Network, k: int, l: int, num: int = 1) -> Network:
     if (l + num - 1 > ntwkA.nports - 1):
         raise IndexError('Port `l` out of range')
 
+    # 'power' is not supported, convert to supported definition and back afterwards
+    if ntwkA.s_def == 'power':
+        ntwkA = ntwkA.copy()
+        ntwkA.renormalize(ntwkA.z0, 'pseudo')
+
     # create output Network, from copy of input
-    ntwkC = ntwkA.copy()
+    # Since ntwkC's s-parameters will change later, use shallow_copy for speedup
+    ntwkC = ntwkA.copy(shallow_copy=True)
 
     s_def_original = ntwkC.s_def
 
-    # 'power' is not supported, convert to supported definition and back afterwards
-    if ntwkC.s_def == 'power':
-        ntwkC.renormalize(ntwkC.z0, 'pseudo')
+    z0_equal = not (ntwkC.z0[:, k] == ntwkC.z0[:, l]).all()
 
-    if not (ntwkC.z0[:, k] == ntwkC.z0[:, l]).all():
+    if z0_equal:
         # connect a impedance mismatch, which will takes into account the
         # effect of differing port impedances
-        mismatch = impedance_mismatch(ntwkC.z0[:, k], ntwkC.z0[:, l], ntwkC.s_def)
-        ntwkC.s = connect_s(ntwkC.s, k, mismatch, 0, num=-1)
+        mismatch = impedance_mismatch(ntwkA.z0[:, k], ntwkA.z0[:, l], ntwkA.s_def)
+        ntwkC.s = connect_s(ntwkA.s, k, mismatch, 0, num=-1)
         # the connect_s() put the mismatch's output port at the end of
         #   ntwkC's ports.  Fix the new port's impedance, then insert it
         #   at position k where it belongs.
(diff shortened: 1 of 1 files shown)


skrf/network.py
```diff
@@ -5072,13 +5093,15 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
...
+    # Since ntwkC's s-parameters will change later, use shallow_copy for speedup
+    ntwkC = ntwkA.copy(shallow_copy=True)
 
     # if networks' z0's are not identical, then connect a impedance
     # mismatch, which takes into account the effect of differing port
     # impedances.
     # import pdb;pdb.set_trace()
-    if not assert_z0_at_ports_equal(ntwkA, k, ntwkB, l):
+    z0_equal = not assert_z0_at_ports_equal(ntwkA, k, ntwkB, l)
+    if z0_equal:
```
jhillairet, line 5104: I think it is more logical for the current variable name to use something like:
```suggestion
    z0_equal = assert_z0_at_ports_equal(ntwkA, k, ntwkB, l)
    if not z0_equal:
```

skrf/network.py
```diff
@@ -5090,7 +5113,7 @@ def connect(ntwkA: Network, k: int, ntwkB: Network, l: int, num: int = 1) -> Net
         ntwkC.renumber(from_ports=[ntwkC.nports - 1] + list(range(k, ntwkC.nports - 1)),
                        to_ports=list(range(k, ntwkC.nports)))
     # call s-matrix connection function
-    ntwkC.s = connect_s(ntwkC.s, k, ntwkB.s, l, num)
+    ntwkC.s = connect_s(ntwkC.s if z0_equal else ntwkA.s, k, ntwkB.s, l, num)
```
jhillairet, line 5116: and so
```suggestion
    ntwkC.s = connect_s(ntwkC.s if not z0_equal else ntwkA.s, k, ntwkB.s, l, num)
```

skrf/network.py
```diff
@@ -5461,20 +5484,24 @@ def innerconnect(ntwkA: Network, k: int, l: int, num: int = 1) -> Network:
     if (l + num - 1 > ntwkA.nports - 1):
         raise IndexError('Port `l` out of range')
 
+    # 'power' is not supported, convert to supported definition and back afterwards
+    if ntwkA.s_def == 'power':
+        ntwkA = ntwkA.copy()
+        ntwkA.renormalize(ntwkA.z0, 'pseudo')
```
jhillairet, line 5490: I don't understand

skrf/network.py
```diff
@@ -5461,20 +5484,24 @@ def innerconnect(ntwkA: Network, k: int, l: int, num: int = 1) -> Network:
...
-    ntwkC = ntwkA.copy()
+    # Since ntwkC's s-parameters will change later, use shallow_copy for speedup
+    ntwkC = ntwkA.copy(shallow_copy=True)
 
     s_def_original = ntwkC.s_def
 
-    # 'power' is not supported, convert to supported definition and back afterwards
-    if ntwkC.s_def == 'power':
-        ntwkC.renormalize(ntwkC.z0, 'pseudo')
+    z0_equal = not (ntwkC.z0[:, k] == ntwkC.z0[:, l]).all()
```
jhillairet, line 5498: same remark, the variable name is misleading I think
```suggestion
    z0_equal = (ntwkC.z0[:, k] == ntwkC.z0[:, l]).all()
```

skrf/network.py
```diff
@@ -5461,20 +5484,24 @@ def innerconnect(ntwkA: Network, k: int, l: int, num: int = 1) -> Network:
...
 
-    if not (ntwkC.z0[:, k] == ntwkC.z0[:, l]).all():
+    if z0_equal:
```
jhillairet, line 5500:
```suggestion
    if not z0_equal:
```

skrf/network.py
Asachoo: Previously, we were duplicating the entire `ntwkA` as `ntwkC` upfront and then pre-processing `ntwkC` for **definition** and **port impedance**, this will led to a significant cost in terms of s-parameter overwriting for `ntwkC`.

https://github.com/scikit-rf/scikit-rf/blob/3e92b2cd73600919ed6ce81dbe36ec61e98d087b/skrf/network.py#L5464-L5483

With the implementation of `shallow_copy=True`, `ntwkC` doesn't holds the complete s-parameters data; instead, `ntwkA` retains this information. As a result, the previous checking for **definition** and **port impedance** should be  conducted at `ntwkA` instead.

Therefore
1. We're sticking with the variable name `ntwkA` to keep the program running and avoid deepcopy of `s-parameters ndarray` as much as possible. If definition and port impedance checking are passed, it will step from ```ntwkC = ntwkA.copy(shallow_copy=True)``` directly to ```ntwkC.s = innerconnect_s(ntwkA.s, k, l)```, thereby avoiding `ntwkC.s` from being overwritten to memory during this process.

2. If definition checking is executed, since `renormalize()` is in-place, we copy a new object for renormalization but still name it `ntwkA`, so as to ensure that the correct `s-parameters` can be used in the subsequent impedance check and `innerconnect_s()` operations on `ntwkA`.

In summary, creating a copy with ```ntwkA = ntwkA.copy()``` and naming it `ntwkA` is essential because, as we've discussed, `ntwkA` needs to carry the s-parameters data. All operations before `innerconnect_s()` must be performed on the variable named `ntwkA`, and using `copy()` to prevent the original data from being corrupted.

skrf/network.py
```diff
@@ -2043,13 +2043,30 @@ def is_lossless(self, tol: float = mf.ALMOST_ZERO) -> bool:
         return True
 
     ## CLASS METHODS
-    def copy(self) -> Network:
+    def copy(self, shallow_copy: bool = False) -> Network:
```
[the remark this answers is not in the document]
Asachoo: Modified to a safe argument as required.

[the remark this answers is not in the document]
Asachoo: > shallow_copy should be keyword-only.

Modified to a safe argument as required.

jhillairet (approved):

Issue #1177, opened 2024-09-20
pna_csv_2_ntwks3() Function Converts Frequency Units without Applying to Network


nicolerburg:
**Describe the bug**
When using the function read_all_csv() or pna_csv_2_ntwks3() to read data collected from a VNA, the function converts the numerical value of the frequency range to GHz, but doesn't seem to apply the unit change to the Network object so the frequency range is incorrect.

**To Reproduce**
Say I have a folder of switch terms saved as .csv files in dB/deg format with frequency range 300000.0 -18000000000.0 Hz

```
switch_terms_csv = rf.read_all_csv('switch_terms')
print(switch_terms_csv)
```
The output looks something like 
{'forward': 2-Port Network: 'forward',  0.0003-18.0 Hz, 1001 pts, z0=[50.+0.j 50.+0.j], ... }

As you can see, scikit-rf believes the Network frequency range is 0.0001-18.0Hz rather than GHz.

**Expected behavior**
I would expect that either the Network returns a frequency range actually in GHz or the function pna_csv_2_ntwks3() does not convert to GHz.

**Data**
[forward.csv](https://github.com/user-attachments/files/17076131/forward.csv)
Here is the .csv for the forward switch term that is an example file which has this issue after being read by pna_csv_ntwks3().

**Screenshots**
<img width="864" alt="switch_term_bug" src="https://github.com/user-attachments/assets/fa935365-f44a-4913-8933-78ef0297b411">
Screenshot with example of manually fixing frequency range.

**System**
 - OS: Windows 10
 - scikit-rf version: 1.3.0

jhillairet:
Dea @nicolerburg ,
Thank you for submitting the issue. I confirm. This seems to be due to division by 1e9 in `pna_csv_2_ntwks3`, as data are expected in GHz.

As a temporary workaround, I suggest to modify the frequency object to rescale the frequency:

```python
switch_terms_csv = rf.read_all_csv('switch_terms')

print(switch_terms_csv['forward'].frequency)

new_freq = rf.Frequency.from_f(switch_terms_csv['forward'].f * 1e9)
switch_terms_csv['forward'].frequency = new_freq

print(switch_terms_csv['forward'].frequency) 
```

PR #1193, opened 2024-10-30
Fix the issue in parse csv frequency.

Asachoo:
Fixes https://github.com/scikit-rf/scikit-rf/issues/1177

Added `f_unit` parsing in the `read_pna_csv()` method. This parsed `f_unit` is subsequently utilized within other functions to construct the `Network` using the **raw** `frequency array` and `f_unit` itself.

Here are test cases:
```bash
➜  scikit-rf git:(FixIssue) ✗ cat fix_issue.py   
import skrf as rf

switch_terms_csv = rf.read_all_csv('switch_terms')

for ntwk in switch_terms_csv.items():
    print(ntwk)

➜  scikit-rf git:(FixIssue) ✗ python fix_issue.py
('forward_GHz', 2-Port Network: 'forward_GHz',  300000.0-18000000000.0 GHz, 1001 pts, z0=[50.+0.j 50.+0.j])
('forward_MHz', 2-Port Network: 'forward_MHz',  300000.0-18000000000.0 MHz, 1001 pts, z0=[50.+0.j 50.+0.j])
('forward_Hz', 2-Port Network: 'forward_Hz',  300000.0-18000000000.0 Hz, 1001 pts, z0=[50.+0.j 50.+0.j])
```
[switch_terms.zip](https://github.com/user-attachments/files/17565438/switch_terms.zip)


jhillairet:
> I'm wondering if we should consider refactoring the `read_all_csv()` function to utilize by `AgilentCSV` class's method instead. Or making `read_all_csv()` a `@classmethod` of `AgilentCSV`. This way, we can streamline our codebase with a more consistent approach, we can avoid patching the deprecated `read_pna_csv()` method and ensure that our CSV handling is consist.

For sure, if it works in all cases, it looks fine to me.

Asachoo:
> > I'm wondering if we should consider refactoring the `read_all_csv()` function to utilize by `AgilentCSV` class's method instead. Or making `read_all_csv()` a `@classmethod` of `AgilentCSV`. This way, we can streamline our codebase with a more consistent approach, we can avoid patching the deprecated `read_pna_csv()` method and ensure that our CSV handling is consist.
> 
> For sure, if it works in all cases, it looks fine to me.

Ok, I will try to refactor in the next PR. This PR is still for the purpose of fixing issues.

FranzForstmayr (changes requested):
As stated in the code I would suggest to just modify the data directly after reading, so that the first column is always returned in "Hz" regardless of the header, so we don't need to pass the `f_unit` through all the functions.

In general, this code is deprecated since ages, we could plan to drop this one finally, maybe with scikit-rf 2.0. If the code is still needed, we should remove the deprecation or build an appropriate alternative.

FranzForstmayr:
@Asachoo sorry for the late review :)

Asachoo:
Transitioned to utilizing a frequency multiplier to automatically adjust frequencies to the Hertz.

FranzForstmayr (approved):

jhillairet (approved):
Clean. Thank you.

PR #1195, opened 2024-11-03
Pass kind to interpolate method (again)

### LINKED ISSUE #1194: Interpolate method appears to return linear interpolation of network regardless of "kind" parameter
I have been using the network 'interpolation' method for over a year. When I most recently ran my code, the interpolation result appears to return a linear interpolation of the network S parameters regardless of what value is passed to the 'kind' parameter. 

Thank you for looking into this.

Erik

FranzForstmayr:
Closes #1194

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2905,6 +2905,7 @@ class Network:
             #Not supported by rational_interp
             is_rational = True
         else:
+            kwargs["kind"] = kind if kind is not None else "linear"
             f_interp = interp1d
 
         # interpret input
--- a/skrf/tests/test_network.py
+++ b/skrf/tests/test_network.py
@@ -1461,11 +1461,17 @@ class NetworkTestCase(unittest.TestCase):
         self.assertTrue( ((a+[1+1j,2+2j]).s == np.array([[[2+3j]],[[5+6j]]])).all())
 
 
-    def test_interpolate(self):
-        a = rf.N(f=[1,2],s=[1+2j, 3+4j],z0=1, f_unit="ghz")
-        freq = rf.F.from_f(np.linspace(1,2,4), unit='ghz')
-        b = a.interpolate(freq)
-        # TODO: numerically test for correct interpolation
+    def test_interpolate_linear(self):
+        net = rf.Network(f=[0, 1, 3, 4], s=[0,1,9,16], f_unit="Hz")
+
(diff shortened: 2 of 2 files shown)


jhillairet (approved):

jhillairet:
Once merged, I'll create a minor release integrating this fix...

FranzForstmayr:
Great, I think about an testcase to catch this issue, but will try to merge soon

jhillairet:
> Great, I think about an testcase to catch this issue, but will try to merge soon

the 0th order test could be testing the difference of the resulting network created using the various interpolation methods. They should be numerically different (not testing their performance). If they are all equals, there is a problem.

ejohnson-cfs:
Thank you. Has this been corrected?

I updated RF-scikit and reran my code, but I am not seeing any difference

jhillairet:
The patch is now included in the version 1.4.1

Please check if you are running this version.



Le mer. 6 nov. 2024, 19:50, ejohnson-cfs ***@***.***> a
écrit :

ejohnson-cfs:
thank you. I have confirmed that I have updated to 1.4.1. I however still am seeing that the interpolation order makes no difference. Attached is a comparision of how the function used to correctly interpolate a network with 5 data points to 21, creating a smooth interpolation. Now it only creates a linear interpolation between the 5 data points. 

correct cubic interpolation of data from 5 samples to 21
incorrect cubic interpolation of data from 5 samples to 21

FranzForstmayr:
@ejohnson-cfs can you show your code?

I just tried this:

```python
import skrf as rf

ntwk = rf.Network("skrf/tests/metas_tdr/short_10ps_10g_50g.s1p")
ntwk.plot_s_smith()

f2 = rf.Frequency(10e9,50e9, 20, unit="Hz")
print(f2.f)

ntwk2 = ntwk.interpolate(f2, kind="cubic")
ntwk2.name = "interp_cubic"
ntwk2.plot_s_smith()

ntwk3 = ntwk.interpolate(f2)
ntwk3.name = "interp_default"
ntwk3.plot_s_smith()

plt.savefig("smith.png")
```
smith

ejohnson-cfs:
Thank you Franz

Your test looks correct. 

When I ran it again on Friday, the interpolation is now working correctly. The patch appears to be correct. I am not sure why it did not work for me last Thursday and then did on Friday. I did not update the installation.

Thank you very much for your quick support on this.

Erik

PR #1202, opened 2024-11-12
Use RLGC distortion model and nominal impedance in DefinedAEpTandZ0 media

### LINKED ISSUE #1184: Difference in S-parameters of lossy line compared to AWR
I am having issues coming to terms with the difference in S-parameters I am getting between scikit's Media.line method and a transmission line formulated using the TLINP element in AWR.

The line's have the same S-parameters when there is no loss, however as soon as loss is introduced (using non-zero tan-delta), the results are different. I believe the same happens when loss is introduced through conductive losses. Although the differences are small for an individual line, they prove to ultimately have a catastrophically large effect in the overall S-parameters for a project I am working on.

I have attached a minimal code sample below, along with the AWR project, data and the graphs I generated using the code sample. I can generate more AWR data, if needed. Note that although the effect is most prominent with larger tan-delta (e.g. 0.5 as used below), the difference when tan-delta is even as small as 0.02 (as it is in my project) is still significant.

What could be causing this difference?

```
import skrf as rf
from skrf.media.definedAEpTandZ0 import DefinedAEpTandZ0

f = rf.Frequency(10, 500, 1001, 'MHz')
skrf_network = DefinedAEpTandZ0(frequency=f, z0=75, tanD=0.5, z0_port=50).line(1000, 'mm', name='skrf')
awr_network = rf.Network('awr_1000mm_75ohm_tanD0p50.s2p', name='awr')

skrf_network.plot_s_db(m=0, n=0)
awr_network.plot_s_db(m=0, n=0)
plt.savefig('S11.png')
plt.show()

skrf_network.plot_s_db(m=1, n=0)
awr_network.plot_s_db(m=1, n=0)
plt.savefig('S21.png')
plt.show()
```

S11

S21

[skrf_vs_awr.zip](https://github.com/user-attachments/files/17325604/skrf_vs_awr.zip)

mhuser:
Make  `DefinedAEpTandZ0` aware of the distortion caused by the losses.
Because of the distortion introduced by the conductor and dielectric losses, the characteristic impedance can be frequency-dependent with an imaginary part. The characteristic impedance is computed with an RLGC model as follow.

$$R = 2 Z_n \alpha_C$$
$$L = \frac{Z_n \sqrt{\epsilon_r}}{C_0}$$
$$G = \frac{2}{Z_n} \alpha_D$$
$$C = \frac{\sqrt{\epsilon_r}}{C_0 Z_n}$$
$$Z_0 = \sqrt{\frac{R + j\omega L}{G + j\omega C}}$$

Where $Z_n$ is the nominal impedance passes as `DefinedAEpTandZ0` `z0` parameter.
This behavior is consistent with AWR.
Fixes #1184


gvcallen (approved):
Changes look good to me

jhillairet (approved):

mhuser:
I just added a test case against awr tlinp .s2p to ensure the residuals keep small in case we do other changes in the future.

Ttl:
It looks like this correction is applied even when z0 is given as array but I wouldn't expect z0 to be modified when it's given as frequency dependent. I have some existing code that depends on this.

Distortion usually implies non-linear behaviour. Dispersion might be better term here.

mhuser:
Done in #1203

Issue #1205, opened 2024-11-15
Media.inductor produces nan's for L=0


gvcallen:
**Describe the bug**
When calling Media's "inductor" function with L=0, the resulting s-parameters are NaN due to divide-by-zero.

**To Reproduce**
```
import skrf as rf
from skrf.media import DefinedGammaZ0

frequency = rf.Frequency(10.0, 10.0, 1, 'MHz')
inductor = DefinedGammaZ0(frequency=frequency).inductor(0)
print(inductor.s)
```

**Expected behavior**
An equivalent "thru" set of S-parameters should be returned. This is consistent with the "capacitor" function, which correctly returns an open.

**Additional context**
I believe this is useful in that it takes some load off of the caller. In my case, I am running an optimization, and want the optimizer to decide whether or not an inductor should be present. Currently, I have to add an explicit check in my middleware code if L==0 and replace the circuit with a thru.

I tried implementing this in scikit-rf but ran into a few issues as to what the ideal implementation is, even though it is quite simple to do with some if conditions, however the main issues lie in both the fact that the caller can pass an array for L, and that you cannot exactly specify a "thru" in y-parameters, resulting in issues when trying to set only certain elements of result.y

Asachoo:
Similarly, should we consider expanding our scope to include components such as `Resistors`, `Capacitors`, and others?
If the input is detected to be zero (which can be effectively checked using `np.allclose()` method), should we directly return the theoretical equivalent of the component? For instance, as you mentioned a `0H Inductor()` would be represented as a 'thru'.

Asachoo:
Thanks, @mhuser. It's true that writing the S-Parameters directly gives more flexibility than converting through the Y-parameters. I'll working on providing explicit S-parameter expressions for both `resistor()` and `capacitor()`.

mhuser:
Unfortunately, with the `capacitor()`

$$Z = \frac{1}{j \omega C}$$

which would cause the division by 0 issue again!
There maybe the Y-parameter formulation is good to keep.

Ttl:
The whole S-parameter definition breaks down if port impedance is zero so there's no reason to worry about it.

gvcallen:
Thanks everyone for the very fast response on this.

PR #1206, opened 2024-11-15
Use s-parameter to build lumped components.

Asachoo:
Closes https://github.com/scikit-rf/scikit-rf/issues/1205


Asachoo:
To validate the `S-Parameter` of lumped components, I have added three extra unit tests to assess mismatch scenarios.

Ttl:
Does this handle complex port impedance correctly? The default `sdef` is `power` which usually causes issues with complex port impedance. It's often required to generate S-parameters with `traveling` definition and the uses `s2s` to change it to the default `power` definition.

It would be good to test against the current Y-parameter implementation. Test including complex port impedance would be also welcome.

mhuser (approved):
Looks good to me.

Asachoo:
> Does this handle complex port impedance correctly? The default `sdef` is `power` which usually causes issues with complex port impedance. It's often required to generate S-parameters with `traveling` definition and the uses `s2s` to change it to the default `power` definition.
> 
> It would be good to test against the current Y-parameter implementation. Test including complex port impedance would be also welcome.

Thanks for your advice @Ttl ! Regrettably, the complex port impedance cases not yielded correct s-parameters in my local tests. I am considering addressing this issue by directly developing accurate S-parameters functions directly, but I worry that this approach may become more complex.

Alternatively, instead of directly constructing the mismatched Component, we could create a matched Component, such as a 50 Ohm, and then use `renormalize()` to accommodate the mismatched port impedance. Perhaps this approach could be more robust and make the code more comprehensible.

mhuser:
Indeed `y2s` was applying different computations depending the `s_def` 🤔 
https://github.com/scikit-rf/scikit-rf/blob/3766c695dc46d8a9059e5ecad18714e50e2f0fb3/skrf/network.py#L7229-L7252

Ttl:
See for example how `line` in `media` is calculated for reference. You can also calculate with fixed real `z0` first and then `renormalize` to correct `z0` and `sdef`. However, you need to make sure that `sdef` of returned network is correct, see again `line`.

Asachoo:
Thanks ! @Ttl @mhuser 
I will reopen this PR once the `complex impedance` and `s_def` check are finalized.

Asachoo:
With the default setting (i.e., `s_def='power'`) for S-parameters now correctly handling complex port impedances.
For more test details, refer to the updated unit tests.

Ttl:
You have modified the equations to be correct for `s_def=power`, but I don't see how this can give correct result for any other `s_def`. You should have `renormalize` to `s_def = kwargs.pop('s_def', S_DEF_DEFAULT)` if other `s_def` is requested, just like in `line`. `result` network should have `s_def='power'` like in `line` with `pop` from `kwargs` before it, to make it easier to see what definition is used.

Ttl:
Please also add test against the old Y-parameter definition like in #1207. It could also test with different `s_def`.

Asachoo:
> You have modified the equations to be correct for `s_def=power`, but I don't see how this can give correct result for any other `s_def`. You should have `renormalize` to `s_def = kwargs.pop('s_def', S_DEF_DEFAULT)` if other `s_def` is requested, just like in `line`. `result` network should have `s_def='power'` like in `line` with `pop` from `kwargs` before it, to make it easier to see what definition is used.

The default `s_def` configuration (`s_def='power'`) and the `renormalize()` in necessary have been implemented.

> Please also add test against the old Y-parameter definition like in #1207. It could also test with different `s_def`.

The unit tests have been expanded to include checks for `real/imag Port impedance` renormalization, `s_def` renormalization, and comparisons with `Y-parameter references` to ensure accuracy and consistency.

Unexpectedly, two unexpected issues have arisen:
1. Under the condition of `s_def = 'pseudo'`, the S-parameter of the inductor shows an numerical error compared to the Y-parameter defined reference (with the decimal precision dropping from 6 to 5, but I think this is acceptable.).

2. There are unexpected and confusing unit test failures that require further investigation (But I didn't find out how this PR caused failures).

mhuser:
@Asachoo the unit test error with `skrf/io/tests/test_ts_spec.py` is strange and does not seem to be linked at all with `Media`.
Can you try to modify line 63 of this file with the following?
```Python
ex_5_6 = rf.Network(s=np.tile(s, [2, 1, 1]), z0=[50, 75, 0.01, 0.01], f=[5, 6], f_unit="GHz")
```

mhuser (approved):
The S-parameters are now built for power-waves and the various `s_def` are handled by `renormalize`.

Asachoo:
> @Asachoo the unit test error with `skrf/io/tests/test_ts_spec.py` is strange and does not seem to be linked at all with `Media`. Can you try to modify line 63 of this file with the following?
> 
> ```python
> ex_5_6 = rf.Network(s=np.tile(s, [2, 1, 1]), z0=[50, 75, 0.01, 0.01], f=[5, 6], f_unit="GHz")
> ```

My local test shows that Network Name and Frequency should not cause `AssertionError`
```python3
(scikit-rf) ➜  scikit-rf git:(FixIssue2) python   
Python 3.12.7 (main, Oct 16 2024, 07:12:08) [Clang 18.1.8 ] on darwin
>>> import skrf as rf
>>> import numpy as np
>>> rf.__file__
'/Users/asachoo/Git/scikit-rf/skrf/__init__.py'
>>> 
>>> s = np.random.rand(4,2,2)
>>> f1 = rf.Frequency(1,10,11,'ghz')
>>> f2 = rf.Frequency(1e9, 10e9,11,'hz')
>>> 
>>> # Different f unit
>>> ntwk1 = rf.Network(s=s, frequency=f1)
>>> ntwk2 = rf.Network(s=s, frequency=f2)
>>> ntwk1
2-Port Network: '',  1.0-10.0 GHz, 11 pts, z0=[50.+0.j 50.+0.j]
>>> ntwk2
2-Port Network: '',  1000000000.0-10000000000.0 Hz, 11 pts, z0=[50.+0.j 50.+0.j]
>>> ntwk1 == ntwk2
True
>>> 
>>> # Different Name
>>> ntwk1 = rf.Network(s=s, frequency=f1, name='withname')
>>> ntwk2 = rf.Network(s=s, frequency=f2)
>>> ntwk1
2-Port Network: 'withname',  1.0-10.0 GHz, 11 pts, z0=[50.+0.j 50.+0.j]
>>> ntwk2
2-Port Network: '',  1000000000.0-10000000000.0 Hz, 11 pts, z0=[50.+0.j 50.+0.j]
>>> ntwk1 == ntwk2
True
>>> 
```

mhuser:
> My local test shows that Network Name and Frequency should not cause `AssertionError`

Indeed. Frequency is compared in Hz by `__eq__` operator. Pytest does not fail on my machine.
`platform win32 -- Python 3.9.16, pytest-7.1.2, pluggy-1.0.0` 🤔

Asachoo:
Unable to pinpoint the cause of `test_ts_example_5_6()` failure 😔, all related fix attempts will be reverted.

Ttl:
I'm not exactly sure what's the issue with the test failure either, but this passes [tests](https://github.com/Ttl/scikit-rf/actions/runs/11875719076/job/33093247093) on my branch: https://github.com/Ttl/scikit-rf/commit/b4c1aece4868a1ca87f56b94bae1584b5e9e1cb7.

Asachoo:
Even the same commit (the latest 02f4f68) [passed actions](https://github.com/Asachoo/scikit-rf/actions/runs/11876295845) in my fork, but fails in the scikit repository, which is quite strange 🤔. Any suggestions @FranzForstmayr @jhillairet ?

jhillairet:
> Even the same commit (the latest [02f4f68](https://github.com/scikit-rf/scikit-rf/commit/02f4f686c4a36704493681ac64fd5eba4a048891)) [passed actions](https://github.com/Asachoo/scikit-rf/actions/runs/11876295845) in my fork, but fails in the scikit repository, which is quite strange 🤔. Any suggestions @FranzForstmayr @jhillairet ?

Very strange indeed... I also passes locally on my windows machine and also on github with a copy of your branch on [my repo](https://github.com/jhillairet/scikit-rf/actions/runs/11878287492/job/33098786177)...

Ttl:
It looks like it's an issue with Touchstone reading using uninitialized memory when loading "lower" or "upper" ts file. https://github.com/scikit-rf/scikit-rf/blob/master/skrf/io/touchstone.py#L536-L550. `np.empty` gives uninitialized array and it just happened to be zeroes previously. `np.nanmax` would then read the uninitialized array and error if entries weren't zero. For local testing replacing `L536`, `np.empty` with `np.ones` only breaks this test.

I'm not sure `np.nanmax` works correctly even when the data is initialized since max requires comparison of complex numbers and numpy complex comparison is lexical first real and then imaginary part. If input data has entries with negative real part it would fail here even if s was initialized to zero. It should be replaced with actual assignment of missing lower/upper triangular elements.

skrf/media/media.py
```diff
@@ -643,14 +643,27 @@ def resistor(self, R: NumberLike, *args, **kwargs) -> Network:
         capacitor
         inductor
         """
-        result = self.match(nports=2, **kwargs)
-        y = np.zeros(shape=result.s.shape, dtype=complex)
+        s_def = kwargs.pop('s_def', S_DEF_DEFAULT)
+        result = self.match(nports=2, s_def=S_DEF_DEFAULT, **kwargs)
```
Ttl, line 647: I doubt we will change `S_DEF_DEFAULT` but to make sure it works if we do, this should be `s_def='power'`. Same in other classes.

Ttl (changes requested):

skrf/media/media.py
Asachoo: All S-Parameters that are only accurate under the definition of 'power' have been changed to explicitly `s_def='power'`.

Ttl: `kwargs.pop` should still have `S_DEF_DEFAULT` to make sure the correct `s_def` is returned if `S_DEF_DEFAULT` is not power.

Asachoo: Modified as suggested.

jhillairet (approved):

PR #1207, opened 2024-11-15
line_floating S-parameter definition

Ttl:


Ttl:
For `s_def = traveling`: `s = (I - z*y*z) * Inverse(I + z*y*z)`, where `z = diag(sqrt(z0))` and `I` is identity matrix. Plugging in Y-parameters gives this result.

mhuser (approved):
Looks good to me and gives the same results as prior implementation for the test case.

jhillairet (approved):

PR #1208, opened 2024-11-17
Fix uninitialized memory use in touchstone parser

Ttl:
Uninitialized memory from `np.empty` was passed to `np.nanmax` when reading lower or upper touchstone 2 file. Should fix recent test failures.

### DIFF
--- a/skrf/io/touchstone.py
+++ b/skrf/io/touchstone.py
@@ -537,7 +537,7 @@ class Touchstone:
         if state.matrix_format == "full":
             self.s[:] = s_flat
         else:
-            index = np.tril_indices(state.rank) if state.matrix_format == "lower" else np.triu_indices(self.rank)
+            index = np.tril_indices(state.rank) if state.matrix_format == "lower" else np.triu_indices(state.rank)
             index_flat = np.ravel_multi_index(index, (state.rank, state.rank))
             self.s[:, index_flat] = s_flat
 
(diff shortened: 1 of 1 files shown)


jhillairet (approved):

mhuser (approved):
Looks great! This definitively remove the need to `np.nanmax` and will work also if the input real part is negative.

Asachoo:
Nice Work!

FranzForstmayr:
I still think the issue at #1206 was caused by something different, as I was able to pass the tests `skrf/io/tests/test_ts_spec.py::test_ts_example_5_6` in this [run](https://productionresultssa15.blob.core.windows.net/actions-results/e4939853-41e8-440d-b752-5f1a14d27699/workflow-job-run-c02e9c00-988f-505b-7cc6-d6642935a7a3/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-11-18T08%3A30%3A15Z&sig=9CtnPUK%2Fj0oV4k0xIKmB8wpB5hn%2Boid3d3XwEPRU1p0%3D&ske=2024-11-18T19%3A54%3A09Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-11-18T07%3A54%3A09Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-11-18T08%3A20%3A10Z&sv=2024-08-04) from this commit: https://github.com/scikit-rf/scikit-rf/pull/1206/commits/78699aa7492764a212c35bbba35dc6b6f0550524.

The log file is rather big, search for the test name, and the debug log from this change is shown:
https://github.com/scikit-rf/scikit-rf/pull/1206/commits/5ebd8caaf47df30677e53f5111adec82d4c59714

Basically all what I did was enabling the logs and bypass the parallel runs to actually see the logs.

Nevermind, the PR works as expected :)

Ttl:
It makes sense that the test would sometimes pass and sometimes fail if it relies on uninitialized memory being zero. In #1206 in your commit with debug logging S-parameters are clearly nonsense which also points to uninitialized memory: https://github.com/scikit-rf/scikit-rf/actions/runs/11880116457/job/33102867032#step:6:191 . master also failed the same test  before this PR: https://github.com/scikit-rf/scikit-rf/actions/runs/11880389392/job/33103479763. It can be just some package or operating system change in how memory was allocated and freed that made this fail now.

The code is clearly buggy the symptoms are what would be expected of it. I don't know what else could be the reason if not this.

Issue #594, opened 2022-01-04
`Network` constructor fails if `file` is a `StringIO` stream


mbastani:
The constructor of `Network` class allows the `file` argument to be either a string (path to the touchstone) or a file-like object (as the comment [here](https://github.com/scikit-rf/scikit-rf/blob/a1c7caec4f3219ac9f5050031b40b4608adf33ef/skrf/network.py#L419) states).

However, passing a file-like stream to the constructor raises an `AttributeError`, as the following MWE shows:

    import io
    import skrf as rf

    with open('path/to/touchstone/file', 'r')  as fh:
        contents = fh.read()

    stream = io.StringIO(contents)
    
    net = rf.Network(stream)  # <--- raises AttributeError
       

The issue is that `StringIO` does not have a `name` attribute. And it seems that knowing the file name is necessary for constructing the network.  (As the number of ports need to be determined from the file extension.)

FranzForstmayr:
Assign a name property on the stream as a quick workaround

```python
stream = io.StringIO(contents)
stream.name = 'simple-network.s2p'
net = rf.Network(stream) 
```

Not sure, if this is a bug or a feature :)

jhillairet:
It is a kind of feature let's say ;)

The documentation of Touchstone reading is under update in #587 but not merged yet. It will be:

```python
        Examples
        --------
        From filename
        >>> t = rf.Touchstone('network.s2p')
        
        File encoding can be specified to help parsing the special characters:
        
        >>> t = rf.Touchstone('network.s2p', encoding='ISO-8859-1')
        From file-object
        >>> file = open('network.s2p')
        >>> t = rf.Touchstone(file)
        
        From a io.StringIO object
        
        >>> link = 'https://raw.githubusercontent.com/scikit-rf/scikit-rf/master/examples/basic_touchstone_plotting/horn antenna.s1p'
        >>> r = requests.get(link)
        >>> stringio = io.StringIO(r.text)
        >>> stringio.name = 'horn.s1p'  # must be provided for the Touchstone parser
        >>> horn = rf.Touchstone(stringio)
```

mbastani:
I'm not sure if requiring the user to add a property to the (standard) `StringIO` is a good solution. 

[...]

PR #1212, opened 2024-11-19
Allow pass network name as kwarg for streams

FranzForstmayr:
Solves #594 finally.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -429,6 +429,8 @@ class Network:
 
             # allows user to pass StringIO, filename or file obj
             if isinstance(file, io.StringIO):
+                if not hasattr(file, "name") and name is not None:
+                    file.name = name
                 self.read_touchstone(file, self.encoding)
 
             else:
--- a/skrf/tests/test_network.py
+++ b/skrf/tests/test_network.py
@@ -329,12 +329,20 @@ class NetworkTestCase(unittest.TestCase):
             sio.name = os.path.basename(filename) # hack a bug to touchstone reader
             rf.Network(sio)
 
+    def test_constructor_from_stringio_hfss(self):
         filename = os.path.join(self.test_dir, 'hfss_oneport.s1p')
         with open(filename) as fid:
             data = fid.read()
             sio = io.StringIO(data)
             sio.name = os.path.basename(filename) # hack a bug to touchstone reader
             rf.Network(sio)
+
+    def test_constructor_from_stringio_name_kwawrg(self):
+        filename = os.path.join(self.test_dir, 'ntwk1.s2p')
+        with open(filename) as fid:
+            data = fid.read()
+            sio = io.StringIO(data)
+            rf.Network(sio, name=filename)
 
     def test_different_ext(self):
         filename= os.path.join(self.test_dir, 'ntwk1.s2p')


skrf/network.py
```diff
@@ -429,6 +429,8 @@ def __init__(self, file: str = None, name: str = None, params: dict = None,
 
             # allows user to pass StringIO, filename or file obj
             if isinstance(file, io.StringIO):
+                if not hasattr(file, "name") and name is not None:
+                    file.name = name
```
Asachoo, line 433: Is it essential to verify the format of the filename and ensure it conforms to the pattern `r"*.[ghsyz](\d+)p"`?

FranzForstmayr: The file name is verified inside the `Touchstone._parse_file` https://github.com/scikit-rf/scikit-rf/blob/master/skrf/io/touchstone.py#L375 function which is called afterwards, so here no verification necessary.

Asachoo (approved):

Asachoo:
I've noticed that the `name` property in the `Network` class can be a bit inconsistent. Let me illustrate my point with an example:
```python3
(scikit-rf) ➜  scikit-rf git:(master) ✗ ipython
Python 3.13.0 (main, Oct 16 2024, 08:05:40) [Clang 18.1.8 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.29.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import skrf as rf; from io import StringIO
matplotlib not found while setting up plotting

In [2]: ntwk1 = rf.Network('./skrf/tests/ntwk1.s2p'); ntwk1
Out[2]: 2-Port Network: 'ntwk1',  1.0-10.0 GHz, 91 pts, z0=[50.+0.j 50.+0.j]

In [3]: with open('./skrf/tests/ntwk1.s2p') as f:
   ...:     sio = StringIO(f.read())
   ...:     ntwk2 = rf.Network(sio, name='./skrf/tests/ntwk1.s2p')
   ...: ntwk2
Out[3]: 2-Port Network: './skrf/tests/ntwk1.s2p',  1.0-10.0 GHz, 91 pts, z0=[50.+0.j 50.+0.j]

In [4]: with open('./skrf/tests/ntwk1.s2p') as f:
   ...:     sio = StringIO(f.read())
   ...:     ntwk3 = rf.Network(sio, name='ntwk1.s2p')
   ...: ntwk3
Out[4]: 2-Port Network: 'ntwk1.s2p',  1.0-10.0 GHz, 91 pts, z0=[50.+0.j 50.+0.j]

In [5]: ntwk4 = rf.Network(s=ntwk1.s, frequency=ntwk1.frequency, name='ntwk4'); ntwk4
Out[5]: 2-Port Network: 'ntwk4',  1.0-10.0 GHz, 91 pts, z0=[50.+0.j 50.+0.j]

In [6]: ntwk5 = rf.Network('./skrf/tests/ntwk1.s2p', name='ntwk1.s2p'); ntwk5
Out[6]: 2-Port Network: 'ntwk1.s2p',  1.0-10.0 GHz, 91 pts, z0=[50.+0.j 50.+0.j]

In [7]: 
```

As you can see, when creating Network objects using "conventional" methods like `ntwk1` and `ntwk4`, the name property omit the `.s2p` extension.

However, for `Network` objects created with `StringIO`, such as `ntwk2` and `ntwk3`, still cling to the `extension` even when the path is redundant.

While there's no strict rule on the format of the Network's name, the sporadic inclusion of the extension can feel a bit inconsistent.

I understand that this is not a bug, but perhaps automatically removing the extension when parsing with `StringIO` would make the naming more consistent. Just brainstorming here, not advocating for a change :laughing:.

FranzForstmayr:
I see your point, and now I cannot unsee it anymore. :)

we need the proper extension for the parser, but we can remove it afterwards.

Ttl:
This has a side effect of adding `name` attribute to the user input. It likely isn't a big issue but it would be better to not modify the user input.

FranzForstmayr:
Name was already a user input, even before this PR. In this case the name is just used if the file argument is a stream.

FranzForstmayr:
Yes that's true, I also think no big deal. There's definitely no breaking change, as this code raised an exception before this change.

PR #1213, opened 2024-11-26
WIP: Try Jax as GPU accelerator

FranzForstmayr:
This is a first try to use Jax as GPU accelerator.

VectorFitting process is mainly depending on `qr` decomposition, so I just replaced `numpy.linalg.qr` with `jax.numpy.linalg.qr`.

I did some tests with the s4p file from [here](https://github.com/scikit-rf/scikit-rf/issues/553#issuecomment-957003759).
The matrix shape for `qr` composition starts at (16,2992,20) and ends at (16,2992,640) after adding some poles, so i benchmarked the last dimension from 20 to 2000.

```python
import numpy as np
import perfplot
import jax
jax.config.update("jax_enable_x64", True)
import jax.numpy as jnp

rng = np.random.default_rng()

qr_vmap = jax.vmap(lambda e: jnp.linalg.qr(e, mode='r'), in_axes=0)

perfplot.show(
    setup=lambda n: rng.random((16, 2992, n)),
    kernels=[
        lambda arr: np.linalg.qr(arr, mode="r"),
        lambda arr: np.array([np.linalg.qr(e, mode="r") for e in arr]),
        lambda arr: np.array([jnp.linalg.qr(e, mode="r") for e in arr]),
        lambda arr: np.array(qr_vmap(arr)),
    ],
    labels=["np vectorized", "np loop", "jax loop", "jax vectorized"],
    n_range=np.linspace(10,2000,20),
    xlabel="matrix shape(16,2992,x)",
)
```

Tests are done on Ubuntu 24.04, Ryzen 5 5600X, 32GB RAM, RTX3070 (8GB)

Result with jax[cpu]:
cpu_2000

Bascally no difference

Result with jax[cuda12]:
gpu_2000

Estimated ~30% ..  50% gain on GPU.

### DIFF
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -5,9 +5,10 @@ import os
 import warnings
 from timeit import default_timer as timer
 from typing import TYPE_CHECKING, Any
-import numpy as np
-import jax.numpy as jnp
+
 import jax
+import jax.numpy as jnp
+import numpy as np
 from scipy.integrate import trapezoid
 
 from .util import Axes, axes_kwarg
@@ -705,9 +706,9 @@ class VectorFitting:
 
         else:
             return None
-        
 
-    
+
+
     @staticmethod
     def _qr(arr: np.ndarray) -> np.ndarray:
         #return np.array([jnp.linalg.qr(e, mode='r') for e in arr])
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -5,8 +5,9 @@ import os
 import warnings
 from timeit import default_timer as timer
 from typing import TYPE_CHECKING, Any
-
 import numpy as np
+import jax.numpy as jnp
+import jax
 from scipy.integrate import trapezoid
 
 from .util import Axes, axes_kwarg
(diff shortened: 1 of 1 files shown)


Vinc0110:
This is interesting. I've never heard of JAX, but will definitely check it out.

Speaking of QR decomposition, I remembered the ugly workaround to old numpy. I've not yet compared the speed, but the loop looks like it might not be the most efficient option...
https://github.com/scikit-rf/scikit-rf/blob/ecaa7cbd54581a5e63979dd842eefb708432d241/skrf/vectorFitting.py#L804-L825

skrf/vectorFitting.py
```diff
@@ -806,10 +812,9 @@ def _pole_relocation(poles, freqs, freq_responses, weights_responses, fit_consta
 
         # direct QR of stacked matrices for linalg.qr() only works with numpy>=1.22.0
         # workaround for old numpy:
-        R = np.empty((n_responses, dim_k, n_cols_unused + n_cols_used))
+        #R = np.empty((n_responses, dim_k, n_cols_unused + n_cols_used))
         A_ri = np.hstack((A.real, A.imag))
-        for i in range(n_responses):
```
Vinc0110, line 811: I just realized you have already started to replace the loop. But can't you simply use the direct method (a few lines above), which is currently commented out?

PR #1219, opened 2024-12-16
extrapolate_to_dc: prevent phase aliasing between DC and first non-interpolated point

jankae:
First PR on this project, please let me know if I did anything wrong.

## The issue

[network.extrapolate_to_dc](https://github.com/scikit-rf/scikit-rf/blob/b4d423642880c2a98b0fcad6627a1c863f766453/skrf/network.py#L3027) does not work when there is a phase wrap between the extrapolated DC point and the first non-interpolated point.

### Example
Create a simple transmission line and a network starting at a high enough frequency:
```python
import skrf as rf

freq = rf.Frequency(0.7, 3.0, 201, 'GHz')
Z_0 = 50  # Ohm
Z_L = 50  # Ohm
d =0.5  # m
VF = 0.67
att = 0.02 # Np/m. Equivalent to 0.1737 dB/m

alpha = att  # Np/m !
beta = freq.w/rf.c/VF
gamma = alpha + 1j*beta

coax_line = rf.media.DefinedGammaZ0(frequency=freq, z0=Z_0, gamma=gamma)
ntw = coax_line.line(d, unit='m')
```
Plotting the phase of S21 looks as expected:
```python
ntw.s21.plot_s_deg(label='original')
```
image

Now we extrapolate to DC and plot that as well:
```python
extrapolated = ntw.extrapolate_to_dc(kind='linear')
extrapolated.s21.plot_s_deg(label='interpolated')
```
image

That is of course wrong. The phase should continue to rotate linearly with decreasing frequencies, wrap around twice more and then reach the DC point.

## Why does this happen?

No DC point was given to extrapolate_to_dc, so it has to extrapolate that by itself. This is done by taking the first two S parameter points and extrapolating them to DC: https://github.com/scikit-rf/scikit-rf/blob/b4d423642880c2a98b0fcad6627a1c863f766453/skrf/network.py#L3082-L3092

This results in a valid DC point. But we need additional points between DC and the first S parameter point at 700 MHz. This is done by this line at the end of the function: https://github.com/scikit-rf/scikit-rf/blob/b4d423642880c2a98b0fcad6627a1c863f766453/skrf/network.py#L3111

And this function has no concept of phase wrapping. It will try its best to interpolate between points but even for other kinds than `linear` interpolation, it can not handle the phase wrap.

## Proposed solution

When a DC point needs to be extrapolated, also extrapolate the other points from DC to the start frequency of the non-interpolated network with the same frequency spacing. Interpolating with `interpolate_self` is still necessary (if the first frequency is not a multiple of the frequency step) but now this function has equal spacing in frequency in that frequency range as well and does not miss phase wraps anymore:
image

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3090,11 +3090,11 @@ class Network:
             interp_rad = interp1d(f, rad, axis=0, fill_value='extrapolate')
             interp_mag = interp1d(f, mag, axis=0, fill_value='extrapolate')
             dc_sparam = interp_mag(0) * np.exp(1j * interp_rad(0))
-            
             # Extrapolate other points and insert
-            fstep = self.frequency.f[1] - self.frequency.f[0]
-            extrapolated_f = Frequency(fstep, result.frequency.f[0]-fstep, int(round(self.frequency.f[0]/fstep)))
-            for freq in extrapolated_f.f:
+            fstep = self.frequency.f[-1]/(points-1);
+            len_interp = points - len(self);
+            extrapolated_f = Frequency(fstep, (len_interp-1) * fstep, len_interp-1)
+            for freq in reversed(extrapolated_f.f):
                 interp_sparam = interp_mag(freq) * np.exp(1j * interp_rad(freq))
                 result.s = np.insert(result.s, 0, interp_sparam, axis=0)
                 result.frequency._f = np.insert(result.frequency.f, 0, freq)
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3090,6 +3090,15 @@ class Network:
             interp_rad = interp1d(f, rad, axis=0, fill_value='extrapolate')
             interp_mag = interp1d(f, mag, axis=0, fill_value='extrapolate')
             dc_sparam = interp_mag(0) * np.exp(1j * interp_rad(0))
+            
+            # Extrapolate other points and insert
+            fstep = self.frequency.f[1] - self.frequency.f[0]
+            extrapolated_f = Frequency(fstep, result.frequency.f[0]-fstep, int(round(self.frequency.f[0]/fstep)))
+            for freq in extrapolated_f.f:
+                interp_sparam = interp_mag(freq) * np.exp(1j * interp_rad(freq))
+                result.s = np.insert(result.s, 0, interp_sparam, axis=0)
+                result.frequency._f = np.insert(result.frequency.f, 0, freq)
+                result.z0 = np.insert(result.z0, 0, result.z0[0], axis=0)
         else:
             #Make numpy array if argument was list
             dc_sparam = np.array(dc_sparam)
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3091,14 +3091,15 @@ class Network:
             interp_mag = interp1d(f, mag, axis=0, fill_value='extrapolate')
             dc_sparam = interp_mag(0) * np.exp(1j * interp_rad(0))
             # Extrapolate other points and insert
-            fstep = self.frequency.f[-1]/(points-1);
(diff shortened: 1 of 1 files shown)


Ttl:
The first frequency point can be less than fstep and there might not be any points between DC and the first frequency point. This should be handled.

Another more fundamental issue is how this handles noise? Sometimes low frequency points might not be very accurate. Phase is extrapolated linearly which is very sensitive to error in the first two frequency points.

jankae:
> The first frequency point can be less than fstep and there might not be any points between DC and the first frequency point. This should be handled.

Valid points and I think I can handle these improvements without significant problems myself.

> Another more fundamental issue is how this handles noise? Sometimes low frequency points might not be very accurate. Phase is extrapolated linearly which is very sensitive to error in the first two frequency points.

Yeah, noise can likely be a problem. But it would be for the existing DC extrapolation as well, wouldn't it? This is certainly not perfect but I don't think it degrades performance when using automatic DC extrapolation.
At least not for linear interpolations. I am not sure really sure how this could potentially affect higher order interpolations in all use cases.

And just to manage expectations: I stumbled across this problem in a side project (used scikit-rf to check my own extrapolation code in another project). I am happy to improve this PR with small fixes (handle fstep edge cases, fix linter errors) but I am not really familiar enough with scikit-rf (or python) to check if it works for all possible input data.

Ttl:
I see, it's not used if the DC point is specified. It does introduce a difference on how the method works depending on if the DC point is given or not. I would expect the method to work the same whether the DC points is automatically extrapolated or if I specifically give the same linearly extrapolated DC point. I think it might make more sense as a parameter that modifies how the interpolation is done, although that does require some modifications for the case when the DC point is given?

You should run the tests locally first to check everything works: https://scikit-rf.readthedocs.io/en/latest/contributing/index.html#tests

jankae:
Thanks, I tested locally first now and at least these tests do not show any errors. I also added a tiny bit of logic that skips the additional interpolated points if the first frequency point is already low enough.

Other than that, I do not think that I will address your other concerns, sorry. I do not know enough about this project to understand all the dependencies and do not want to add additional interpolation parameters or anything without really knowing what I am doing.

Feel free to close this without merging. I can also convert this into an issue if you prefer. Maybe someone else might take on this problem with a more systematic approach. As it stands, this solution is good enough for my use case right now and I probably will not find enough time to implement this properly for everyone.

FranzForstmayr:
I'm not completely conviced yet, but have a few questions.

A few points from my side:
- In general there's a risk extrapolate that many values from a Network. In case of a noisy measurement, I'm not sure how well this can work. 
- Is the DC S param always completely real? I'm not sure if this is enforced currently, but did not check tbh.
-  Did you check with S11/S22 values as well?

jankae:
> In general there's a risk extrapolate that many values from a Network. In case of a noisy measurement, I'm not sure how well this can work.

Yes, ideally we do not want to extrapolate. But if there are only measurements above a certain frequency and they are required  all the way to DC for further calculations, it has to happen. This PR only attempts to improve the extrapolation in certain cases. The same amount of values were already extrapolated before.

> Is the DC S param always completely real? I'm not sure if this is enforced currently, but did not check tbh.

It should be, I can not imagine any network for which it would not be real. The imaginary part of the DC value is forced to zero at the end of extrapolate_to_dc. This PR does not change that.

> Did you check with S11/S22 values as well?

No. I can not think of any reason why the behavior would be different (the issue is phase wrapping and that will happen for all S parameters), but I did not check.

FranzForstmayr:
> It should be, I can not imagine any network for which it would not be real.

Yes in reality it's always completely real, but i wasn't sure if the calculation enforced it too.

PR #1228, opened 2024-12-31
VectorFitting: New SPICE netlist topology

Vinc0110:
A different circuit topology for the equivalent netlists of vector-fitted scattering responses enables massive improvements in the simulation runtime. This is at least true for circuit simulators using modified nodal analysis, such as ngspice and Xyce. See discussion https://github.com/scikit-rf/scikit-rf/discussions/1225.

I tested three different pole configurations with two different networks and compared the simulation time in _ngspice_ with the current admittance-based topology (master branch) and with the new impedance-based topology (this PR).

Both approaches provide accurate results, at least for SP simulations in ngspice, but differences in runtime are huge!

Example: `scikit-rf/doc/source/examples/vectorfitting/190ghz_tx_measured.S2P`
- Vector fit with model order 15
- Simulation time with admittance-based topology (master branch): **0.051 s**
- Simulation time with impedance-based topology (this PR): **0.028 s**

Example: `4port.s4p` (reduced fit)
- Vector fit with model order 105
- Simulation time with admittance-based topology (master branch): **257.759 s**
- Simulation time with impedance-based topology (this PR): **0.179 s**

Example: `4port.s4p` (full fit)
- Vector fit with model order 317
- Simulation time with admittance-based topology (master branch): **>> 15 minutes** (cancelled because of impatience)
- Simulation time with impedance-based topology (this PR): **0.819 s**

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -2,14 +2,10 @@ import os
 import sys
 import tempfile
 import unittest
-from contextlib import suppress
 from pathlib import Path
 
 import numpy as np
 import pytest
-
-with suppress(ImportError):
-    from PySpice.Spice.Parser import SpiceParser
 
 import skrf
 
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -77,32 +77,6 @@ class VectorFittingTestCase(unittest.TestCase):
         # quality of the fit is not important in this test; it only needs to finish
         self.assertLess(vf.get_rms_error(), 0.2)
 
-    @pytest.mark.skipif(
-        "PySpice" not in sys.modules,
-        reason="test_spice_subcircuit uses PySpice parser but it is not available.")
-    def test_spice_subcircuit(self):
-        # fit ring slot example network
-        nw = skrf.data.ring_slot
-        vf = skrf.vectorFitting.VectorFitting(nw)
-        vf.vector_fit(n_poles_real=4, n_poles_cmplx=0, fit_constant=True, fit_proportional=True)
-
-        # write equivalent SPICE subcircuit to tmp file
-        tmp_file = tempfile.NamedTemporaryFile(suffix='.sp', delete=False)
-        name = tmp_file.name
-        tmp_file.close()
-        vf.write_spice_subcircuit_s(name)
-
-        parser = SpiceParser(name)
-
-        # Number of elements on global level
-        assert len(parser.subcircuits[0]._statements) == 38
-        # Number of elements in RLCG subckt
-        assert len(parser.subcircuits[1]._statements) == 4
-        # Number of elements in RL subckt
-        assert len(parser.subcircuits[2]._statements) == 2
-
-        os.remove(name)
-
     def test_read_write_npz(self):
         # fit ring slot example network
         nw = skrf.data.ring_slot
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -2447,7 +2447,7 @@ class VectorFitting:
 
             f.write(f'.ENDS {fitted_model_name}\n\n')
 
-            # Subcircuit for an LCRR equivalent impedance of a complex-conjugate pole-residue pair
+            # Subcircuit for an RCL equivalent impedance of a complex-conjugate pole-residue pair
             f.write('.SUBCKT rcl_active n_pos n_neg cap=1e-9 ind=100e-12 res1=1e3 res2=1e3 gt1=2e-3 gt2=2e-3\n')
             f.write('L1 n_pos 1 {ind}\n')
             f.write('R1 1 n_neg {res1}\n')
@@ -2458,7 +2458,7 @@ class VectorFitting:
             f.write('.ENDS rcl_active\n\n')
 
             # Subcircuit for an RC equivalent impedance of a real pole-residue pair
-            f.write('.SUBCKT rc n_pos n_neg res=1e3 cap=1e-9\n')
+            f.write('.SUBCKT rc_passive n_pos n_neg res=1e3 cap=1e-9\n')
             f.write('C1 n_pos n_neg {cap}\n')
             f.write('R1 n_pos n_neg {res}\n')
-            f.write('.ENDS rc\n\n')
+            f.write('.ENDS rc_passive\n\n')
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -2321,7 +2321,7 @@ class VectorFitting:
             # write title line
             f.write('* EQUIVALENT CIRCUIT FOR VECTOR FITTED S-MATRIX\n')
             f.write('* Created using scikit-rf vectorFitting.py\n')
-            f.write('*')
+            f.write('*\n')
 
             # Create subcircuit pin string and reference nodes
             if create_reference_pins:
@@ -2332,7 +2332,8 @@ class VectorFitting:
             f.write(f'.SUBCKT {fitted_model_name} {str_input_nodes}\n')
 
             for i in range(self.network.nports):
-                f.write(f'\n* Port network for port {i + 1}\n')
+                f.write('*\n')
+                f.write(f'* Port network for port {i + 1}\n')
 
                 if create_reference_pins:
                     ref_node = f'p{i + 1}_ref'
@@ -2351,8 +2352,8 @@ class VectorFitting:
 
                 # VCCS and CCCS adding their currents to represent the incident wave a_i
                 # I_a_i = U_i / 2 / sqrt(Z0_i) + sqrt(Z0_i) / 2 * I_i
-                f.write(f'Ga{i + 1} 0 n_0_{i + 1}_0 p{i + 1} {ref_node} {gain_vccs_a_i}\n')
-                f.write(f'Fa{i + 1} 0 n_0_{i + 1}_0 V{i + 1} {gain_cccs_a_i}\n')
+                f.write(f'Ga{i + 1} 0 n_1_{i + 1}_0 p{i + 1} {ref_node} {gain_vccs_a_i}\n')
+                f.write(f'Fa{i + 1} 0 n_1_{i + 1}_0 V{i + 1} {gain_cccs_a_i}\n')
 
                 for j in range(self.network.nports):
                     # transfer impedances connected in series to current sources representing a_i
@@ -2361,6 +2362,7 @@ class VectorFitting:
                     # I_a_i ~ a_i
                     # Z_j_i_k ~ S_j_i_k
                     # U_j_i_k ~ b_j_k
+                    f.write('*\n')
                     f.write(f'* Transfer from port {i + 1} to port {j + 1}\n')
 
                     z0_j = np.real(self.network.z0[0, j])
@@ -2368,7 +2370,8 @@ class VectorFitting:
 
                     # Stacking order in VectorFitting class variables:
                     # s11, s12, s13, ..., s21, s22, s23, ...
-                    i_response = i * self.network.nports + j
+                    # `i_response` is index for S_j_i
+                    i_response = j * self.network.nports + i
 
                     # Start with proportional and constant term of the model
                     # H(s) = d + s * e  model
@@ -2381,7 +2384,7 @@ class VectorFitting:
                     if n_nodes_remaining == 0:
                         # should not occur; let's still catch it
                         # write useless resistor to GND to simulation won't crash
-                        f.write(f'Rfail{i + 1}_{j + 1} n_{i + 1}_{j + 1}_0 0 100\n')
+                        f.write(f'Rfail{j + 1}_{i + 1} n_{j + 1}_{i + 1}_0 0 100\n')
                         break
                     if n_nodes_remaining == 1:
                         n_next = 0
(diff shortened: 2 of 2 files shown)


Vinc0110:
circuit_scattering_impedance

Vinc0110:
circuit_equivalent_impedance_cc

Vinc0110:
Here are some detailed comparisons for my second example with the reduced 4-port (incomplete fit):

Inventory and runtime analysis for admittance topology (master branch):
```
Circuit Inventory

Capacitor: 832
CCVS: 8
Inductor: 848
Resistor: 868
VCCS: 832
VCVS: 8
Vsource: 8
```
Runtime analysis with `rusage`:
```
Total elapsed time (seconds) = 257.759 
Total DRAM available = 15673.430 MB.
DRAM currently available = 9412.820 MB.
Maximum ngspice program size =   99.605 MB.
Current ngspice program size =   84.672 MB.

Shared ngspice pages =   11.344 MB.
Text (code) pages =    5.992 MB.
Stack = 0 bytes.
Library pages =   76.031 MB.

Number of lines in the deck = 9345
Netlist loading time = 0.007
Subckt and Param expansion time = 0.027
Netlist parsing time = 0.006

Nominal temperature = 27
Operating temperature = 27
Total iterations = 3
Transient iterations = 0
Circuit Equations = 2581
Circuit original non-zeroes = 11920
Circuit fill-in non-zeroes = 695159
Circuit total non-zeroes = 0
Transient timepoints = 0
Accepted timepoints = 0
Rejected timepoints = 0
Total analysis time (seconds) = 257.714
Matrix load time = 1.134
Matrix synchronize time = 0
Matrix reorder time = 50.751
Matrix factor time = 199.275
Matrix solve time = 6.507
Transient analysis time = 0
Transient load time = 0
Transient sync time = 0
Transient factor time = 0
Transient solve time = 0
Transient trunc time = 0
Transient iterations for the last time point = 3
AC analysis time = 205.984
AC load time = 1.112
AC sync time = 0
AC factor time = 196.746
AC solve time = 6.476
```

Inventory and runtime analysis for impedane topology (this PR):
```
Circuit Inventory

Capacitor: 848
CCCS: 4
Inductor: 832
Resistor: 1700
VCCS: 2532
Vsource: 8
```
Runtime analysis with `rusage`:
```
Total elapsed time (seconds) = 0.179 
Total DRAM available = 15673.430 MB.
DRAM currently available = 9427.379 MB.
Maximum ngspice program size =   74.180 MB.
Current ngspice program size =   59.156 MB.

Shared ngspice pages =   11.207 MB.
Text (code) pages =    5.992 MB.
Stack = 0 bytes.
Library pages =   50.605 MB.

Number of lines in the deck = 13544
Netlist loading time = 0.011
Subckt and Param expansion time = 0.038
Netlist parsing time = 0.011

Nominal temperature = 27
Operating temperature = 27
Total iterations = 3
Transient iterations = 0
Circuit Equations = 2549
Circuit original non-zeroes = 10160
Circuit fill-in non-zeroes = 4070
Circuit total non-zeroes = 0
Transient timepoints = 0
Accepted timepoints = 0
Rejected timepoints = 0
Total analysis time (seconds) = 0.119
Matrix load time = 0.023
Matrix synchronize time = 0
Matrix reorder time = 0.01
Matrix factor time = 0.028
Matrix solve time = 0.026
Transient analysis time = 0
Transient load time = 0
Transient sync time = 0
Transient factor time = 0
Transient solve time = 0
Transient trunc time = 0
Transient iterations for the last time point = 3
AC analysis time = 0.105
AC load time = 0.021
AC sync time = 0
AC factor time = 0.028
AC solve time = 0.026
```

Vinc0110:
The fully fitted 4-port has a huge inventory, but it's still fast with the impedance topology:

```
Circuit Inventory

Capacitor: 2544
CCCS: 4
Inductor: 2528
Resistor: 5092
VCCS: 7620
Vsource: 8
```

```
Total elapsed time (seconds) = 0.819 
Total DRAM available = 15673.430 MB.
DRAM currently available = 9823.023 MB.
Maximum ngspice program size =  169.449 MB.
Current ngspice program size =  149.656 MB.

Shared ngspice pages =   11.051 MB.
Text (code) pages =    5.992 MB.
Stack = 0 bytes.
Library pages =  145.875 MB.

Number of lines in the deck = 40680
Netlist loading time = 0.028
Subckt and Param expansion time = 0.109
Netlist parsing time = 0.032

Nominal temperature = 27
Operating temperature = 27
Total iterations = 3
Transient iterations = 0
Circuit Equations = 7637
Circuit original non-zeroes = 30512
Circuit fill-in non-zeroes = 16937
Circuit total non-zeroes = 0
Transient timepoints = 0
Accepted timepoints = 0
Rejected timepoints = 0
Total analysis time (seconds) = 0.656
Matrix load time = 0.211
Matrix synchronize time = 0
Matrix reorder time = 0.165
Matrix factor time = 0.093
Matrix solve time = 0.092
Transient analysis time = 0
Transient load time = 0
Transient sync time = 0
Transient factor time = 0
Transient solve time = 0
Transient trunc time = 0
Transient iterations for the last time point = 3
AC analysis time = 0.476
AC load time = 0.205
AC sync time = 0
AC factor time = 0.092
AC solve time = 0.091
```

Vinc0110:
Another benchmark with a funky 8-port from the internet. Probably a measurement of a backplane.

Vector fitting results from scikit-rf:
Backplane_fitted

SP simulation results from ngspice
image

Inventory and rusage from ngspice:
```
Circuit Inventory

Capacitor: 11328
CCCS: 8
Inductor: 11264
Resistor: 22664
VCCS: 33928
Vsource: 16
```

```
Total elapsed time (seconds) = 98.120 
Total DRAM available = 15673.430 MB.
DRAM currently available = 7191.707 MB.
Maximum ngspice program size =  665.574 MB.
Current ngspice program size =  636.121 MB.

Shared ngspice pages =   11.422 MB.
Text (code) pages =    5.992 MB.
Stack = 0 bytes.
Library pages =  642.000 MB.

Number of lines in the deck = 181064
Netlist loading time = 0.106
Subckt and Param expansion time = 0.61
Netlist parsing time = 0.201

Nominal temperature = 27
Operating temperature = 27
Total iterations = 3
Transient iterations = 0
Circuit Equations = 33961
Circuit original non-zeroes = 135808
Circuit fill-in non-zeroes = 109457
Circuit total non-zeroes = 0
Transient timepoints = 0
Accepted timepoints = 0
Rejected timepoints = 0
Total analysis time (seconds) = 97.398
Matrix load time = 7.005
Matrix synchronize time = 0
Matrix reorder time = 4.95
Matrix factor time = 5.323
Matrix solve time = 8.892
Transient analysis time = 0
Transient load time = 0
Transient sync time = 0
Transient factor time = 0
Transient solve time = 0
Transient trunc time = 0
Transient iterations for the last time point = 3
AC analysis time = 92.339
AC load time = 6.956
AC sync time = 0
AC factor time = 5.305
AC solve time = 8.888
```

fnordpole:
Nice work, looks very good!! I didn't check out all the details yet but here are my comments so far:

It seems we have two separate changes here: 
1. You derived a new equivalent series impedance that seems to have the same transfer function than the RLC+CCVS equivalent impedance from the Antonini paper but without the CCVS. Instead we have now two RCL+R+VCCS+VCCS.
It looks a bit complicated to me on the first glance but if it simulats faster than the RLC+CCVS, why not. Maybe we can find something even simpler in the future.

2. The topology of the current approach is changed from admittance-based to impedance-based.
I think in this conversion one of the good properties of the admittance-based solution got lost, which is that we don't a separate transfer-source for every positive or negative Y because we had the positive-a and negative-a incident wave sources and then just summed up all positive and negative currents to get the reflected wave b.

If I directly "convert" this approach to impedance-based, I would expect something like in the picture below.
Which is, we have now a positive-a and negative-a current source for the incident wave. This we can achieve with just one single current source because the current that comes out on the positive side is exactly in the opposite direction than the current that comes out on the negative side.

Now we have two strings of transfer-impedances: All the positive transfer impedances on the positive current side and all the negative transfer impedances on the negative side.

With this topology we again benefit from the summation of all the positive impedances in terms of voltage Vtp (voltage, transfer, positive) at two nodes in the positive chain and Vtn (voltage, transfer, negative) in the negative chain. The absolute voltages are very different in the two chains because the two chains for negative and positive impedances will be different. So we can't just sum all the positive and negative voltages easily but still:

For the port network we now only need two G sources for the bji waves: One for the positive contributions of port j to port i and one for the negative contributions of port j to port i.

In the picture I drew the exact conversion of the current approach to impedance-based, including the RLC-CCVS approach from Antonini instead of the admittance from Antonini that we have right now in the admittance-based approach but I guess it should be plug-and-play replaceable with your new RLC-R-VCCS+VCCS realization. Maybe we can have both and compare them. 

Sorry for the bad picture quality, I thought it's better to reply early instead of waiting until I drew it properly

signal-2025-01-01-191209

Vinc0110:
I like your ideas. My solution with the many individual voltage transfers for every single impedance is based on laziness and on the (questionable) assumption that VCCS don't cause much additional computation time, so we can be generous with them. Your proposal with the summations of positive and negative voltages and a transfer with only 2 VCCS might be even faster. Hard to tell...

I guess we will have to try and compare both methods. Not sure if I will have enough time to do this. I already neglected quite a few things for this PR in the last few days and should not stretch it too much :laughing:

Vinc0110:
Regarding the topology for the equivalent impedance networks: it would be very easy to replace them for a quick test, of course.

Vinc0110:
Thinking about it some more, I don't believe there would be much speedup if we reduce the number of VCCS for the transfer from the voltages across the impedances to the port current.

If you look at the linear system of equations for nodal analysis, the matrix size only depends on the number of nodes in the circuit.

Screenshot from http://emlab.illinois.edu/ece546/Lect_16.pdf, page 4. Here, the network as 2 nodes, whose voltages need to be solved for:
image

In modified nodal analysis, (certain) branch currents are added to the solution vector, increasing the number of unknowns and the size of the matrix. But basically, it's the same story, we want as few nodes as possible. Also, we want to avoid those extra currents in the solution vector, which would be required to current-controlled sources and voltage sources in general.

This means that if a node is already present in a circuit, the solver will have to calculate it's node voltage anyways and it does not hurt to connect additional branches from other existing nodes to it. It only causes the respective element stamp of that branch to be added to the existing matrix entries. In case of a VCCS, that stamp is small and does not add any new unknows to the linear system.

At the same time, it can get even more messy in the netlist synthesizer if we need to keep track of the nodes for positive and negative impedance branches. Feel free to give it a try. Maybe you find an elegant method that does not result in messy code, but I'm not sure it would be worth it...

Vinc0110:
Speaking of nodes: 
It would make a huge difference if there was an equivalent impedance network for a complex-conjugate pole-residue pair with only **two nodes** (input and output) instead of the **three nodes** currently required in my network. That reduction would really reduce the matrix size in cases with a large number of poles in the model.
But I guess a three-node network is already pretty small and I'm not sure a truly parallel 2-node circuit exists. Any ideas? :rocket:

Vinc0110:
@FranzForstmayr I hope you don't mind that I removed your testcase for the SPICE netlist export. It obviously does not work anymore with the new method. I could change the values, but I don't really see the point. In my opinion, we should simply accept that this code cannot be tested without running the netlist in SPICE.

FranzForstmayr:
I'm perfectly fine with this.

Another point from my side... Could someone check if the Network is compatible with LTspice? It's used very regular, and previously there was an issue with one certain component, wich behaved different to ngspice.
I'm on vacation currently and cannot check.

Vinc0110:
Good point. LTspice seems to work just fine with both models (admittance-based and impedance-based). I did not check the results for correctness, because it takes some effort to calculate the S parameters with equations. I just looked at the port voltages and they look alright.

Interestingly, the performance is much different from ngspice and Xyce. LTspice is faster with both models and there is not much of a difference between the models. They must be handling the linear system differently. Maybe a different solver?

**Example 1:** 4-port with reduced order (model order 105):
- AC simulation with 500 frequency points
- Admittance based: Total elapsed time: 0.289 seconds.
- Impedance based: Total elapsed time: 0.307 seconds.

**Example 2:** 8-port with model order 353
- AC simulation with 500 frequency points
- Admittance based: Total elapsed time: 8.208 seconds.
- Impedance based: Total elapsed time: 7.929 seconds.

fnordpole:
I implemented my suggestions above on my branch dev-ts and tested it thoroughly in transient and ac, including multiple instances of the subckt in parallel to increase the overall component count.

Turns out that it produces about -20 % less component count, but it runs 20 % slower. Bummer.
(Transient runs the same speed, only linear is different, but still.)

So I would say 1 G source for transfer per transfer impedance is definitely the fastest in Xyce.

Just as an experiment but still interesting:
In the admittance based variant I tried replacing the port-network CCVS with CCCS and it reduced the linear solve time in Xyce by 50 % even though it were only 4 sources but still it's way slower than the impedance based version. Those controlled voltage sources seem to be really bad for linear solver speed.

I would strongly suggest to merge in this pull requests version of the synthesizer.

fnordpole:
In Xyce it's in the reference manual pg. 200
"Resistors can have either positive or negative resistance values (R). A zero resistance value (R) is also allowed."

In ngspice I found this reply from one of the developers:
https://sourceforge.net/p/ngspice/discussion/120973/thread/d3e184cc/

"There is no minimum, any positive or negative value is allowed, EXCEPT 0. This is because NGSPICE calculates g = 1/R and puts that in the circuit matrix."

For LTspice the information is a bit thin because they have no reference manual and on the website I can only find their forum where someone reported that they used negative resistors in ltspice:
https://groups.io/g/LTspice/topic/negative_resistance/50142730
I didn't try it on my own. 

I guess the ngspice manual advises against using them because in "normal" circuits you will almost always get convergence problems if you put a negative resistor in your circuit, but for a sythesized netlist it should be ok IMHO.

As it also doesn't cost a lot to leave them in, I would leave them in. And we are definitely on the safe side by keeping the extra G sources and positive R values only.

PR #1231, opened 2025-01-05
New features in vectorFitting: Weighting, Multi-SISO and MIMO fitting

fnordpole:
This pull request introduces three new features to the vectorFitting class:

1. Frequency dependent weighting: Every single sample of every response can be weighted in the least squares fit thus, trading off the accuracy between big and small magnitudes. I created an example how to use it in doc/source/examples/vectorfitting/vectorfitting_ex5_weighting.ipynb
2. Multi-SISO-Fitting: The vectorFitting class is now prepared to support different number of poles for one or more responses. The class data structures 'poles', 'residues', 'constant' and 'proportional', that have previously been numpy arrays can now also be python-lists containing numpy arrays. The entire class and all methods support those four data structures to be lists or numpy arrays. The parameter pole_sharing can now be passed to vector_fit and auto_fit. The default is pole_sharing='MIMO', which means the same behavior as before (poles are shared between responses). It is also possible to set pole_sharing='Multi-SISO', which will use a separate set of poles for every response. I am working on also implementing 'Multi-SIMO', which will use a separate set of poles for every input, but common to all outputs. This will come in a future pull request, however. I created an example how to use it in doc/source/examples/vectorfitting/doc/source/examples/vectorfitting/vectorfitting_ex6_shared_poles.ipynb
3. Automatic initial poles estimation from data: Previously, both auto_fit and vector_fit had a fixed number of initial poles, distributed along the frequency axis, which had to be provided by the user. An automatic placement of initial poles is now implemented in a new method called _get_initial_poles(). It scans the responses for the amount of significant ripples and adds complex conjugate pole pairs for each. If the responses have no ripples but are totally smooth, it adds real poles. These choices are suggested in the original Gustavsen vector fitting paper and seem to work well for all the example sNp files in scikit-rf and some more that I tested. 
4.  Memory saver: When fitting larger networks with > 70 ports, I encountered memory problems and thus introduced serialization of the solving of the A_dense matrix in _relocate_poles. This can be enabled via the new argument memory_saver to auto_fit and vector_fit. If memory_saver=False, the system matrix is built for all responses and then solved by numpy. If memory_saver=True, the system matrix is built only for each response, then solved by numby and so on, collecting the way smaller result matrix R22.
5. Model-Summary-Print: After auto_fit and vector_fit, a short summary of the model that contains the number of poles, real poles, complex poles, absolute and relative erros, and model passivity, etc. is printed.
6.  Spurious pole detection improvement: The code in get_spurious is based on an integration of the transfer function L2-Norm of every complex conjugated pole-residue pair. This integral was based on a fixed-spacing-sampling with 101 samples and I observed cases in which this can fail because the responses can be very sharp and might not be covered accurately by those fixed spaced samples. I added a proper intergral using scipy.integrate.quad() which has a more sophisticated numerical integration that is more robust against this. It can happen though, if that it shows a user warning if the responses to be integrated are extremely sharp. The warning is about reduced accuracy of the integral in these cases, which is good, because it indicates potential problems. Still, it will not fail, just warn and try to integrate as good as possible. Also, I changed the spurious pole detection to only consider complex conjugate pole pairs instead of all poles. This is suggested in the corresponding VF adding and skimming (VF-AS) paper and saves a bit of computing time also.
7.  Pole collision avoidance algorithm:  In the VF-AS paper, it is described that new added poles need to keep a minimum distance to already existing poles. If they collide, the new poles will be moved to the best possible position without a collision. This is implemented differently in the previous vectorFitting class. It apparently still works but I nevertheless wrote a new method _add_poles() that implements exactly the procedure that is described in the mentioned paper.
8. Explicit computation of pole candidates for adding after skimming: In the current vectorFitting class, this is implemented in _find_error_bands and in auto_fit. I had a hard time to understand the error band finding code in _find_error_bands because it is pretty dense and has a lot of complexity to make it work in numpy. But actually this code is not "hot" at all, because it operates only on n_freqs samples (200-1000 maybe), so I just wrote it out in plain python, thus making it readable and easy to understand for new readers and also debuggable because you can see what's going on. I moved it into _get_pole_candidates(). Thus keeping error-band related stuff inside of this method, making auto_fit() more concise because it just calls this and gets the new pole candidates.
9. Important fix in the calculation of the RMS error: In get_rms_error(), for every response the RMS errors are summed up to calculate the final RMS error. But we actually need to sum up all mean-squared errors of all responses and then take the square root to get the total RMS error. I corrected this and this will affect all reported RMS errors. They will not be comparable with the RMS errors reported by the current implementation.
10. Initial poles placement and type: The current vectorFitting class has parameters n_poles_init_real and n_poles_init_complex. They are both independently distributed along the frequency axis and this can lead to problems if they are both the same value because then, a real pole and a complex conjugated pole pair will be placed exactly on the same frequency which will lead to an ill conditioned system matrix. In the VF-AS paper it is also warned to avoid this by all means. We could fix that by adding a messy code that avoids collisions moving the poles apart from each other if this problem occurs. But I decided to fix it differently: You can now provide n_poles_init and poles_init_type to auto_fit and vector_fit instead. This is also according to the recommendation of the original Gustavsen paper that states that for smooth responses, real poles should be used only and for non-smooth responses, complex-conjugate pole pairs should be used as initial poles. Specifying a mixture of both is not recommended in the literature even though it could of course, be done. But then we have to find a way to avoid collisions and need to specify where the real poles go and where the complex poles go, making it more complicated. In practical applications, specifying all initial poles as real or all poles as complex conjugate pole pairs is also not a problem because they will be changed one into the other by the vector fitting algorithm in any case, if necessary.
11. Argument and variable naming: I changed the parameter "alpha" and "gamma" in auto_fit to "error_stagnation_threshold" and "spurious_pole_threshold". It's just confusing to forget what gamma and alpha represent and you have to look it up every time. With the new names it's clear instantly without need to look up the documentation.
12. Argument error_target in auto_fit was renamed to "abstol". Abstol is the standard name for the absolute error below which we don't care. It's not a target, but rather a tolerance that represents how much absolute error we are willing to tolerate. 
13. Vector-Fitting pole relocation loop: in vector_fit(), the pole relocation loop used various flags like "stop" and "converged" and "n_iterations" to determine if the loop should stop or not. I had a really hard time to understand what is going on with all of those variables and the many logical combinations of them checked by all kinds of conditionals. I think the whole process can be boiled down to two clean stopping criterions: delta-relative-change-of-singular-values < threshold and n_iteration < max_iterations. If we run into max_iterations, we print the convergence hints. Otherwise we consider it converged. I changed the stopping conditions like this and it seems to work just like in the current version but it is way easier to read and understand. If I missed something maybe we can also change it back to the current code.  I also added a third stopping condition: maximum_error < abstol. This is actually the same stopping condition as used in auto_fit. If we reached the abstol, we can stop. I do it only every 25 iterations because it involves creating the entire model which in turn involves a residue fit. So we maybe don't do it every single iteration.
14. auto_fit and vector_fit are now "high-level" functions dealing with the SISO/MIMO/etc stuff and the actual code is in _auto_fit() and _vector_fit(), which are called by the high level functions multiple times to achieve the siso/mimo/etc. sub-fits.
15. Here and there, I changed a variable name in the implementation to match the corresponding paper's notation. It makes it easier for future contributors to match the code against the literature I hope. I don't remember all of them but they are obvious and hopefully don't add much confusion for those familiar with the current code.
16. Relative Errors: In order to assess the weighted fits, also the relative errors become important. So I added new methods get_total_abs_err, get_total_rel_err, get_abs_err_vs_responses, get_rel_err_vs_responses, get_abs_err and get_rel_err. The existing get_rms_error() reports the get_total_abs_err() value. (but including the bugfix mentioned above)

That was a lot of text, but to make it a bit more exciting, here you can see what the new weighting feature can do:

Normal fit: (bad accuracy for small magnitudes)
normal-fit

Weighted fit: (balanced the accuracy between small and large magnitudes, sacrificing accuracy at large magnitudes in favor of more accuracy for small magnitudes) 
weighted-fit

### DIFF
--- a/doc/source/examples/vectorfitting/vectorfitting_ex1_ringslot.ipynb
+++ b/doc/source/examples/vectorfitting/vectorfitting_ex1_ringslot.ipynb
@@ -56,7 +56,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "vf.vector_fit(n_poles_real=3, n_poles_cmplx=0)"
+    "vf.vector_fit(n_poles_init=3, poles_init_type='real')"
    ]
   },
   {
@@ -145,8 +145,8 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
+   "display_name": "Python 3 (Spyder)",
+   "language": "python3",
    "name": "python3"
   },
   "language_info": {
@@ -159,7 +159,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.8"
+   "version": "3.12.7"
   }
  },
  "nbformat": 4,
--- a/doc/source/examples/vectorfitting/vectorfitting_ex2_190ghz_active.ipynb
+++ b/doc/source/examples/vectorfitting/vectorfitting_ex2_190ghz_active.ipynb
@@ -67,17 +67,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "print(f'model order = {vf.get_model_order(vf.poles)}')\n",
-    "print(f'n_poles_real = {np.sum(vf.poles.imag == 0.0)}')\n",
-    "print(f'n_poles_complex = {np.sum(vf.poles.imag > 0.0)}')"
-   ]
-  },
-  {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
@@ -121,12 +110,19 @@
     "freqs = np.linspace(0, 500e9, 501) # plot model response from dc to 500 GHz\n",
     "vf.plot_s_db(freqs=freqs)"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
-   "language": "python",
+   "display_name": "Python 3 (Spyder)",
+   "language": "python3",
    "name": "python3"
   },
   "language_info": {
@@ -139,7 +135,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.2"
+   "version": "3.12.7"
   },
   "nbsphinx": {
    "timeout": 360
--- a/doc/source/examples/vectorfitting/vectorfitting_ex3_Agilent_E5071B.ipynb
+++ b/doc/source/examples/vectorfitting/vectorfitting_ex3_Agilent_E5071B.ipynb
@@ -85,7 +85,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "After trying different numbers of real and complex-conjugate poles, the following setup was found to result in a very good fit. The initial model order is $1 + 2 * 26 = 53$, which will remain unchanged during the fitting process. Other setups were also found to work well for this network (e.g. 0-2 real poles and 25-26 cc poles):"
+    "After trying different numbers and types of poles, the following setup was found to result in a very good fit. The initial model order is $2 * 27 = 54$, which will remain unchanged during the fitting process"
    ]
   },
   {
@@ -94,16 +94,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "vf.vector_fit(n_poles_real=1, n_poles_cmplx=26)"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "print(f'model order = {vf.get_model_order(vf.poles)}')"
+    "vf.vector_fit(n_poles_init=27, poles_init_type='complex')"
    ]
   },
   {
@@ -126,7 +117,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The fitted model parameters are now stored in the class attributes `poles`, `residues`, `proportional_coeff` and `constant_coeff` for further use. To verify the result, the fitted model responses can be compared to the original network responses. As the model will return a response at any given frequency, it makes sense to also check its response outside the frequency range of the original samples. In this case, the original network was measured from 0.5 GHz to 4.5 GHz, so we can plot the fit from dc to 10 GHz:"
+    "The fitted model parameters are now stored in the class attributes `poles`, `residues`, `proportional` and `constant` for further use. To verify the result, the fitted model responses can be compared to the original network responses. As the model will return a response at any given frequency, it makes sense to also check its response outside the frequency range of the original samples. In this case, the original network was measured from 0.5 GHz to 4.5 GHz, so we can plot the fit from dc to 10 GHz:"
    ]
   },
   {
@@ -205,17 +196,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "print(f'model order = {vf.get_model_order(vf.poles)}')\n",
-    "print(f'n_poles_real = {np.sum(vf.poles.imag == 0.0)}')\n",
-    "print(f'n_poles_complex = {np.sum(vf.poles.imag > 0.0)}')"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
     "print(f'rms error = {vf.get_rms_error()}')"
    ]
   },
@@ -235,12 +215,19 @@
     "fig.tight_layout()\n",
     "mplt.show()"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
-   "language": "python",
+   "display_name": "Python 3 (Spyder)",
+   "language": "python3",
    "name": "python3"
   },
   "language_info": {
@@ -253,7 +240,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.2"
+   "version": "3.12.7"
   },
   "nbsphinx": {
    "timeout": 240
--- a/doc/source/examples/vectorfitting/vectorfitting_ex4_passivity.ipynb
+++ b/doc/source/examples/vectorfitting/vectorfitting_ex4_passivity.ipynb
@@ -22,7 +22,7 @@
     "# load and fit the ring slot network with 3 poles\n",
     "nw = skrf.data.ring_slot\n",
     "vf = skrf.VectorFitting(nw)\n",
-    "vf.vector_fit(n_poles_real=3, n_poles_cmplx=0)\n",
+    "vf.vector_fit(n_poles_init=3, poles_init_type='real')\n",
     "\n",
     "# plot fitting results\n",
     "freqs = np.linspace(0, 200e9, 201)\n",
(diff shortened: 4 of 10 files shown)


Vinc0110:
Wow, these are lots of new features and improvements! Scrolling through your text description, it sounds very nice. Seems like you are the first to discover some of my dirty hacks in the auto_fit() code, for example in the integration for spurious pole detection. Maybe it's time to clean this up.

Reviewing all the code changes in detail and running some tests will take me some time. Please be patient :wink:

Vinc0110 (changes requested):
Ok, I finally did a more detailed review, but it would probably take me two more weeks to really dig into every aspect. After all, you basically did a full re-implementation of everything. I'm adding a few comments to the code as a starting point. This will certainly require some back-and-forth.

In general, there are many nice ideas and several fixes or optimizations, which are much appreciated. However, I have mixed feelings about the implementation of all these changes. It seems like someone ran a bulldozer over the existing API. None of my test scripts is working anymore. Some changes might be required, but a lot of them seem unnecessary to me.

One solution I can think of is to leave the regular (MIMO) fitting as it was and add the new SISO/MISO as a new method with it's own arguments. That might also improve the readability and limit the amount of arguments to the methods, which in my opinion should not be such a huge list. Let's keep things as simple as possible for the users. Don't forget that vector fitting is a rather specialized topic, meaning it can already be quite overwhelming for new users. Having a forest of settings and arguments to configure is not always ideal.

fnordpole:
Thanks for your review Vincent! I really appreciate it a lot! I realized from your comments that I probably changed too many names for this pull request to ever pass even if we change them all back. So maybe we should close this PR and I can make a new one with hopefully very local minimum changes so we can still add the new features but otherwise changing as little as possible.

A little bit of background why I changed some names: While reviewing the code, whenever I spent a bigger amount of time figuring out what something is, I changed the names, usually for one of those reasons:

- prefer longer names that are more descriptive instead of short names
- don't abbreviate parts of names if not absolutely necessary
- consistency: names of similar things should be similar
- comments should be complete sentences with first letter capitalized (PEP8-Standard, doensn't mean we have to adhere to it, but I just got used to read code in that style)

Changes in the API: Yes some will not work like before. And there are many parameters now to adjust the new features. So maybe I can try to change as little as possible in a new PR.

Some methods were static initially and I changed them to non-static because I needed access to the pole-groups somehow that are stored in the class. I think they can also be converted back to static by passing in the stuff that they need via arguments, for example get_model_order. I didn't think that they are meant to be used without the class.

Number of methods: I put stuff that is repeated many times into own methods, even if the methods are pretty short.
The reason is readability and not so much speed or actual need of a method. For example, if I read reset_history() I understand immediately what it's doing and I don't need to see those four lines of code resetting some arrays. Also, if we ever happen to add stuff to the history, there will be only one place to modify. Even if omega won't ever change, get_omega() looks way more readable than that 1j*network.freq... etc.

So yes, I agree with you that they are "pseudo changes" and a matter of personal taste.

Vinc0110:
It's me again with a few more general comments and suggestions. I don't mean to be overly critical, but I'm not really sure how to proceed with this PR. So here are some of my thoughts, totally open for discussion. And again, I like most of the new features very much and I do appreciate your extensive work on this topic. It's only about the structure of the implementation.
- This PR introduces many different and often unrelated changes, most of them would justify a separate PR on their own. But only a total of **2 commits** is used. This makes it very difficult now to split them. Please be more generous with commits while implementing new changes. Doing them in separate branches is also a good idea for the future ;) Then you could create 10 different PRs from your branches and we can deal with them separately. Reviewing would also be much easier then.
- Personal coding preferences should not lead to a total refactoring of the existing code. Otherwise, things might get out of control (what we have now) and everybody would change things back and forth. Now you might be familiar with the variable names and stuff, but now I'm lost in the code. So, I think existing code should be left alone as much as possible, even if you don't agree with the style. Only change it if you want to change/improve the functionality.
- Splitting functionality into (private) methods (static or not) is fine, but only do it if it serves the purpose of reusability. Having things like `omega = _get_omega()` is not much better than `omega = 2 * np.pi * f`. Having longer passages of code that is used only once should also remain, no need to move it into a new method.
- Reading the new code some more, I'm getting more and more convinced that the SISO, SIMO, ..., should have their own method(s). We should keep the regular (MIMO) vector fitting alone. This would also help in the preservation of the existing API structure.
- `_get_n_ports()` is misleading. We already have the number of original network ports in `self.network.nports`. But this method seems to provide a different number, depending on the SIMO configuration. So maybe rename it to something like `_get_simo_ports()`.
- I agree with problem in the pole initialization. Let's to it by specifying the model order instead. Then we can decide internally what type of initial poles to use and where to place them. 

Trying to preserve the existing API for a transition period, I propose something like the following structures:
- `vector_fit(n_poles_real=None, n_poles_cmplx=None, model_order=6, init_pole_spacing='lin', parameter_type='s', fit_constant=True, fit_proportional=False)`
- `auto_fit(n_poles_init_real=None, n_poles_init_cmplx=None, model_order_init=6, n_poles_add=None, model_order_groth=3, model_order_max=300, iters_start=5, iters_inter=5, iters_final=5, target_error=0.01, alpha=0.03, gamma=0.03, nu_samples=1.0, parameter_type='s')`
- `multi_siso_fit(... whatever you want)`
- `simo_fit(... whatever you want)`

This way, we can still support old arguments, if provided by old user scripts, but throw `DeprecationWarning`s and announce the new arguments as a replacement. After a while, we can then delete the old arguments for good. For example:
```
if n_poles_real is not None or n_poles_cmplx is not None:
    warnings.warn("Arguments `n_poles_real` and `n_poles_cmplx` are deprecated. Please use `model_order` instead.", DeprecationWarning)

   if model_order is None:
      model_order = n_poles_real + 2 * n_poles_cmplx
      # maybe a bit more sanity checking required here...
```

Let's keep the arguments `alpha`, `gamma`, and so on. Most users should not have to change them anyway. If they do, they can look up the meaning in the docstrings and give them different names in their user scripts. I don't think it's worth to break the API for such minor things. For `target_error`, I think the name is actually good already. It is basically just a "wish" for the final error. In case of strong noise, that wish might never become real.

fnordpole:
You are totally right, that's also why I first suggested to close this PR and make smaller ones that can be easier reviewed and merged.

I actually had a bunch of smaller commits in my private repo, but none of them had a fully working vectorFit class because I worked myself through all the methods one by one and modified the data structures in every method to support the pole groups feature (siso, simo, mimo, miso, arbitrary). I did commit all of them separately but those can't be separate pull requests because the make only sense all at once.

In the end I decided to rebase all those commits into one single big commit that actually represents a working total thing. It's a bit unfortunate that the changes are in so many methods. So maybe it's a good idea to have an extra siso_fit() as you suggested and only that gets the new data structures leaving the other methods untouched.

I think your suggestions are a good way to proceed. I can make a new PR once I am ready isolating the changes into separate packages. Right now I am working on the passivity test and passivity enforcement, which doesn't work yet with the pole groups because of symmetry problems. So it's more or less useless without the passivity algorithms also supporting it. When I have that working I plan to make a new PR (or update this one) with a smaller set of changes.

Vinc0110:
Ok, I understand. Of course, it's fine if you rather delete this PR and start fresh. Whatever works best...

To save some work for both of us, would it make sense to first only create and discuss the skeletons until everybody is happy? Not sure, it's just an idea.

If the algorithms for SISO fitting etc. all require different data structures and also different implementations of the auxiliary algorithms (passivity test, passivity enforcement, ...), we could even think about creating a new class within the `VectorFitting` module for them. So far, the `VectorFitting` module only has one class (`VectorFitting`). Not sure, but I think it will have to be in the same file (`vectorFitting.py`) to be a member of the module. 

During initialization, we currently just import everything from `vectorFitting.py`:
https://github.com/scikit-rf/scikit-rf/blob/b001594fc3c6d783d2bd2f26148bcdc238735c43/skrf/__init__.py#L43

We could even create a third (private?) class for the internal core functionality such as pole relocation, residue fitting, etc., implemented in a way that supports the old and the new algorithms. Or it can be a collection of static functions.

fnordpole:
I worked a bit on the passivation and it doesn't seem to work reliably for Multi-SISO.
I'll close this PR for now and once I got it working I'll text you and then we can discuss what to do in a potential new PR.

PR #1242, opened 2025-01-19
Vector Fitting: DC point enforcement

Vinc0110:
This PR adds a new feature to the vector fitting functions to enforce the dc point in the solution, if provided in the data.
The enforcement is realized using linear equality constraints ~~in the pole relocation and~~ in the residue fitting routine. 

In the vector fitting book (S. Grivet-Talocia, B. Gustavsen, "Passive Macromodeling", Wiley 2016), this method is described in a very general way. I adopted and simplified it, because most of it does not seem to be required if only the dc point is to be enforced. In general, all sorts of frequency samples could be enforced in the solution.

As always, everything comes at a price. In case of dc point enforcement, we pay with accuracy in the remaining fit. The difference is not huge, but it's there. However, the fit of the dc point is nearly perfect, as desired. In case of the automatic fitting with `auto_fit()` the model order can also be slightly greater.

Currently, I have not added an argument to the fitting functions to turn dc point enforcement on or off. Instead, it's always enforced if the dc point is sampled in the data. I can't think of a use case where someone would not want this, but who knows?. Any other opinions on this?

Here is a comparison with the current code (branch _master_) and the new code using dc point enforcement (branch _vf_dc_) for a large 18-port network (324 frequency responses).

- _master_ using `auto_fit()`: `dc error magnitude < 1e-5`
- _vf_dc_ using `auto_fit()`: `dc error magnitude < 3e-15`

Via_array_impedance_dc-error_master
Via_array_impedance_dc-error_vf_dc

### DIFF
--- a/skrf/tests/test_vectorfitting.py
+++ b/skrf/tests/test_vectorfitting.py
@@ -74,17 +74,18 @@ class VectorFittingTestCase(unittest.TestCase):
         # rough check on fit quality
         self.assertLess(vf.get_rms_error(), 0.2)
 
-        # evaluate model error at the dc point
+        # evaluate model error at the dc point (real part)
+        # the dc point should always be real (it still often has a tiny imaginary part)
         vf_fit = np.empty(nw.nports ** 2, dtype=complex)
-        abs_errors = np.empty(nw.nports ** 2)
+        abs_real_errors = np.empty(nw.nports ** 2)
 
         for i in range(nw.nports):
             for j in range(nw.nports):
                 vf_ij = vf.get_model_response(i, j, nw.f[0])
                 vf_fit[i * nw.nports + j] = vf_ij
-                abs_errors[i * nw.nports + j] = np.abs(vf_ij - nw.s[0, i, j])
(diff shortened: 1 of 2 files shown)


Vinc0110:
Test with `skrf/tests/cst_example_4ports.s4p`
cst_example_4ports_dc-error_master
cst_example_4ports_dc-error_vf_dc

Vinc0110:
Test with skrf/tests/cst_example_6ports.s6p
cst_example_6ports_dc-error_master
cst_example_6ports_dc-error_vf_dc

fnordpole:
Very nice! I like it! I recently read a paper related to that: 

"DC-Preserving Passivity Enforcement for S-Parameter Based Macromodels", Dirk Deschrijver, Tom Dhaene, IEEE TRANSACTIONS ON MICROWAVE THEORY AND TECHNIQUES, VOL. 58, NO. 4, APRIL 2010

Maybe both can be combined so that the passivity enforcement can't negatively affect the now super good DC point.

Vinc0110:
Thanks for mentioning the paper. I will give it a try.

PR #1247, opened 2025-02-05
Vector Fitting: State-space topology for equivalent circuits

Vinc0110:
This PR implements another equivalent circuit topology for vector-fitted networks in scattering representation. In contrast to the previous admittance-based and impedance-based topologies, this one is a direct implementation of the state-space equations. While it still involves many controlled source (mostly VCCS, not so bad), the resulting network is much smaller in terms of node count, i.e. number of unknowns to solve for. LTspice is still not impressed, but ngspice and Xyce run much faster with the state-space circuits.

I will add some documentation later in another PR, once this one is discussed, finalized and merged.

Here are some runtime comparisons for small and large networks with small and large models:

**`4port_reduced`, 4-port network, model order 105:**
- ngspice (admittance variant): 788.640 s
- ngspice (impedance variant): 0.390 s
- ngspice (state-space variant): 0.105 s
- Xyce (admittance variant): 21.033 s
- Xyce (impedance variant): 4.84399 s
- Xyce (state-space variant): 0.765833 s
- LTspice (admittance variant): 0.249 s
- LTspice (impedance variant): 0.233 s
- LTspice (state-space variant): 0.154 s

**`Backplane`, 8-port network, model order 385:**
- ngspice (admittance variant): did not finish in reasonable time
- ngspice (impedance variant): 97.564 s
- ngspice (state-space variant): 70.181 s
- Xyce (admittance variant): did not finish in reasonable time
- Xyce (impedance variant): 133.325 s
- Xyce (state-space variant): 18.8203 s
- LTspice (admittance variant): 8.119 s
- LTspice (impedance variant): 8.763 s
- LTspice (state-space variant): 8.661 s

**`Via_array`, 18-port network, model order 13:**
- ngspice (admittance variant): did not finish in reasonable time
- ngspice (impedance variant): did not finish in reasonable time
- ngspice (state-space variant): did not finish in reasonable time
- Xyce (admittance variant): did not finish in reasonable time
- Xyce (impedance variant): 53.4626 s
- Xyce (state-space variant): 5.91896 s
- LTspice (admittance variant): 0.251 s
- LTspice (impedance variant): 0.237 s
- LTspice (state-space variant): 0.787 s

### DIFF
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -2574,15 +2574,18 @@ class VectorFitting:
                     d = self.constant_coeff[idx_S_i_j]
                     # e = self.proportional_coeff[idx_S_i_j]    # E is currently ignored
 
-                    # VCCS and CCCS adding their currents to represent the incident wave a_j
-                    gain_vccs_a_j = 1 / 2 / np.sqrt(z0_j)
-                    gain_cccs_a_j = np.sqrt(z0_j) / 2
+                    if d != 0.0:
+                        # avoid zero-valued coefficients (in case of fit_constant=False)
 
-                    # input a_j is scaled by constant term d_i_j and by current gain for b_i
-                    g_ij = gain_b_i * d * gain_vccs_a_j
-                    f_ij = gain_b_i * d * gain_cccs_a_j
-                    f.write(f'Ga{i + 1}_{j + 1} {node_ref_i} s{i + 1} p{j + 1} {node_ref_j} {g_ij}\n')
-                    f.write(f'Fa{i + 1}_{j + 1} {node_ref_i} s{i + 1} V{j + 1} {f_ij}\n')
+                        # VCCS and CCCS adding their currents to represent the incident wave a_j
+                        gain_vccs_a_j = 1 / 2 / np.sqrt(z0_j)
+                        gain_cccs_a_j = np.sqrt(z0_j) / 2
+
+                        # input a_j is scaled by constant term d_i_j and by current gain for b_i
+                        g_ij = gain_b_i * d * gain_vccs_a_j
+                        f_ij = gain_b_i * d * gain_cccs_a_j
+                        f.write(f'Ga{i + 1}_{j + 1} {node_ref_i} s{i + 1} p{j + 1} {node_ref_j} {g_ij}\n')
+                        f.write(f'Fa{i + 1}_{j + 1} {node_ref_i} s{i + 1} V{j + 1} {f_ij}\n')
 
                     # each residue rk_i_j at port i is multiplied by its respective state signal xk_j
                     for k in range(len(self.poles)):
--- a/skrf/vectorFitting.py
+++ b/skrf/vectorFitting.py
@@ -2543,9 +2543,9 @@ class VectorFitting:
                         f.write(f'Gd{i + 1}_{j + 1} {node_ref_i} s{i + 1} p{j + 1} {node_ref_j} {g_ij}\n')
                         f.write(f'Fd{i + 1}_{j + 1} {node_ref_i} s{i + 1} V{j + 1} {f_ij}\n')
 
-                    if e != 0.0:
+                    if build_e and e != 0.0:
                         # avoid zero-valued coefficients (in case of fit_proportional=False)
-                        # proportional coefficients require an extra node for the differentiation using a capacitor
+                        # proportional coefficients require an extra node for the differentiation using an inductor
                         # [Y(s) ~ s * E * U(s)]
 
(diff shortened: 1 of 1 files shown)


fnordpole:
Great work, this is really cool! It's getting better and better. I didn't test it yet but I am exited to try it out it soon. I read a short section about state space synthesis in "Circuit synthesis techniques of rational models of electromagnetic systems: A tutorial paper", Romano et al. but I didn't dive into the details. Just out of curiosity, is your code based on any paper or new?

fnordpole:
I have the impression that the proportional term gets very little attention in most papers. (for example in the dc-preserving VF implementation that we worked on recently, there's no mentioning of the proportional term in the paper). 

I have not seen it being used in practice yet. Maybe it is used when fitting other types of parameters like Z or Y but some of the algorithms implemented right now in the vectorFitting class are restricted to S-parameters already, if I remember correctly. I would just raise a warning. 

The gmin problem I have seen in every simulator not just LTspice. Spectre automatically adds a gmin if there's no DC path to ground. In XYCE there are options to enable something similar (.preprocess addresistors nodcpath 1G
 and .preprocess addresistors oneterminal 1M).

Maybe also in LTspice something like this can be enabled. Maybe XYCE and ngspice are a bit more tolerant to nodes without DC path to ground. In my experience this can work for small circuits but large circuits with many nodes without DC paths to ground will lead to convergence problems. 

I am not sure what's the best way to add the gmins. It can be done right in the synthesizer code as you did it. It can also be done "by the user" via the simulator. The way via the simulator maybe has the advantage that you can vary gmin without having to recreate the model netlist. But with the option to enable gmins or disable them both ways are possible so that sounds like a good idea.

Vinc0110:
Yes, they don't often mention the proportional term. I guess it's not really part of the usual state-space representation used in control theory and alike, see https://en.wikipedia.org/wiki/State-space_representation.

Anyway, I think I have it working now with the proportional terms. I ended up renaming nodes and sources. I should be fine, but let's test it thoroughly again. @fnordpole It would be nice if you could run some tests as well.

Simple test with my beloved ring slot with and without the proportional term:

**`vector_fit(3, 0, fit_proportional=False)`:**
ringslot_False_fitted
image

**`vector_fit(3, 0, fit_proportional=True)`:**
ringslot_True_fitted
image

The proportional terms are only really useful for Y and Z parameters. For S, they are not required and they are somewhat violating physics. This becomes obvious if you think about passivity with a proportional term. Asymptotic passivity is impossible if you have a term that grows infinitely with frequency. I guess that's why they don't even mention it in most papers.

Vinc0110:
Regarding `g_min`, I think in these circuits it really does have any significant effect on the results. 

The "problematic" nodes have 3 or 4 controlled current sources in parallel with a capacitor. One of the current sources is controlled by its own voltage, i.e. the voltage over the capacitor. This could also be done with a resistor (as in the Romano paper), but the values for the conductance/resistance are negative, so I kept the VCCS. However, there is always a proper dc path the ground, but it's not in the form of an explicit resistor.

Still, adding an extra R = 1 / g_min in parallel should not hurt. The transconductances of the controlled sources equal the real part of the poles, which usually have a large values (like >1e9). In other words, those nodes have a really low impedance. Adding `g_min = 1e-12` affects the result in my example only at the 21st decimal place, right?.

Vinc0110:
Ok, I did some accuracy tests using the latest state-space implementation in Xyce with the examples from above (including `g_min = 1e-12`). Also, I fitted the 4port with more poles (model oder 317, "4port_full"). 

The error magnitude in decibels are calculated with `10 * np.log10(np.abs(vf_ij - sim_ij))` for each of the vector-fitted responses `vf_ij` and the simulates response `sim_ij` from Xyce, evaluated at the simulated frequencies. The error is always below -90 dB, which is probably as accurate as it gets. 

Interestingly, there sometimes are these 10 dB steps, which I've already seen in previous tests. Not sure what happens there. Is it the limited precision of the floating point numbers somewhere along the toolchain (in the circuit synthesis, in Xyce, or in the sNp file)?

In any case, I think this PR is ready. Other tests or comments are always welcome, of course.

4port_reduced_stsp_xyce_error
4port_full_stsp_xyce_error
Backplane_stsp_xyce_error
Via_array_stsp_xyce_error

Vinc0110:
I just realized my brain was twisted. In the book, they also claim the VCCS on the "problematic" node can be replaced by a (positive) resistor. It's totally right, I was just confused with the sign. With this there is an intrinsic dc path to ground and we don't need to worry about any g_min.

I ran all of my above tests again and the results are pretty much identical. The error is always below -90 dB.

jhillairet:
Hi @Vinc0110 , 
Is this PR ready to be merged?

Issue #1239, opened 2025-01-15
NetworkSet.from_mdif() fails to import MDIF files with noise parameters


jrmagers:
**Describe the bug**
`NetworkSet.from_mdif()` fails to import MDIF files with noise parameters.

**To Reproduce**
```Python
import skrf

ns_from_mdif = skrf.NetworkSet.from_mdif(r"amplifier.mdf")

Traceback (most recent call last):
  File "c:\repos\github\scikit-rf\awr\test.py", line 15, in <module>
    ns_from_mdif = skrf.NetworkSet.from_mdif(r"amplifier.mdf")
  File "C:\repos\github\scikit-rf\skrf\networkSet.py", line 348, in from_mdif
    return Mdif(file).to_networkset()
  File "C:\repos\github\scikit-rf\skrf\io\mdif.py", line 376, in to_networkset
    ns = NetworkSet(self.networks)
  File "C:\repos\github\scikit-rf\skrf\networkSet.py", line 160, in __init__
    raise(ValueError('All elements in list of Networks must have same number of ports'))
ValueError: All elements in list of Networks must have same number of ports
```

**Expected behavior**

A `NetworkSet()` object should be returned with the noise parameters populated.

**Data**

`amplifier.mdf` is attached inside `amplifier.zip`. This MDIF file has been validated to import into Microwave Office 24.1. See PR https://github.com/scikit-rf/scikit-rf/pull/1238.

**System**
 - OS: Windows 11
 - scikit-rf version: v0.1.5

[amplifier.zip](https://github.com/user-attachments/files/18430677/amplifier.zip)

PR #1258, opened 2025-02-25
Enable read of noisy mdf file

FranzForstmayr:


FranzForstmayr:
Closes #1239

FranzForstmayr:
The test shows that the "content" of the files is correct, however we loose some information from the comments currently. We can fix this later if really needed.

jhillairet (approved):

Issue #1249, opened 2025-02-08
5 Port MDIF File fails to import to NetworkSet


jrmagers:
**Describe the bug**

A 5-port MDIF file created from the s-parameters of the SP4T switch [ADRF5054](https://www.analog.com/en/products/adrf5045.html). This part is confirmed to correctly import into AWR Microwave Office.

**To Reproduce**
```Python
import skrf

filename = "ADRF5054.mdf"
adrf5054 = skrf.NetworkSet.from_mdif(filename)

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File Untitled-1:6
      [3](untitled-1:3) import skrf
      [5](untitled-1:5) filename = "ADRF5054.mdf"
----> [6](untitled-1:6) adrf5054 = skrf.NetworkSet.from_mdif(filename)

File c:\venv\sandbox\Lib\site-packages\skrf\networkSet.py:348, in NetworkSet.from_mdif(cls, file)
    [329](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:329) """
    [330](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:330) Create a NetworkSet from a MDIF file.
    [331](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:331) 
   (...)
    [345](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:345) 
    [346](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:346) """
    [347](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:347) from .io import Mdif
--> [348](file:///C:/venv/sandbox/Lib/site-packages/skrf/networkSet.py:348) return Mdif(file).to_networkset()

File c:\venv\sandbox\Lib\site-packages\skrf\io\mdif.py:86, in Mdif.__init__(self, file)
     [84](file:///C:/venv/sandbox/Lib/site-packages/skrf/io/mdif.py:84) self._comments = self._parse_comments(fid)
     [85](file:///C:/venv/sandbox/Lib/site-packages/skrf/io/mdif.py:85) self._params = []
---> [86](file:///C:/venv/sandbox/Lib/site-packages/skrf/io/mdif.py:86) self._networks = self._parse_mdif(fid)

[...]

PR #1265, opened 2025-03-12
Rewrote GMDIF read/write to be compatible with MWO

gurthang2:
- Added in preprocessing to handle larger MDIF files whose lines overflow onto the next line. 
- Rewrote MDIF write functionality to match MWO format (and thus work with above changes)
- Added 3 and 5 port GMDIF tests. 

This implements my fix to #1249. When writing, I replaced the space character with a tab character in the continuation lines. This matches MWO's output and makes skrf generated MDF files work with the preprocessing fix. All new outputs are recognized as valid files in MWO v17.

Apologies if I am overstepping here; I am not sure what the protocol is for contributing.


FranzForstmayr:
Hi @gurthang2 
> Apologies if I am overstepping here; I am not sure what the protocol is for contributing.

This is perfectly fine, we appreciate any contribution!

I hope I find some time the next days to give you a more detailed review!

jhillairet (approved):
LGTM. However I have the impression that the original issue is more on the MWO side... Does this issue have been raised to MWO people?  (there is not blank line space linter issues which has to be fixed)

FranzForstmayr:
Sorry for the long delay, looks fine as long you fix the remaining linter issues.

gurthang2:
No worries, I am pretty busy with actual work right now but I will try and find some time to working on the linting issues this weekend. 

Although, after some thought, I am not sure if you guys should push this. My implementation is completely reliant on the 'continuation lines' starting with at a `\t` character. If that is not true, my code breaks down. This bug is not a problem with touchstone files since you guys dynamically calculate the number of data points needed per frequency using the `ParserState` class. 

I think it would be better if the read MDF functionality used the `ParserState` class or something similar to it. That way it will (in theory) work regardless of what delimiter the file uses. Also, in that case we wouldn't have to modify the write touchstone functionality. That is a part of this codebase that I would rather not be responsible for breaking (it's probably the feature I use the most). 

I guess you could push this PR as a temporary fix, but really I don't think it should be permanent. Let me know what you guys want to do.

jhillairet:
How about this PR @gurthang2 ? @FranzForstmayr , should we merge it ?

FranzForstmayr:
The issue from my side is currently that I cannot test any compability with MWO as I don't have access.
The solution itself could work at least for MWO, however it's not standardized (for mor than two ports) and can break any time.

jhillairet:
OK. So keeping this PR open for the moment.

PR #1272, opened 2025-04-10
Update reference to nfmin in nf_circle 'see also' section

miek:
Just noticed that the [docs](https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.network.Network.nf_circle.html#skrf.network.Network.nf_circle) link to a non-existent property `nf_min`, this corrects that to `nfmin` and also adds a reference to `nfmin_db` (since it took me a minute to realise `nfmin` wasn't already in dB).


jhillairet:
Thanks!

PR #1278, opened 2025-04-22
Fix time_gate documentation

shanham:
The documentation for time_gate seems to indicate that both sec and nsec are the default. PR fixes this.


jhillairet (approved):
Thank you for your contribution !

The linter is picky: could you remove the trailing space after "nanoseconds" ?

PR #1290, opened 2025-07-03
Add skrf version in touchstone header comment

daholzfeind:
This PR adds the version of skrf to the touchstone header, which is useful for later references.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -175,6 +175,7 @@ from scipy import stats  # for Network.add_noise_*, and Network.windowed
 from scipy.integrate import cumulative_trapezoid
 from scipy.interpolate import interp1d  # for Network.interpolate()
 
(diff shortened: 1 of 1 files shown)


jhillairet (approved):
OK, why not. should we add "v" before the number or is it clear enough like that?

daholzfeind:
For me, it's clear. But if you want, I can add that.

jhillairet:
by the way, did you run into any issue with Touchstone and versions of scikit-rf?

daholzfeind:
No issues. so far  In any case, it's helpful to have a reference for the skrf version used for calculating error boxes. Similar to VNAs that write their FW version to the Touchstone file.

jhillairet:
thanks @daholzfeind !

PR #1304, opened 2025-09-03
Allow `skrf.Network` to input touchstone 2.0 file without `*.ts` extension

SHF101202021:
Hi Scikit-RF Community,

For EDA tools like Helic, a EM tool to extract electrical properties of DIE TOP layouts,  the extension of generated touchstone 2.0 file is `*.sp` rather than `*.ts`.

For this situation, when file extension pattern is not `\.[sS]\d+[sP]`, we should detect if `^\[Version\]` exists for the first line of the given file.
If not, we then continue to reject analyze the input file.


FranzForstmayr:
Could you please add an example file and create a unit test for the file?

SHF101202021:
@FranzForstmayr 
OK.

SHF101202021:
@FranzForstmayr 
Examples and pytest codes are added.

jhillairet:
> @FranzForstmayr Examples and pytest codes are added.

Thank you. As this is just for testing purpose, could you reduce a bit the size (number of frequency points) of the test files?

SHF101202021:
@jhillairet 
Already shrink the frequency points.

SHF101202021:
@FranzForstmayr  @jhillairet 

Cuz this PR is approved, I'll close it.
Thank you all.

jhillairet:
Thank you @SHF101202021

PR #1309, opened 2025-09-19
Copper Mountain Technologies Cobalt series virtual instrument interface.

bdelwood:
This PR implements a VI module for the [CMT Cobalt series of VNAs](https://coppermountaintech.com/50-ohm-vnas/). I have only implemented the SCPI commands/interfaces that I commonly use; the [list of supported commands](https://coppermountaintech.com/wp-content/uploads/2020/09/SxVNA-SCPI-Programming-Manual-2.pdf) is quite extensive for the CMT VNAs.

I implemented a few tests which rely on running the control software. Fortunately, Copper Mountain provides the [control software for their VNAs as freeware](https://coppermountaintech.com/download-free-vna-software/), and they come with a demo/simulation mode--so one can run tests without having the actual hardware. I have run these tests both with the control software running with a real C4220 and in the C4220 demo mode.


leonknight99-flann:
Just noticed that we have both been working on the same driver for cmt! Having a quick look through yours compared to mine you have implemented the `get_sdata()` function whereas I have implemented the `get_snp_network()`. A lot of the standard functions are the same just with slight text changes. 

The way I implemented `get_snp_network()` was to make use of a windows temp file to export the current calibrated touchstone file and then reopen as a skrf class.

I also only have a C4220 to validate the testing although made the class more generic to hopefully include other devices (as they all seem to use the S#VNA software.

See what you think

[https://github.com/leonknight99-flann/scikit-rf/blob/VNA_Dev/skrf/vi/vna/copper_mountain/cmt.py](url)

jhillairet:
Tests are failing due to HP8510C tests, but I can't reproduce it locally. Seems a CI-related issue. Should we ignore these RessourceWarnings generated by mocker?

bdelwood:
Sorry for the long delay. Somehow thought this was merged! 

@jhillairet I have been using this for a year in the lab without issues. But I reproduced the earlier issue with the test: failed CMT connections leaked sockets, with warnings surfacing during later HP8510C tests. I have fixed that. 

@leonknight99-flann I ended up needing `get_snp_network` at some point, so I have added it.

Rebased on current master.

jhillairet (approved):
Thanks for the update ; please also check the discussion #1434 which is of interest for you probably.

PR #1312, opened 2025-09-22
media.py: reword misleading z0 documentation, close #1310.

### LINKED ISSUE #1310: Misleading Documentation for skrf.media.Media.match()
The documentation of `skrf.media.Media.match()` says:

> z0 (number, or array-like or None) – port impedance. Default is None, in which case the Media’s z0 is used. This sets the resultant Network’s z0.

But according to the actual code:

        if z0 is None:
            if self.z0_port is None:
                z0 = self.z0
            else:
                z0 = self.z0_port

Thus, the documentation is misleading. Clearly, the wording of the documentation was not updated for the new impedance API syntax. This is problem is important because the documentation for `capacitor()` and `inductor()` redirects readers to `match()`. Without clarification, the reference impedance of inductance or capacitance will be confusing. 

I suggest the following:

> z0 (number, or array-like or None) – port impedance. This sets the resultant Network’s z0. Default is None, in which case the Media’s `z0_port` is used. By default, a Media's `z0_port` is the same as `z0`, but it can be explicitly set for renormalization to a new reference impedance.

biergaizi:
The documentation of `load()`, `match()`, `short()`, and `thru()` says:

> z0 (number, or array-like or None) – port impedance. Default is None,
> in which case the Media’s z0 is used. This sets the resultant Network’s z0.

But the documentation is misleading. Clearly, the wording was not updated after the new impedance API syntax has landed. This is problem is important because the documentation for `capacitor()` and `inductor()` redirects readers to `match()`. Without clarification, the reference impedance of inductance or capacitance will be confusing.

The new wording is:

> z0 (number, or array-like or None) – port impedance.
> This sets the resultant Network’s z0. Default is None, the Media’s
> z0_port value is used. If z0_port is also None (by default), in which case
> the Media's native z0 is used. A Media's z0_port is an optional attribute for
> renormalizing resultant Network to a new reference impedance that differs
> from a Media's native z0.

### DIFF
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -465,10 +465,18 @@ class Media(ABC):
         nports : int
             number of ports
         z0 : number, or array-like or None
-            port impedance. Default is
-            None, in which case the Media's :attr:`z0` is used.
-            This sets the resultant Network's
+            port impedance. This sets the resultant
+            :class:`~skrf.network.Network`’s
             :attr:`~skrf.network.Network.z0`.
+            Default is None, the Media’s :attr:`~skrf.media.Media.z0_port`
+            value is used. If :attr:`~skrf.media.Media.z0_port` is
+            also None (by default), in which case the Media's native
+            :attr:`~skrf.media.Media.z0` is used.
+            A Media's :attr:`~skrf.media.Media.z0_port` is an optional
+            attribute for renormalizing resultant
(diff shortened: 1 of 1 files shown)


jhillairet (approved):

jhillairet:
Thanks 👍 

Ok for merging it?

biergaizi:
Yes please, let's merge it.

Issue #506, opened 2021-08-02
Type hints in API docs


Vinc0110:
Now that scikit-rf starts to have type hints included in many classes, I'd like to propose a change in the `autodoc` configuration for the generation of the API documentation with Sphinx.

In summary, I think we should set `autodoc_typehints = "description"`.

Details:

Currently, the displayed signatures of class functions include the type hints, which provides redundant information and degrades readability when many parameters are listed. For example, here is a screenshot of the current `VectorFitting.vector_fit()` [documentation](https://scikit-rf.readthedocs.io/en/latest/api/vectorFitting.html#skrf.vectorFitting.VectorFitting.vector_fit) with a rather messy signature at the top:

![image](https://user-images.githubusercontent.com/63644854/127916398-4f45e8c2-d2d7-4eff-9e9c-63ab0758b3c2.png)

This behavior can be changed in the Sphinx configuration, for example by setting `autodoc_typehints = "description"`. See www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints

For comparison, here is a screenshot of a [NumPy function doc](https://numpy.org/doc/stable/reference/generated/numpy.linalg.svd.html), which I think looks much nicer:

![image](https://user-images.githubusercontent.com/63644854/127916763-827bcbfc-5518-48d5-a4c1-8d0a7f01ad6d.png)

FranzForstmayr:
In addition to that, the type in the docstring is redundant, it's enough to annotate the parameter in the signature.
I already thought of that when adding the first type hints in `network.py`, but the documentation style seemed to be incomplient with the `autodoc-typehints` scheme. However it seems that this should be possible [https://github.com/agronholm/sphinx-autodoc-typehints#compatibility-with-sphinxextnapoleon](typehints with napoleon)

Vinc0110:
You are right, but I'm not sure if it's worth the extra effort of installing and configuring the extension and updating all the docstrings (do they have to be updated?). After all, the function signatures have already been updated and the rendered results on RTD would be exactly the same as they currently are.

jhillairet:
Looking to the issues in autodoc-typehints, [in particular this one](https://github.com/agronholm/sphinx-autodoc-typehints/issues/167), it seems it is redundant to install autodoc-typehints and to define:
`autodoc_typehints = "description"` 
in the sphinx config. 

[...]

PR #1323, opened 2025-09-28
Documentation improvements

jhillairet:
- added some 2025 citations
- fixed sphinx warnings related to docstring format
- fixed several English typos
- added some descriptions of TypeDict constants


jhillairet:
I also wanted to improve the way the documentation renders some Literal parameters (as a follow-up of #506),  like for example for the Frequency class:  

image

However, I didn't manage to obtain a better result... Adding `sphinx-autodoc-typehints` (even after the napoleon extension in the config file as recommended) makes the doc building crash for a reason I didn't manage to find...

Issue #1313, opened 2025-09-22
Reference Impedance is not Taken into Account, Creating Incorrect Error Networks


biergaizi:
In `scikit-rf`, the `skrf.calibration.calibration.Calibration` logic works as follows:

1. Solve error coefficients using a specific algorithm, such as SOLT, SDDM, etc.
2. Write solved error coefficients into `self._coefs`.
3. When the error network is requested, construct `skrf.Network` from raw `self._coefs`.

Unfortunately, this only works for a 50 Ω system impedance. If other system impedance is used, the returned error network will have an incorrect reference impedance.

To solve this problem, I propose adding a key called `reference impedance` into self._coefs, which is set according to a calibration input network (since we already require all calibration inputs have the same reference impedance, any network should work).

Also, calibrating networks with mixed-impedance data points is accepted, only a warning is generated (because this use case is untested). For future extensibility, this `reference impedance` should be a per-frequency list instead of a single value.

This is required to fully fix #902 and #1307.

I'm working on an implementation.

Issue #1307, opened 2025-09-08
SOLT (TwelveTerm) calibration results degrade when the impedance of the standards or the port impedance are not 50Ω


Hasko7:
Hi all,

The SOLT calibration (both TwelveTerm and SOLT) yields poorer results when the impedance of the standards deviates from 50 ohms and no port impedance of 50 ohms is enforced. This causes all ideal and measured standards to use the characteristic impedance of the line as the port impedance. Therefore, after calibration, the result should actually be perfect, i.e., there should be no deviation from the DUT — just as it is with mTRL. The reason is that SOLT does not always normalize to 50 ohms but rather uses the impedance of the standards as the basis for the calibration. Nevertheless, the results become worse when the characteristic impedance of the standards deviates from 50 ohms.

Here is an example:

```python
# %%
import skrf as rf
from skrf.media import CPW, Coaxial
import skrf
rf.stylely()

import matplotlib.pyplot as plt
import numpy as np

freq = rf.F(1, 100, 201, 'GHz')
cpw = CPW(freq, w=35e-6, s=51e-6, ep_r=12.9,
                     t=5e-6, rho=2e-8) 
print(cpw.z0[0])
f_ghz = cpw.frequency.f*1e-9

# %%
# Generate random error boxes
X = cpw.random(n_ports=2, name='X')
Y = cpw.random(n_ports=2, name='Y')

# Random switch terms
gamma_f = cpw.random(n_ports=1, name='gamma_f')
gamma_r = cpw.random(n_ports=1, name='gamma_r')

short = cpw.inductor(L=5.8e-12) ** cpw.short(nports=1)
short = skrf.two_port_reflect(short, short)
short.name = 'short'

open = cpw.shunt_capacitor(C=6.1e-15) ** cpw.open(nports=1)
open = skrf.two_port_reflect(open, open)
open.name = 'open'

# thru und thru mit offset
thru = cpw.line(d=0, name='thru')


## Load aus Serienwiderstand + Induktivität (R + 1j*w*L)
load_res = cpw.resistor(cpw.z0[0].real) ** cpw.short(nports=1)
match_l = -10e-12
load = cpw.inductor(L=match_l) ** load_res
load = skrf.two_port_reflect(load, load)
load.name = 'load'

SOLT_ideals = [short, open, load, thru]
SOLT_measured = [X**k**Y for k in SOLT_ideals]
[...]

PR #1324, opened 2025-09-29
calibration.py: implement non-50 ohm calibration, close #902 #1307 #1313

### LINKED ISSUE #902: OnePort calibration `apply_cal` incorrect result for Z0 != 50 ohms
**Describe the bug**
The `apply_cal` method creates a new `Network` which will use z0=50 default with no regard to that the given network z0 or or what all the network used to do the calibration used as z0.

So if you're operating in a system with z0=120 for instance, this will end up failing and producing bogus results because that z0=50 will enter in some math when doing the final `return er_ntwk.inv**ntwk`

**To Reproduce**
Run any attempt of calibration in a system where z0 is != 50

**Expected behavior**
Things to produce correct result without making any assumption about the system impedance.

**System**
 - OS: Linux python 3.9
 - scikit-rf version 0.26.0

biergaizi:
This commit implements calibration for non-50 ohm system impedance by correctly tracking the reference impedance of the DUT, the error networks, and the calibration coefficients. The logic works as follows:

1. Ensure all calibration standard definitions and measurements have the same reference impedance at all ports, and for all frequencies.

2. Add a new entry to the 3 terms, 8 terms and 12 terms calibration coefficients, named "system impedance". The system impedance is extracted from the reference impedance of the first measured network (this information extraction convention is already used in the Calibration class).

3. When running apply_cal(), each calibration subclass must write this information to the coefficient dictionary. When creating error networks, the system impedance is used as the reference impedance. Since the reference impedance is always imported and exported together with the dictionary, a reconstructed calibration instance preserves the information.

4. It's named the "system impedance" to highlight the fact that its the reference impedance used to calculate the error terms and error boxes, its behavior is similar to the VNA measurement systems: a global setting, constant at all ports, and is unchanged even after TRL calibration.

5. Mixed port reference impedance is not implemented. Currently, all calibration algorithms assume a constant system impedance. This feature will be extremely difficult, if ever possible to support, a major redesign would be necessary. However, implementing per-frequency reference impedance is in principle possible and much easier. Thus, for future extensibility, the "system impedance" is a 1D array at each calibrated frequency point.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -3542,18 +3542,24 @@ class NISTMultilineTRL(EightTerm):
         Calibration.from_coefs_ntwks
 
         """
-        # assigning this measured network is a hack so that
+        # assigning garbage data as "measured" networks is a hack so that
         # * `calibration.frequency` property evaluates correctly
         # * __init__() will not throw an error
+        # Note that z0_ref is meaningful because it's used to set the
+        # self.z0_ref attribute
         n = Network(frequency = frequency,
-                    s = rand_c(frequency.npoints,2,2))
+                    s = rand_c(frequency.npoints,2,2),
+                    z0=coefs["system impedance"]
+                    )
         measured = [n,n,n]
 
         if 'forward switch term' in coefs:
             switch_terms = (Network(frequency = frequency,
-                                    s=coefs['forward switch term']),
+                                    s=coefs['forward switch term'],
+                                    z0=coefs["system impedance"]),
                             Network(frequency = frequency,
-                                    s=coefs['reverse switch term']))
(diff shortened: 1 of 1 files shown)


skrf/calibration/calibration.py
```diff
@@ -5947,7 +6054,10 @@ def embed(self, ntwk):
 
         # Consistent internal port Z0.
         nout.z0 = 50
-        Z = Network(s=Z, frequency=nout.frequency, z0=50)
+        Z = Network(
+            s=Z, frequency=nout.frequency,
+            z0=self.coefs['system impedance']
```
jhillairet: self.coefs is a list, isn'it?
```suggestion
            z0=self.coefs[0]['system impedance']
```

PR #1330, opened 2025-10-12
Add API to set/get the global numpy RNG, close #1314.

### LINKED ISSUE #1314: Random Seed Should be Settable via Global Variable or a Numpy Generator Instance in All Random Functions
In Monte Carlo analysis, including unit tests of scikit-rf itself, it's important to have reproducible randomness so that each failing test case can be replicated exactly when necessary. However, this is currently not possible in scikit-rf.

The `skrf.media.Media.random()` method generates random numbers as follows:

    result.s = mf.rand_c(self.frequency.npoints, n_ports, n_ports)

This is passed to `skrf.mathFunctions.rand_c()`:

    rng = np.random.default_rng()
    return 1-2*rng.random(size) + \
        1j-2j*rng.random(size)

The numpy random seed is a per-instance value, and can not be made persistent.

```
>>> np.random.default_rng().bit_generator.seed_seq
SeedSequence(
    entropy=75077652803687848143963328055419854028,
)
>>> np.random.default_rng().bit_generator.seed_seq
SeedSequence(
    entropy=210177805362268308102150041706209358051,
)
>>> np.random.default_rng().bit_generator.seed_seq
SeedSequence(
    entropy=83152795534019079780087961643608204450,
)
>>> np.random.default_rng().bit_generator.seed_seq
SeedSequence(
    entropy=312428692536571252558938966855008119863,
)
>>> np.random.default_rng().bit_generator.seed_seq
SeedSequence(
    entropy=144649769598635413415288828231534046565,
)
```
All random functions should optionally accept an instance of `np.random.default_rng()` to allow end-users to control the random seed. Also, perhaps `mathFunctions` should contain a global variable for specifying a global, non-thread safe random seed for simple use cases.

biergaizi:
## mathFunctions.py: API to set/get the global numpy RNG, close #1314

This commit introduces `rand_set_rng()` and `rand_rng()` for setting and getting the global numpy RNG used by scikit-rf. This is useful for fixing a random seed for reproducible Monte Carlo analysis and unit tests.

This function is expected to be called before using any scikit-rf features, since it's a global variable and thread-unsafe. To temporarily change the random number generator of a particular method (e.g. `skrf.media.random`), use the `rng` argument instead.

## networkSet.py: use the global RNG, with an optional override

This commit changes the `rand()` method to use the global RNG by default, with an optional override via the `rng` argument. The latter is useful for multi-threaded programs since `skrf.mathFunctions.rand_set_rng` is not thread-safe.

## media.py: optional RNG override for media.random()

This commit adds an optional argument `rng` to media.random(), so that the global numpy random number generator can be overridden. The latter is useful for multi-threaded programs since `skrf.mathFunctions.rand_set_rng` is not thread safe.

### DIFF
--- a/skrf/media/media.py
+++ b/skrf/media/media.py
@@ -1643,7 +1643,8 @@ class Media(ABC):
         return result
 
     def random(self, n_ports: int = 1, reciprocal: bool = False, matched: bool = False,
-               symmetric: bool = False, **kwargs) -> Network:
+               symmetric: bool = False, rng: None | np.random.Generator = None,
+               **kwargs) -> Network:
         r"""
         Complex random network.
 
@@ -1660,6 +1661,10 @@ class Media(ABC):
             makes s-matrix diagonal have single value ($S_{mm}=S_{nn}$)
         matched : bool
             makes diagonals of s-matrix zero
+        rng : :class:`numpy.random.Generator` or None
+            override the global :mod:`numpy` random number generator,
+            useful for multi-threaded programs since
+            :func:`skrf.mathFunctions.set_rand_rng` is not thread-safe.
 
         \*\*kwargs : passed to :class:`~skrf.network.Network`
                 initializer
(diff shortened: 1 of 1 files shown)


biergaizi:
Merge ready.

PR #1332, opened 2025-10-13
calibration.py: raise ValueError if 2-port inputs are used, close #1321.

### LINKED ISSUE #1321: Inconsistent Exception When Using Two-Port Networks as Inputs to One-Port Calibration Methods
In the current calibration test suite, it checks whether a `RuntimeError` is thrown when users accidentally use two-port networks as inputs to `OnePort`. However, other methods such as `SDDL`, `PHN` do not check their inputs nor they're tested, and they all threw different errors in the middle of a calibration run, sometimes it's `IndexError`, sometimes it's `ValueError`.

For consistency, all OnePort calibration methods should check the number of ports and throw an exception. I prefer `ValueError` instead of `RuntimeError` as the latter is a generic exception for unspecified problems.

biergaizi:
In the current calibration code, `OnePort` class checks whether users accidentally use 2-port input networks for 1-port calibration, and raises a `RuntimeError` if so. But there's no check for `SDDL`, `SDDLWeikle`, and `PHN` calibrations. As a result, they all threw different errors in the middle of a calibration run, sometimes it's `IndexError`, sometimes it's `ValueError`.

This commit adds a class-wide `self._check_input()` methods for all one-port calibration classes, which is called in `self.run()`. For consistency, we use `ValueError` instead of the original `RuntimeError`.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -1133,6 +1133,12 @@ class OnePort(Calibration):
         Calibration.__init__(self, measured, ideals,
                              *args, **kwargs)
 
+    def _check_input(self):
+        if not all([n.number_of_ports==1 for n in self.ideals]):
+            raise ValueError(f'ideals for {self.family} should be 1-port Networks')
+        if not all([n.number_of_ports==1 for n in self.measured]):
+            raise ValueError(f'measured networks for {self.family} should be 1-port Networks')
+
     def run(self):
         """ Run the calibration algorithm.
         """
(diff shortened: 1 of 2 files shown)


biergaizi:
Merge ready.

Please ignore the test failure, which is unrelated to the commit. The `TRLMultiline` test fails because of an I/O error while reading a `pickle` network, which looks like a race condition somewhere.

biergaizi:
I managed to clear the error by rerunning the CI/CD pipeline by force-pushing the same commit with a different hash via `--amend`.

PR #1335, opened 2025-10-14
doc/SOLT Calibration Standards Creation: language and formatting fixes.

biergaizi:
As it always happens, the correction of language and formatting issues is a limiting process - fewer and fewer issues are discovered per reread, but they never completely go to zero because the error curve is an asymptote.


biergaizi:
Merge ready. Hopefully no future change will be needed.

PR #1351, opened 2025-11-18
Hp8510c single point

jamiestepanian:
Modification to the HP 8510C VNA VI code to allow for single point sweeps.

I tested this on my system with success. The behavior of ramp sweeps is unchanged based on my tests.


jamiestepanian:
Are the code testing failures related to the cloudflare outages?
I see in the logs a "500 server error" accessing a web address.

jhillairet:
I don't know, but maybe. The tests didn't pass for a change in the docstring too, while passing locally. so I believe it's a GitHub side issue...

jhillairet (approved):

Issue #1358, opened 2025-12-04
Documentation bug in se2gmm


AlecVercruysse:
in the `se2gmm` documentation:

> For example, a 3-port single-ended network is converted to mixed-mode parameters:
> ```
> | Port 0 (single-ended, 50 ohms) --> Port 0 (single-ended, 50 ohms)
> | Port 1 (single-ended, 50 ohms) --> Port 1 (differential mode, 100 ohms)
> | Port 2 (single-ended, 50 ohms) --> Port 2 (common mode, 25 ohms)
> ```
> ```
> >>> ntwk.renumber([0,1,2], [2,1,0])
> >>> ntwk.se2gmm(p=1)
> >>> ntwk.renumber([2,1,0], [0,1,2])
> ```
> In the resulting network, port 0 is single-ended, port 1 is differential mode, and port 2 is common mode.

Following the example, however, after the network is converted to mixed mode, the port ordering is:
0: differential
1: common-mode
2: single-ended

The renumbering maps ports 2->0, 1->1, 0->2. So just the differential and single-ended ports are swapped. Therefore, I disagree with the final line that claims "port 1 is differential mode" (port 2 is differential mode).

A MWE:
```
import numpy as np
import skrf as rf

# fake 3-port
ntwk = rf.Network(
    f=[1],
    s=np.zeros((1, 3, 3), dtype=complex),
)
print(f"initial {ntwk.z0=}")

# do the reordering/conversion of the existing example
ntwk.renumber([0, 1, 2], [2, 1, 0])
ntwk.se2gmm(p=1)
ntwk.renumber([2, 1, 0], [0, 1, 2])

print(f"final {ntwk.z0=}")
```
Returns:
```
initial ntwk.z0=array([[50.+0.j, 50.+0.j, 50.+0.j]])
final ntwk.z0=array([[ 50.+0.j,  25.+0.j, 100.+0.j]])
```
The 100 ohm impedance port is port 2, indicating it is the differential mode port.

PR #1359, opened 2025-12-04
Update se2gmm example for N=3

AlecVercruysse:
Associated with issue #1358.

By redoing the ordering in the example:

```
import numpy as np
import skrf as rf

# fake 3-port
ntwk = rf.Network(
    f=[1],
    s=np.zeros((1, 3, 3), dtype=complex),
)
print(f"initial {ntwk.z0=}")

# do the reordering/conversion in the *new* example
ntwk.renumber([0, 1, 2], [2, 1, 0])
ntwk.se2gmm(p=1)
ntwk.renumber([0, 1, 2], [1, 2, 0])

print(f"final {ntwk.z0=}")
```
We see the differential port is correctly placed into port 1, as written in the description:
```
initial ntwk.z0=array([[50.+0.j, 50.+0.j, 50.+0.j]])
final ntwk.z0=array([[ 50.+0.j, 100.+0.j,  25.+0.j]])
```

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -3925,7 +3925,7 @@ class Network:
 
         >>> ntwk.renumber([0,1,2], [2,1,0])
         >>> ntwk.se2gmm(p=1)
-        >>> ntwk.renumber([2,1,0], [0,1,2])
+        >>> ntwk.renumber([0,1,2], [1,2,0])
 
         In the resulting network, port 0 is single-ended, port 1 is
         differential mode, and port 2 is common mode.


mhuser:
> should we change the documentation or the code?

My opinion is that the code cann be left unchanged. The first `p` pair of ports are converted to mixed mode, and then the others are left single-ended.

An improvement - that would be out of scope for this issue - would be to provide a way to renumber networks between the three most common numbering schemes.

This would make sense mostly for 4-ports or 2N-ports and not for 3-ports or (2N+1)-ports, however.

Issue #573, opened 2021-11-22
scikit-rf 2021 survey results


jhillairet:
Dear scikit-rf users,

We have made a short survey intended for all the users of the scikit-rf, to:
- better know the users and the usages,
- determine what are the functionalities used the most and the less,
- find the ones which could be improved,
- define new directions for developments.

Could you please fill this short (5 min or less) and anonymous survey by following the link below:

https://forms.gle/uonurVVb4FPhTLQm9

Thanks in advance for your time,

The results will be posted in this issue.

jhillairet:
![image](https://user-images.githubusercontent.com/4642848/143686123-38855799-1d64-4683-8d4c-c6e082c631f2.png)

![image](https://user-images.githubusercontent.com/4642848/143686128-967d53ec-179a-44b7-9484-6eeeb6e59ed4.png)

![image](https://user-images.githubusercontent.com/4642848/143686129-9b1d853a-c2a9-42d1-a1eb-36614d4a8788.png)

![image](https://user-images.githubusercontent.com/4642848/143686132-69d638c7-be45-484b-a267-474f3b0fb66a.png)

![image](https://user-images.githubusercontent.com/4642848/143686139-da85923a-b23c-4f0c-8931-248e6bb85d33.png)

![image](https://user-images.githubusercontent.com/4642848/143686146-5597b8d8-2e61-4acf-bfea-9a16f172358e.png)

![image](https://user-images.githubusercontent.com/4642848/143686148-4dea33d1-8548-45bd-b537-c9ed64ae3de5.png)

![image](https://user-images.githubusercontent.com/4642848/143686155-6a773ee6-439a-4b1f-859a-26f500d431a4.png)

![image](https://user-images.githubusercontent.com/4642848/143686159-90abc594-2d1b-41b9-9c72-ea15a039dc91.png)

![image](https://user-images.githubusercontent.com/4642848/143686162-04bd66a4-a68e-4cc0-bee7-315940dd96c1.png)

![image](https://user-images.githubusercontent.com/4642848/143686164-d8b4cab2-021c-48a1-ba69-92f41a6b83e4.png)

![image](https://user-images.githubusercontent.com/4642848/143686166-b7840fb5-ff9c-498f-85b4-12131139304f.png)

![image](https://user-images.githubusercontent.com/4642848/143686167-5444e894-3ee3-446d-a5fa-2adf48a8986b.png)

On improvements/doc :
Could you give some details on your previous answer?
[...]

PR #1365, opened 2026-01-26
Rework skrf namespace

luar123:
Follow up on https://github.com/scikit-rf/scikit-rf/pull/1065#issuecomment-3782278205

Currently everything is loaded into the skrf namespace leading to unexpected behavior like for example `import skrf.io` imports pythons io module and `import skrf.calibration.deembeding` rises an exception.

So here is a proposal for a cleaner structure:
1. Everything can be imported as expected, e.g.  `import skrf.calibration.deembeding`
2. All main modules and some modules from subpackages are available in skrf namespace, e.g. after `import skrf as rf`, `rf.network` or `rf.deembeding` are valid
3. Main classes with shorthands are still in the namespace: Network, NetworkSet, Circuit, Frequency
4. To avoid breaking existing code, everything can still be imported as before (through modified `__getattr__`) but will print a deprecation warning.
5. Tests are fixed (this is the biggest part of the PR)

Examples and notebooks are still tbd. For some reason those won't fail when running `pytest --nbval-lax`, but normal tests do.

Please let me know what you think and if you would like to change something.

### DIFF
--- a/skrf/calibration/tests/test_calibrationSet.py
+++ b/skrf/calibration/tests/test_calibrationSet.py
@@ -29,7 +29,7 @@ class DotOneport(unittest.TestCase,CalsetTest):
 
     """
     def setUp(self):
-        self.wg = rf.instances.RectangularWaveguide(rf.Frequency(75, 100, 11, unit='GHz'),
+        self.wg = rf.media.RectangularWaveguide(rf.Frequency(75, 100, 11, unit='GHz'),
                                           a = 100*rf.constants.mil, z0_override = 50)
         wg = self.wg
         self.n_ports = 1
@@ -63,7 +63,7 @@ class DotEightTerm(unittest.TestCase, CalsetTest):
     @suppress_warning_decorator("divide by zero encountered in true_divide")
     def setUp(self):
         self.n_ports = 2
-        self.wg = rf.instances.RectangularWaveguide(rf.Frequency(75, 100, 3, unit='GHz'),
+        self.wg = rf.media.RectangularWaveguide(rf.Frequency(75, 100, 3, unit='GHz'),
                                           a = 100*rf.constants.mil, z0_override = 50)
         wg= self.wg
         wg.frequency = rf.Frequency.from_f([100], unit='GHz')


luar123:
> Personally, I would prefer to either remove all shorthand imports in the future, or to keep shorthand imports for all modules ('io' then remains a problem).

I agree and both is fine for me. For now I just kept the four where we have these shorthands (N, F, C, NS), but to be honest I also don't like those because they are hiding what they really are.

Why is `io` a problem? You can import `skrf `and then access it with `skrf.io` or you can import it directly `from skrf as io` or `as rf_io` if you also need to `import io`. There is no io class, but we could import `skrf.io.touchstone.Touchstone` or `skrf.io.citi.Citi`, but it is difficult what to import and what not (I would not import `skrf.io.touchstone.ParserState` for example)

(Currently the issue with skrf.io is that skrf imports everything from network and network imports io, so the io from network is overwriting skrf.io)

Vinc0110:
Ok, now I see what's going on with 'io'. Then both ways would work.

Let's wait for other comments.

arsenovic:
i would just like to say i support  a change of this! 
most of namespace choices i made long ago where not good.

skrf/__init__.py
```diff
@@ -6,68 +6,72 @@
...
 
 ## Shorthand Names
 F = Frequency
 N = Network
 NS = NetworkSet
 C = Circuit
-lat = load_all_touchstones
+# lat = load_all_touchstones
 # saf  = save_all_figs
 saf = None
```
jhillairet: for saf, I wonder if it should be kept or not. Again, it's not documented, so I wonder if some users are using it...

jhillairet:
After merging this, since it may break compatibility with some code, should we bump a major version (2.0)?

skrf/__init__.py
luar123: Same as with lat I could remove it or move into `__getattr__()` with a warning. Same could be done with stylely. I plan to move the whole setup_plotting part into .plotting in the next PR when addressing the lazy_loading, but wanted to keep it here for now.

luar123:
> Personally, I would prefer to either remove all shorthand imports in the future, or to keep shorthand imports for all modules.

@jhillairet do you have an opinion here? Should I remove imports of the classes Network, NetworkSet, Circuit and Frequency? Or should I add other classes like Calibration, Deembedding, Media....

> After merging this, since it may break compatibility with some code, should we bump a major version (2.0)?

Good question, it should only break tests that fail on warnings, but of course I might have missed something in the `__getattr__` functions. Removing the getattr lookups should definitely be done in a major release.

I used FutureWarnings, should I switch to DeprecationWarnings? The latter are usually ignored and only show up in interactive mode or when explicitly enabled (e.g. in tests with `pytest -W error`).

Anyway, I will go ahead and fix the notebooks in the meantime.

arsenovic:
i defer to @jhillairet  on this, but in my opinion importing the most commonly used classes (`Network`, `Frequency`) to the root namespace makes for simpler imports for 90% of use cases.  the shorthand  notation (`N` `F`, etc) , i think the users should setup themselves.

jhillairet:
> the shorthand notation (`N` `F`, etc) , i think the users should setup themselves.

Actually, I'm not convinced that these shorthand notations are widely used, because they are not (much?) documented or do not appear in the examples. On the other hand, these are simple alias, so it should not be much of a concern for performance (?). 

> @jhillairet do you have an opinion here? Should I remove imports of the classes Network, NetworkSet, Circuit and Frequency? Or should I add other classes like Calibration, Deembedding, Media....

In order to not make a decision based on my tastes only, I suggest to look at the surveys we've made in [2024](https://github.com/scikit-rf/scikit-rf/discussions/1086) and [2022](https://github.com/scikit-rf/scikit-rf/issues/573). If we look to the question "what scikit-rf features have you ever used?", the first two are clearly `Network` and `Frequency`. (I should make a 2026 survey!)

Then, `NetworkSet`, Calibration, deembeding, `Media` or `Circuit` are the same ballpark.
 
So, base on this, I'd keep only `Network`, and `Frequency` in the "base" (and possibly their alias too). It should cover a high majority of use cases (with `stylely()`).

luar123:
Thanks for the feedback.
I removes Circuit, NetworkSet and all shorthands from main namespace, keeping only Network, Frequency and stylely.

Also fixed tutorials and examples.

skrf/calibration/tests/test_calibrationSet.py
```diff
@@ -62,10 +63,10 @@ class DotEightTerm(unittest.TestCase, CalsetTest):
     @suppress_warning_decorator("divide by zero encountered in true_divide")
     def setUp(self):
         self.n_ports = 2
-        self.wg = rf.RectangularWaveguide(rf.F(75, 100, 3, unit='GHz'),
-                                          a = 100*rf.mil, z0_override = 50)
+        self.wg = rf.instances.RectangularWaveguide(rf.Frequency(75, 100, 3, unit='GHz'),
```
jhillairet:
```suggestion
        self.wg = rf.media.RectangularWaveguide(rf.Frequency(75, 100, 3, unit='GHz'),
```

luar123:
> I admit the code is more clear. However it will break many users' codes. I propose to release a new minor version without this PR first, then release a major version after.

Sounds good. I will still leave the deprecation warnings in, right? So most code should not break.

jhillairet:
> > I admit the code is more clear. However it will break many users' codes. I propose to release a new minor version without this PR first, then release a major version after.
> 
> Sounds good. I will still leave the deprecation warnings in, right? So most code should not break.

Yes, leave it. We will eventually remove them after.

jhillairet (approved):

Vinc0110:
> Sounds good. I will still leave the deprecation warnings in, right? So most code should not break.

Exactly! We have the deprecations warnings in place, so old code can still run until we eventually decide to make the changes and remove the old convenience imports. I don't think this warning phase requires a major release or a bump to version 2.0. Maybe after the changes are finally done, we can move to a version 2.0?

jhillairet:
> > Sounds good. I will still leave the deprecation warnings in, right? So most code should not break.
> 
> Exactly! We have the deprecations warnings in place, so old code can still run until we eventually decide to make the changes and remove the old convenience imports. I don't think this warning phase requires a major release or a bump to version 2.0. Maybe after the changes are finally done, we can move to a version 2.0?

OK, eventually, yes (if we are sure that this would not break previous code in any case...)

luar123:
> OK, eventually, yes (if we are sure that this would not break previous code in any case...)

I double-checked and the warnings should cover everything. However, there is no way to test everything that was possible before, so there is a chance i missed something.

jhillairet:
ok, so let's merge it, for the moment considered as for a v1.x (with the warnings). A future version 2.0 without the warnings anymore (but finally not that close)

luar123:
Thank you

PR #1374, opened 2026-02-02
VectorFitting: Fixing division by zero

### LINKED ISSUE #1373: VectorFitting: Division by zero
In `VectorFitting.get_spurious()` there is a possibility for a division-by-zero error occurring in case of purely real-valued poles:
https://github.com/scikit-rf/scikit-rf/blob/c7110104878812222564761c67c49e1daa051c9e/skrf/vectorFitting.py#L155-L160

For real poles, we have `poles.imag == 0`, so both the list of norms and its mean value are zero.

Vinc0110:
Fixes #1373 and the docstring typo


Vinc0110:
Had to edit the testcase because the new code does change the results from `auto_fit()`.

PR #1389, opened 2026-03-14
Avoid throwing confusing warnings

jhillairet:
Addresses #1388 

avoid throwing confusing warnings about imported numpy/scipy functions such as `angle`, `zeros, etc...


skrf/__init__.py
```diff
@@ -74,6 +74,10 @@ def __getattr__(name: str):
         ]:
             result = getattr(module, name, None)
             if result is not None:
+                # Do not expose external (numpy, scipy) imported stuffs as deprecated.
+                mod = getattr(result, '__module__', '')
+                if mod and mod.startswith(('numpy', 'scipy')):
+                    return result
                 _warn(f"skrf.{name} is deprecated. Please import {name} from "
                      f"skrf.{module.__name__.split('.')[-1]} instead.", FutureWarning, stacklevel=2)
```
luar123: This check is a great solution!
I would still issue a warning because it won't work anymore if this look-up is removed, but pointing to the correct module. Maybe something like:

```suggestion
                # Do not expose external (numpy, scipy) imported stuffs as deprecated.
                mod = getattr(result, '__module__', '')
                if not (mod and mod.startswith(('numpy', 'scipy'))):
                    mod = f"skrf.{module.__name__.split('.')[-1]}"
_warn(f"skrf.{name} is deprecated. Please import {name} from "
f"{mod} instead.", FutureWarning, stacklevel=2)
```
(Did not test if this works)

PR #1390, opened 2026-03-23
Enhance the parsing ability for port names

SHF101202021:
Currently, port names are parsed by the patterns of `! port ....`, where the spacing is only allowed to occur one times between exclamation mark (`!`) and the letters (`port`).
Unfortunely, it is too strict for some touchstone files generated by non-Ansys tools.

Port names of the following cases cannot be correctly recognized:

1. More-than-one spacing used between `!` and `Port`   
  ```
  !  Port[1] = 1:1
  !  Port[2] = 2:1
  !  Port[3] = 3:1
  !  Port[4] = 4:1
  !  Port[5] = 5:1
  !  Port[6] = 6:1
  !  Port[7] = 7:1
  !  Port[8] = 8:1
  ```

2. Standard port names though there is no bracket `[` and `]`  for port indices.  
   For helic-type touchstone files, port names may belong to this case.

  ```
  ! Port 1 : P01
  ! Port 2 : P02
  ! Port 3 : P03
  ! Port 4 : P04
  ! Port 5 : P05
  ! Port 6 : P06
  ```

4. Sigrity-Style port names  
  This kind of styling is invented when touchstone file format is not complete enough.
  The port names are sequentially arragned from port 1 to port N, and the total name should be recognized as the port name
  For signal integrity & power integrtiy engineers, Sigrity EM tools are very common to use (like `PowerSI` or `Clarity`). 
  Hence, it should be considered.

  ```
  ! Port1_Bump::VDD25
  ! Port2_Bump::VDD125
  ! Port3_Bump::DATA1
  ! Port4_Bump::DATA2
  ! Port5_BGA::VDD25
  ! Port6_BGA::VDD125
  ! Port7_BGA::DATA1
  ! Port8_BGA::DATA2
  ``` 

Because the code structure is hard to interrup for me, I will use the method `_parse_port` to re-parse port names in my way when `state.port_names` is empty dictionary `{}` or port count (`len(state.port_names)`) doesnot match `state.rank`.


skrf/io/touchstone.py
```diff
@@ -354,6 +354,35 @@ def version(self, x: str) -> None:
         if x == "2.0":
             self._parse_dict.update(self._parse_dict_v2)
 
+    def _parse_port(self, fid: typing.TextIO) -> list[str]:
```
jhillairet, line 357: please add a bit of docstring for context and goal.

jhillairet (approved):

skrf/io/touchstone.py
SHF101202021: @jhillairet 
Done. Please take a look.

jhillairet (approved):

PR #1391, opened 2026-03-31
S2P Port Swapper

mad26:
implemented a GUI based interface to swap ports of a s2P file


FranzForstmayr:
Hi

Just my personal opinion here ... 
I'm not sure if this repository is the right place for standalone apps like this for several reasons. First it makes the library itself harder to maintain, because every change in the library could mean a rework in the application. Without any automated tests this would require manual work / testing, which is impossible in the long term. Another reason is, that this app won't get deployed if scikit-rf is installed via pip, which is the normal way. 

There are some apps in this repository already, but In my opinion all of them should be moved out into a separate repo. Each application could have it's own dependencies and requirements and can get properly packaged and deployed.

jhillairet:
Thank you, @mad26, for your PR

However, I agree with @FranzForstmayr. Standalone applications/GUI are hard to maintain because GUIs are very user/application dependent. This is the reason why the ones which existed before have been moved to another repo: https://github.com/scikit-rf/skrf-apps (which is not much maintained nor tested). I'm afraid we will not merge your PR in this repo, I hope you understand why, with the reasons provided.

mad26:
@jhillairet thanks, I understand. I put this application code in there as I am using it at my company a lot, so wanted to share with others, but yes I understand your concern.

Thanks

> Thank you, @mad26, for your PR
> 
> However, I agree with @FranzForstmayr. Standalone applications/GUI are hard to maintain because GUIs are very user/application dependent. This is the reason why the ones which existed before have been moved to another repo: https://github.com/scikit-rf/skrf-apps (which is not much maintained nor tested). I'm afraid we will not merge your PR in this repo, I hope you understand why, with the reasons provided.

PR #1392, opened 2026-04-04
Fix string parsing bug in n_twoports_2_nport (Closes #1217)

### LINKED ISSUE #1217: The function n_twoports_2_nport cannot accurately distinguish port numbers.
Thank you for reporting a bug! To help us understanding the issue and how to reproduce it, please fill the following sections as best as you can.

**Describe the bug**
When I was reading a large-dimensional S-parameter port matrix (e.g., 18 ports), I found that the data did not correspond to the generated S-parameter matrix ports. For example, S{14,18} was read as S{14,1}.

**To Reproduce**
If a S-parameter matrix with more than 10 ports is constructed, theoretically S{2,1} may be confused with S{2,10}.
Specifically, the number of ports I am using is 18, and in the results, S{14,18} is confused with S{14,1}.
> note that , the s2p subnetwork is named with 'S{i,j}', where i, j are port indexes.

**Expected behavior**
The output n-port network should correspond to each S2p file.

**System**
 - OS: Windows 10
 - scikit-rf version:1.3.0

**Here is one approach to fix this problem**
Using the re-expr, searching the port index in subntwk.name, like this: 
image

meslane:
Currently, `n_twoports_2_nport()` will fail to return a correct s-parameter matrix for networks with 11 or more ports due to an ambiguity in how the function handles network name strings. See: https://github.com/scikit-rf/scikit-rf/issues/1217 for details.

This PR fixes this behavior by using the python `re` module to explicitly search for integers in the network name string when a port separator is used. It also adds a unit test to add regression coverage for `n_twoports_2_nport()` when using networks with a large number of ports.

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -6452,9 +6452,21 @@ def n_twoports_2_nport(ntwk_list: Sequence[Network], nports: int,
     for subntwk in ntwk_list:
         for m, n in nport.port_tuples:
             if m != n and m > n:
-                if f"{m + offby}{port_sep}{n + offby}" in subntwk.name:
+                if port_sep == "":
+                    # If we have no separator, assume the first two digits we find are the port numbers
+                    re_digits = re.findall(r'\d', subntwk.name)
+                    p1 = int(re_digits[0])
+                    p2 = int(re_digits[1])
+                else:
+                    # If we have a separator, use split/regex match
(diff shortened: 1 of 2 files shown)


jhillairet (approved):
Thanks! LGTM.

meslane:
@jhillairet would you be able make #1217 an issue closed by this PR? I don't believe I have the permissions to do that. Thanks!

PR #1394, opened 2026-04-28
Implement Klopfenstein taper

PositroniumJS:


jhillairet (commented):
Thanks for the PR. Could you add a test for it?

By the way, which version is it ? The original paper's or the Kajfez and Prewitt's?

PositroniumJS:
> Thanks for the PR. Could you add a test for it?

Sure. I need to figure out where are the tests implemented but I should make it.

> By the way, which version is it ? The original paper's or the Kajfez and Prewitt's?

I followed the two references that are in the description of the class
https://github.com/scikit-rf/scikit-rf/blob/master/skrf/taper.py#L306

PositroniumJS:
I have implemented some tests for Tapers, it is not exhaustive by I think that it is honest work

jhillairet (approved):
Thank you!

PR #1401, opened 2026-05-29
update to TUGMultilineTRL

ZiadHatab:
I have updated some of the logic in my algorithm since last time I ported it to skrf. This PR brings most up-to-date implementation from my standalone repo, [ZiadHatab/multiline-trl-calibration](https://github.com/ZiadHatab/multiline-trl-calibration). The core algorithm didn't change, but I improved the robustness in solving gamma and how thru normalization is done (based on **[4]**) and added new parameters for scaling the eigenvalue problem related to line length (based on **[5]**)

## New references
- **[4]** Z. Hatab, M. E. Gadringer, W. Bösch, *"A Thru-Free Multiline Calibration,"* IEEE TIM, vol. 72, 2023, doi: 10.1109/TIM.2023.3308226.
- **[5]** Z. Hatab, M. E. Gadringer, W. Bösch, *"The Choice of Line Lengths in Multiline Thru-Reflect-Line Calibration,"* arXiv:2512.18641.

## Changes from [4]
- Thru normalization reworked to the S-parameter formulation of [4] (forces the calibrated thru to S21=S12=1), replacing the old T-parameter normalization. Implemented via a new `LFTinv` linear-fractional de-embedding and the 16-term error box. Original implementation only forced S21=1, while the effects of S11,S22 influenced S12 results (was minor, but now will match S21).

## Changes from [5]
- New constructor options **`compensate_repeated_lines=False`** and **`lnorm=1`** that scale the weighting matrix to down-weight repeated line lengths and to change the L-norm of the eigenvalue problem (appendix of [5]). Defaults are no-ops, so existing behavior is unchanged.
- New quality metrics from [5]: **`kappa`** (normalized eigenvalue) and **`effective_phase_deg`** = `arcsin(kappa/2)` in degrees (capped at 90°), indicating how well-conditioned the line set is.

## Propagation-constant robustness
The propagation constant is now estimated two ways; from the Takagi `G` matrix and from the de-embedded lines; and the solution that is most consistent with the known eigenvalue lambda is selected. This is more robust and avoids wrong sign selection for poorly conditioned line-length sets.

## Misc
- Multiple reflect standards are now combined through rank-1 (SVD) recovery. This is mathmaticaly optimal compared to a simple average approach.

## Tests
Refactored `TUGMultilineTest` to be parametrizable and added coverage for the previously untested paths, each reusing the full `EightTerm` accuracy suite where valid: non-zero-length thru, `lnorm=2`, repeated line length + `compensate_repeated_lines`, a no-reflect case (verifies gamma and S21/S12 correction), and check `lambd`/`kappa`/`effective_phase_deg`.

### DIFF
--- a/skrf/calibration/calibration.py
+++ b/skrf/calibration/calibration.py
@@ -3668,6 +3668,10 @@ class TUGMultilineTRL(EightTerm):
     The overall algorithm is based on [1]_, but the weighting matrix calculation is based on [2]_.
     You can read the mathematical details online at [3]_.
 
+    The thru normalization uses the S-parameter formulation from [4]_. Optionally, the weighting
+    matrix can be scaled to compensate for repeated line lengths and to change the L-norm weighting
+    of the eigenvalue problem, as described in [5]_ (see ``compensate_repeated_lines`` and ``lnorm``).
+
     The calibration reference plane is at the edges of the first line.
     By default, the reference impedance of the calibration is the characteristic impedance
     of the transmission lines. If the characteristic impedance is known,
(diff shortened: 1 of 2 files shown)


jhillairet (approved):

jhillairet:
@ZiadHatab , can you synchronize your PR with the master?

jhillairet:
Merged, despite one GH on py310 which takes an infinite amount of time

Issue #1395, opened 2026-04-29
Circuit diagrams would improve documentation


DanielSank:
For example, [ShuntAdmittance](https://scikit-rf.readthedocs.io/en/latest/api/generated/skrf.circuit.Circuit.ShuntAdmittance.html) would be clarified by adding something like this
```
(Port 1)-----(Port 2)
          |
         [Y]
          |
         Gnd
```

PR #1407, opened 2026-07-14
Document Circuit helper topologies

sapunyangkut:
## Summary

- add simple topology diagrams to the five `Circuit` network helpers: `Port`,
  `SeriesImpedance`, `ShuntAdmittance`, `Ground`, and `Open`
- render each diagram as a reStructuredText literal block so its alignment is
  preserved in the API documentation

Addresses #1395.

## Validation

- compiled `skrf/circuit.py` from source and parsed it with the standard-library
  AST parser
- asserted that all five helper docstrings contain the topology marker and their
  exact diagram text
- `git diff --check`

The original `ShuntAdmittance` fragment was built with Sphinx 9.1.0 and its
rendered alignment was verified. Sphinx, Ruff, and the project runtime
dependencies were unavailable when validating the follow-up diagrams locally,
so those checks were not rerun for the expanded patch; the new literal blocks
use the same syntax as the rendered fragment.

## AI assistance

OpenAI Codex was used to investigate the issue, add the documentation blocks,
and run local validation. The final diff and validation evidence were reviewed
before submission.

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -599,6 +599,14 @@ class Circuit:
         """
         Return a 2-port network of a shunt admittance.
 
+        The network topology is::
+
+            (Port 1)-----(Port 2)
+                      |
+                     [Y]
+                      |
+                     Gnd
+
         Passing the frequency and name is mandatory.
 
         Parameters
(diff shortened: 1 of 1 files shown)


jhillairet:
Thanks, but ideally should be done for all the others too.

jhillairet (approved):

PR #1409, opened 2026-07-18
add surface_impedance() for rough and coated conductors

ZiadHatab:
This PR adds a new function, surface_impedance(), to skrf.tlineFunctions. It computes the surface impedance of a rough conductor boundary, and more generally of a stack of coated conductors (e.g. plating finishes).

Surface impedance is the best way to describe conductive boundaries, especially for EM simulation, but also for analytical models of transmission lines. The purpose of introducing the function is that I could use it later for the Media class in the hope of updating the analytical models there to include a surface impedance definition and allow users to specify it however they want (a future PR).

The implementation is ported from my standalone repository [ZiadHatab/rough-surface-impedance](https://github.com/ZiadHatab/rough-surface-impedance), trimmed to the surface-impedance part only.

Each boundary is described by a cumulative distribution function (CDF) of its surface height, which grades the material properties across the transition. The graded profile is split into short transmission-line segments and the input impedance is transformed from the deepest material up to the surface. The approach follows the papers:

- B. Tegowski et al., A Transmission Line Approach for Rough Conductor Surface Impedance Analysis, IEEE TMTT, 2023.
- G. Gold and K. Helmreich, Modeling of transmission lines with multiple coated conductors, EuMC, 2016.
- S. Blanes et al., The Magnus expansion and some of its applications, Phys. Rep., 2009.

Only one new public function is added (surface_impedance). The helpers (_roughness_cdf, _parse_material_properties, _segment_edges) remain private. No new dependencies are introduced; only numpy and scipy, both already required for skrf.

A use case example is mentioned in the docstring. I might include a tutorial or full example later when I update Media for some of the transmission line models.

I included couple of unit tests under SurfaceImpedanceTests in the file test_tlineFunctions.py

### DIFF
--- a/skrf/tests/test_tlineFunctions.py
+++ b/skrf/tests/test_tlineFunctions.py
@@ -2,7 +2,8 @@ import unittest
 
 import numpy as np
 from numpy import array, imag, linspace, pi, real
-from numpy.testing import assert_almost_equal, assert_equal
+from numpy.testing import assert_allclose, assert_almost_equal, assert_equal
+from scipy.constants import epsilon_0, mu_0
 
 import skrf.tlineFunctions as tlFuncs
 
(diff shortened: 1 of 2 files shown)


mhuser (approved):
Looks good. Great to have fresh stuffs about the transmission lines!

jhillairet (approved):

jhillairet:
Thanks for the addition! Looks promising with future changes suggested in Media.

A comment on the function name: shouldn't be more specific (adding tough conductor in the function name for example), in case in the future one would allow different physical model/ material /topology or directly setup it's own Zs values itself?

ZiadHatab:
I think of this as a very general tool that can compute the surface impedance of any stacked material definition with arbitrary properties. It is not only for rough boundaries. The `rms_roughness` can be zero, so it may be better to default `rms_roughness=0`. That way, the function returns the same result as `surface_resistivity()` for the same material properties. If you prefer that behavior, let me know. I can create another PR with that edit.

jhillairet (approved):
Ok, fine. Let's leave as it is for the moment.

PR #1410, opened 2026-07-20
Fix typing import

Ttl:
I need this on my machine or otherwise I get `AttributeError: module 'numpy' has no attribute 'typing'` on `import skrf`. Python 3.10 and numpy version 1.26.4.


luar123:
skrf dependency is `numpy >= 1.21` but no tests are running against `numpy < 2.x`. Maybe set the oldest test environment (3.10) explicitly to minimum versions of numpy/scipy/pandas? There could be more issues...

jhillairet (approved):

Issue #1241, opened 2025-01-17
port_names is not handled correctly in many functions


Ttl:
`Network` class has `port_names` property that assigns a name for each port. This seems to be mainly used for touchstone import and export. It would be useful when connecting N-port networks to keep track of which port is where but many current operations don't support it.

`renumber` and `subnetwork` seem to be the two operations that keep track of it currently.

`Circuit` doesn't populate names when generating a `Network` even though `Port`s should be named. `connect` doesn't seem to handle it correctly and it gives incorrect length for `port_names` when connecting two N-port networks. `se2gmm` should do something with it. There are probably other functions.

mhuser:
Related to #1257 for `connect`.

PR #1411, opened 2026-07-22
Fix incorrect z0 and port_names order in 2-port connect

Ttl:
`connect(ntwkA, 0, ntwkB, 0)` or `connect(ntwkA, 0, ntwkB, 1)` for `ntwkA` and `ntwkB` both 2-ports with different port impedances would result in swapped port impedances and port names of the output network. This bug does not apply to cascade (`**` operator).

### DIFF
--- a/skrf/tests/test_network.py
+++ b/skrf/tests/test_network.py
@@ -595,6 +595,15 @@ class NetworkTestCase(unittest.TestCase):
         np.testing.assert_allclose(c.z0, np.array([[25, 50]]))
         self.assertEqual(c.port_names, ['25ohm', '50ohm'])
 
+        xformer_flip = xformer.copy()
+        xformer_flip.flip()
+
+        # Test that cascade and connect agree
+        c2 = xformer_flip ** xformer_flip
+        np.testing.assert_allclose(c.s, c2.s)
+        np.testing.assert_allclose(c.z0, c2.z0)
+        self.assertEqual(c.port_names, c2.port_names)
+
     def test_connect_nport_2port(self):
         freq = rf.Frequency(1, 10, npoints=10, unit='GHz')
 
--- a/skrf/tests/test_network.py
+++ b/skrf/tests/test_network.py
@@ -587,8 +587,13 @@ class NetworkTestCase(unittest.TestCase):
         xformer.frequency=rf.Frequency(1, 1, 1, unit='GHz')
         xformer.s = ((0,1),(1,0))  # connects thru
         xformer.z0 = (50,25)  # transforms 50 ohm to 25 ohm
+        xformer.port_names = ['50ohm', '25ohm']
         c = connect(xformer,0,xformer,1)  # connect 50 ohm port to 25 ohm port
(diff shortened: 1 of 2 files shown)


skrf/tests/test_network.py
```diff
@@ -609,13 +614,85 @@ def test_connect_nport_2port(self):
...
+                              s=self.rng.random((1, nports, nports)))
+            ntwk.port_names = [f'{prefix}{i}' for i in range(nports)]
+            return ntwk
+
+        ntwk_a, ntwk_b = nport(3, 'a', 50), nport(3, 'b', 50)
+        ntwk_c = connect(ntwk_a, 1, ntwk_b, 0)
+        self.assertIsInstance(ntwk_c.port_names, list)
+        self.assertEqual(ntwk_c.port_names, ['a0', 'a2', 'b1', 'b2'])
+        # Name based indexing must work on the result
+        ntwk_c['a0', 'b2']
```
jhillairet, line 693: is this a remaining test?

skrf/tests/test_network.py
```diff
@@ -587,8 +587,13 @@ def test_connect(self):
         xformer.frequency=rf.Frequency(1, 1, 1, unit='GHz')
         xformer.s = ((0,1),(1,0))  # connects thru
         xformer.z0 = (50,25)  # transforms 50 ohm to 25 ohm
+        xformer.port_names = ['50ohm', '25ohm']
         c = connect(xformer,0,xformer,1)  # connect 50 ohm port to 25 ohm port
         self.assertTrue(np.all(np.abs(c.s-rf.network.impedance_mismatch(50, 25)) < 1e-6))
+        # the surviving ports are the 25 ohm port of the first xformer and the
+        # 50 ohm port of the second one, in that order.
+        np.testing.assert_allclose(c.z0, np.array([[25, 50]]))
+        self.assertEqual(c.port_names, ['25ohm', '50ohm'])
```
jhillairet, line 596: I don't understand why this non-intuitive result is desirable..

Ttl: Transformer port 1 impedance is 50 ohms and port 2 is 25 ohms. When two transformers are connected together such that port 1 of the first one is connected to port 2 of the second one it will connect 50-ohm port on the first one to 25-ohm port on the second one: [25, 50] <-> [25, 50] = [25, 50]. The first transformers 25-ohm port 2 is the first port in the resulting network. Note that in `connect_s` function 2-ports special cased to return in this order.

skrf/tests/test_network.py
Ttl: This is indexing with `port_names`. There are lot of issues with them currently (#1241), but this is existing functionality and this line is just making sure it doesn't raise exception.

Ttl:
I added a new small test to check that connect and cascade agree. This wouldn't pass without this PR.

skrf/tests/test_network.py
jhillairet: ok, more clear

jhillairet (approved):

PR #1414, opened 2026-07-26
Port name fixes

### LINKED ISSUE #1241: port_names is not handled correctly in many functions

Ttl:
Fixes #1241.

Fix many bugs with `port_names` handling:
* `port_names` is now a property with some basic error checking on assignment.
* Circuit now has `port_names` in generated Networks.
* `port_names` can be gives in `Network` constructor.
* `connect`, `copy_from`, `stitch`, `inv`, `concat_ports`, `twoport_reflect`, `one_port_2_two_port` functions now handle `port_names`.
* `se2gmm` will automatically generate name for ports and it can undo the automatically generated names when transforming in the opposite direction.

I have few more changes, for example accepting port names in connect, but I split them into separate PR.

### DIFF
--- a/skrf/circuit.py
+++ b/skrf/circuit.py
@@ -1276,6 +1276,27 @@ class Circuit:
                 port_indexes.append(idx_cnx)
         return port_indexes
 
+    @property
+    def port_names(self) -> list[str]:
+        """
+        Return the names of the "external" ports.
+
+        The names are given in the same order as the ports of the Network
+        returned by :func:`network`, ie. in the order the port Networks appear
+        in the connection list.
+
+        Returns
+        -------
+        port_names : list of str
+
+        See Also
+        --------
+        port_indexes
+        network
+        """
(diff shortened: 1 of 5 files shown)


jhillairet (approved):
Thanks @Ttl , great addition. 
Could you add some examples using the port name indexing in the Network tutorial too?

Ttl:
I will add examples in a separate PR.

PR #1415, opened 2026-07-27
Read HFSS NxN Gamma and port impedance comments

gaoflow:
`skrf/io/tests/ansys_terminal_data.s4p` states `51+1j 52+2j 53+3j 54+4j` in its own header, but `rf.Network(...).z0` returns `[51+1j, 0, 0, 0]`; on the HFSS 2025.2 8-port file posted in #354 seven of the eight terminals come back as 0 Ω, and port 2's gamma silently picks up an off-diagonal coupling term instead of its own propagation constant. The `! Gamma` and `! Port Impedance` handlers ask `_parse_n_floats` for exactly `2*rank` values, so in Driven Terminal mode only the first row of the NxN matrix is consumed and the remaining rows fall through to the generic comment handler.

`test_ansys_terminal_data` and a fix for this were both written in df205c0e and disabled again in 25ea92f5, "Do not test ansys NxN port impedance matrix files for now", because keying the layout off the `! Terminal data exported` marker does not work — enabling those two commented-out hunks verbatim makes 21 of the `HFSS_2020R2` fixtures plus `circuit_vi_Coaxial_T.s3p` fail to parse at all, since 22 of the 25 terminal-marked fixtures in the repo are in fact 1xN. This reads the whole comment block instead and compares its length against `2*rank` and `2*rank**2`, so the layout comes from the data and the marker is not consulted; across all 11832 such blocks in the repo's 295 Touchstone files there is none whose length is neither, and `rank == 1` is the only case where the two coincide (harmlessly, since a 1x1 matrix is its own diagonal).

Parsing every Touchstone file in the tree before and after, 309 of 311 are identical in f/s/z0/gamma/comments/s_def/port names, and the two that change are the NxN files, which now match the diagonals their own headers declare. The suite goes 1675 → 1679 passed and 46 → 45 skipped, with the pre-existing `test_se2gmm_renorm` failure unchanged. The expected values in the un-skipped test are Franz's from df205c0e, so I only removed the skip; the new 2-port fixture is the real 2020R1 export quoted in #354 and is the first coverage of the NxN *gamma* path, which had no pre-written fix.


jhillairet (approved):
LGTM, thanks @gaoflow

PR #1416, opened 2026-07-29
Fix time gate level shift caused by a periodic window

### LINKED ISSUE #1399: Time gating using convolution in 1.9 gives DC offset
Thank you for reporting a bug! To help us understanding the issue and how to reproduce it, please fill the following sections as best as you can.

**Describe the bug**
Comparing version 0.23.1 and 1.9.1 the results using the convolution method are different. The gate has been made wider in 1.9.1 by 1 see lines 435 and 440 in time.py. This addition of 1 gives  a small DC offset in the time gated result which is incorrect.  

**To Reproduce**
Run 1.9.1 with and without the 1 in lines 435 and 440 using the convolution method

**Expected behavior**
0.23.1 and 1.9.1 to give the same results

**Data**
The .s2p file is attached as .txt  

[Gold_wgc_ac_01.txt](https://github.com/user-attachments/files/27706938/Gold_wgc_ac_01.txt)

**Screenshots**
Here is a plot from my test harness.  I have local copies on time_gate installed from 0.23.1(3.6) and 1.91(1.9.1)

Image

The orange dotted  line is from scikit-rf 1.9.1 without modification.

**System**
 - OS: Windows 11
 - scikit-rf 1.9.1

**Additional context**
I have been updating the MATLAB code which ESA use on their millimetre wave scatterometer from 2016B, to 2020b and eventually to 2024b.  Each of these versions uses a different version on python 3.6, 3.8 and 3.12 which in turn loads a different version of sckit-rf 0.23.1, 1.9.1 and ?(Need to check). They noticed a small DC shift in the time gated output when going from 0.23.1 to 1.9.1 across many different data sets.

DanielMBouyou:
## Summary

- Build the discrete time-domain gate from a symmetric SciPy window.
- Keep the inclusive gate support introduced in #900.
- Add regression coverage for level preservation and closed gate support.

## Context

`scipy.signal.get_window()` creates a periodic window by default. For an odd
window length, that window is centered about `Nx / 2` rather than
`(Nx - 1) / 2`.

`time_gate()` places the window on a discrete time grid containing a center
sample at `t = 0`. The periodic window is therefore shifted by half a sample,
and its value at the center remains below unity.

The zero-lag value of the frequency-domain convolution kernel equals the gate
value at `t = 0`. A flat frequency response is consequently scaled by this
value. On the dataset attached to #1399, the current implementation produces
a level shift of approximately -0.0172 dB. The effect becomes larger for
narrower gates.

Before #900, a half-open slice happened to compensate for the periodic
window's offset, but its support was asymmetric and one sample too short.
This change keeps the inclusive support and instead builds the gate from a
symmetric window.

## Implementation

`get_window()` now accepts an explicit `fftbins` argument.

String and tuple window specifications pass this option to SciPy. Callable
windows keep their existing contract and are still called with only the
requested window length. This is why the option is consumed by the helper
rather than passed through `**kwargs` at the `time_gate()` call site.

Only the time-domain gate uses `fftbins=False`. Frequency-domain tapering and
`Network.windowed()` retain their existing behavior.

## Verification

- `pytest -v -c "" skrf/tests/test_network.py -k "time_gate"`
- `pytest -v -c "" skrf/tests/test_network.py`
- `ruff check skrf/time.py skrf/tests/test_network.py`
- `git diff --check`

The full test suite previously completed with 1665 passed, 53 skipped,
4 xfailed, and one Windows-specific `NamedTemporaryFile` permission failure in
`test_networkSet.py::test_to_mdif`. The same failure was reproduced on the
unmodified `upstream/master` branch.

FFT and circular-convolution gating remain equivalent within floating-point
precision.

## Scope

This change does not modify:

- the inclusive gate width;
- `fft_window`;
- `Network.windowed()`;
- convolution boundary handling;
- the convolution kernel definition.

## Limits

The requested span remains quantized to the available time grid. Callable
windows remain responsible for returning an appropriately symmetric window.

Fixes #1399.

### DIFF
--- a/skrf/tests/test_network.py
+++ b/skrf/tests/test_network.py
@@ -167,10 +167,70 @@ class NetworkTestCase(unittest.TestCase):
         for window in ["hamming", ('kaiser', 6)]:
             gated1 = self.ntwk1.s11.time_gate(0,.2, t_unit='ns', window=window, fft_window=window)
 
-            get_window = partial(signal.get_window, window)
+            # the time-domain gate is symmetric, the frequency-domain window is periodic
+            gate_window = partial(signal.get_window, window, fftbins=False)
+            fft_window = partial(signal.get_window, window)
 
-            gated2 = self.ntwk1.s11.time_gate(0,.2, t_unit='ns', window=get_window, fft_window=get_window)
+            gated2 = self.ntwk1.s11.time_gate(0,.2, t_unit='ns', window=gate_window, fft_window=fft_window)
             assert gated1 == gated2
+
+    def test_time_gate_preserves_flat_response(self):
+        """A flat response gated around t=0 must come back unchanged.
+
+        The gate is a symmetric window whose peak sample sits on t = 0, so the zero-lag term
+        of the frequency-domain kernel is exactly one and a constant reflection coefficient is
+        passed through untouched. A periodic window is symmetric about sample `window_width / 2`
+        instead, which puts the peak half a sample late and scales the gated data by a constant
+        slightly below one.
+        """
+        s0 = 0.9 - 0.3j
+
+        # Sweep an even and an odd number of frequency points and an even and an odd
+        # half-width k. The resulting gate width 2 * k + 1 is always odd, which is
+        # intended: only an odd window has a single center sample on which
+        # gate(t = 0) = 1 can be imposed, so the invariant does not apply to even widths.
+        for npoints in (256, 257):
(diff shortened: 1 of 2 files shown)


Vinc0110 (approved):
This is a nice solution to implement the `fft_bins=False` argument for `time_gate()`. It's also a good idea to add those test cases. All very good in my opinion. Thank you for this PR!

jhillairet:
thanks @DanielMBouyou and welcome as a contributor !

PR #1417, opened 2026-07-30
Add circular waveguide attenuation

asirinelli:
Based on Balanis' Advanced Engineering Electromagnetics book, the alpha_c function for the class CircularWaveguide is implemented for all TE and TM modes.

A simple example is provided which replicates the 2 figures used for illustration in the reference book.


jhillairet (commented):
Thanks @asirinelli , great addition. I wonder if you can get rid of the division per zero warnings (seen in the notebook) ?

(Please commit the notebook without the cell results so save space on the repository)

asirinelli:
Yes, the division per zero is coming from my example and a simplistic and common definition of the frequency range for all the modes. I will fix that. I will also commit the notebook without cell results.

asirinelli:
I have improved the example as requested. The notebook is also committed without the cell results.

jhillairet (approved):

PR #1418, opened 2026-07-30
More port name fixes

Ttl:
* Accept port name in `connect`, `parallelconnect`, `renumber`, and `subnetwork`. If there are duplicate port names exception is raised when trying to connect those ports.
* `Network.delay` keeps the port name.
* Added a section about port names in Networks tutorial.

This should be all of the port name related fixes for now, but there are so many places that do weird things with networks that I suspect there are still some broken places somewhere with less used functions.

`nonreciprocity` uses one based indexing instead of zero based indexing used elsewhere. This was there before, but I added a note about it in the docstring. Passing zero as port number there doesn't raise an error and instead indexes the last port. This should be probably fixed in `__getattr__`, but I did not include that in this PR.

### DIFF
--- a/doc/source/tutorials/Networks.ipynb
+++ b/doc/source/tutorials/Networks.ipynb
@@ -737,6 +737,284 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Port names\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


jhillairet (approved):
This is very cool, thank you @Ttl ! Documentation is very complete too!

PR #1419, opened 2026-08-03
New implementation for coaxial media

ZiadHatab:
I updated the coaxial media to a newer analytical model based on

S. A. Schelkunoff, "The electromagnetic theory of coaxial transmission lines and cylindrical shields," The Bell System Technical Journal, vol. 13, no. 4, pp. 532-579, Oct. 1934, doi: [10.1002/j.1538-7305.1934.tb00679.x](https://doi.org/10.1002/j.1538-7305.1934.tb00679.x)

I also included a simpler, faster model that you can switch to, based on

F. M. Tesche, "A Simple Model for the Line Parameters of a Lossy Coaxial Cable Filled With a Nondispersive Dielectric," IEEE Transactions on Electromagnetic Compatibility, vol. 49, no. 1, pp. 12-17, Feb. 2007, doi: [10.1109/TEMC.2006.888185](https://doi.org/10.1109/TEMC.2006.888185)

Both models were extended to include a surface impedance description, so they can account for conductor roughness and coating. The dielectric can also be a magnetic material.

The original API is unchanged. I added the new parameters `tout`, `simple_model`, `dielectric`, `inner_conductor`, and `outer_conductor`. The latter three are dictionaries that let the user describe the materials however they want.

I included a simple usage example in the docstring and updated the unit tests.

While at it, the first commit was to clean up the docstring of surface_impedance() function and set rms_roughness to default to zero.


skrf/media/coaxial.py
```diff
@@ -66,18 +67,82 @@ class Coaxial(DistributedCircuit, Media):
     sigma : number, or array-like, optional.
         conductors electrical conductivity, in S/m.
         Default is infinity (non lossy metal).
-
-
-    TODO : different conductivity in case of different conductor kind
+    tout : number, or array-like, optional
+        wall thickness of the outer conductor, in m.
+        Default is infinity, i.e. a shield much thicker than the skin depth.
+    simple_model : bool, optional
```
jhillairet: Does it worth making this as a string instead? In case that in the future a third model be implemented (?)

skrf/media/coaxial.py
```diff
@@ -66,18 +67,82 @@ class Coaxial(DistributedCircuit, Media):
...
+    inner_conductor : dict, list of dict, or None, optional
+        Material of the inner conductor. A single dict describes a bulk conductor,
+        a list of dict a stack of coatings ordered from the dielectric inwards, of
+        which the deepest layer is the bulk that carries the dc current. Each dict
+        takes the keys ``'sigma'``, ``'mu_r'`` and ``'ep_r'`` of the layer itself,
+        and ``'rms_roughness'``, ``'boundary_loc'`` and ``'distribution'`` of the
+        boundary on top of it, which are handed to
+        :func:`~skrf.tlineFunctions.surface_impedance`.
+        If None (default), the conductor is smooth and made of bulk `sigma`.
+    outer_conductor : dict, list of dict, or None, optional
```
jhillairet, line 92: ok, what if someone only give either the inner or the outer but not both?

jhillairet (approved):
Great addition @ZiadHatab , thanks ! Just a comment or two concerning the argument type.

skrf/media/coaxial.py
ZiadHatab: I agree. I will change it to string.

skrf/media/coaxial.py
ZiadHatab: By default, the conductors are defined by variable sigma. If the user doesn't provide any of the new parameters, it uses whatever sigma is (same old behavior). The same applies if you define inner explicitly but not outer. In that case, outer will default to whatever sigma is.

What would be better? Here are some options:
1. keep as it is. If user defines one, the other defaults to whatever default when both not defined, i.e., sigma.
2. Detect if inner or outer defined and duplicate to the one not defined.

ZiadHatab: Also, would it make more sense to default `sigma=58e6` (i.e., copper) instead of INF ?

jhillairet: I don't have a preference. Should a warno6be issued in such case to avoid setting a value implicitly?

jhillairet: - A warning -

ZiadHatab:
- `sigma` now defaults to copper (58e6 S/m) instead of infinity.
- `tout` now defaults to `None` instead of the constant `INF`. Internally, it defaults to `np.inf` when detects `None`. In general, `tout` can handle large values.
- Dropped `skrf.constants.INF` from the module; plain `np.inf` works for both sigma and tout.
- `epsilon_r` and `tan_delta` now report the filling actually in use when `dielectric=` is given, instead of their unused constructor defaults.
- New warning when a material dict is given to one conductor but not the other.
- `Rs` (Surface resistivity) property documented as unused, kept for backward compatibility.
- Both conductor models rewritten so one expression covers every wall thickness including infinite. The special cased thick branches are gone, and `tesche` model no longer overflows for a very thick wall.


For unit tests: `INF` removed, thick-wall convergence checked against the closed-form limit, perfect-conductor test extended to a large finite conductivity.

skrf/media/coaxial.py
ZiadHatab: done!

jhillairet (approved):
Great work!

ZiadHatab:
Waveguide Media is next PR.

PR #1420, opened 2026-08-04
Check port indices in Network.__getattr__

Ttl:
Raise exception on zero index port or port that exceeds the number of ports on `Network.__getattr__`. For example `Network.s00` used to return the last port but after this PR it raises instead. Same with `nonreciprocity` that has 1-based indices, 0 index doesn't wrap around silently anymore.


jhillairet (approved):
Wow, nice catch, I was not aware of this.

PR #1421, opened 2026-08-09
New implementation for waveguide media

ZiadHatab:
I rewrote a large part of the code, while making sure the existing API is not impacted.

- New parameters to describe the materials: `dielectric`, `wall_a` and `wall_b`, plus a new parameter `model` to switch between loss models.

- A new model based on the two-wire approach is now the default. It is more accurate  and physical, since wall loss shapes the phase velocity as well as the attenuation, which is what measurement and EM simulation show. The implementation follows the paper by Lomakin, which is derived for TE10 (dominant mode):
  - K. Lomakin, G. Gold and K. Helmreich, "Analytical Waveguide Model Precisely Predicting Loss and Delay Including Surface Roughness," IEEE Trans. Microw. Theory Techn., vol. 66, no. 6, pp. 2649-2662, June 2018: https://doi.org/10.1109/TMTT.2018.2827383

- Roughness is now modeled by a surface impedance `skrf.tlineFunctions.surface_impedance()`, which lets the user specify not only an rms roughness but also plating. A wall is a single dict for a bulk conductor, or a list of dicts for a stack of coatings ordered from the filling inwards, each layer carrying its own `sigma`, `mu_r` and `ep_r`, and the roughness of the boundary on top of it. The two wall pairs are described separately and can differ.

- Cleaned up the original power loss model and updated its references to Marcuvitz's book, as well as to Skinner's paper, which documents this model as the one adopted by IEC 60153-2 and IEEE 1785.1. Note that this model only modifies the loss (alpha) and keeps the phase constant (beta) at its ideal value, which is not accurate when the walls are rough or slotted. It is kept available as a reference. Modes that the two-wire model does not cover, i.e. TM modes and TE modes with both indices nonzero, also fall back to Marcuvitz's model, which is essentially the ideal waveguide response. Defining a loss per unit length is not straightforward for those modes, since they do not exist alone. This was the behavior of the code before I edited it, so I kept it, and a warning is given whenever the fallback happens.
  - N. Marcuvitz, "Transmission-line modes," in Waveguide Handbook, 2nd ed. New York: McGraw-Hill, 1951, Chapter 2.
  - J. Skinner, D. Koller, H.-U. Nickel, N. M. Ridler and S. Lucyszyn, "Derivation of Rectangular Metallic Waveguide Attenuation Constant for IEC 60153-2 and IEEE 1785.1 International Standards," IEEE Trans. THz Sci. Technol., vol. 15, no. 4, pp. 734-737, July 2025. https://doi.org/10.1109/TTHZ.2025.3573847

- The existing material parameters `ep_r`, `mu_r`, `rho` and `roughness` are used by default for anything the user does not describe directly with the new parameters `dielectric`, `wall_a` and `wall_b`. Leaving all three unset reproduces the old behavior. If, for example, one wall pair is defined but not the other, a warning is given and the undefined pair falls back to `rho` and `roughness`.

- Default `rho` is now copper, i.e. 1/58e6 ohm*m. You can pass `rho=0` or `rho=None` for perfectly conducting walls.

- Added docstring examples showing how to use the new parameters to define the materials.

- Unit tests updated and new ones added.

- Since I changed `rho` to copper, I had to update one line in `test_calibration.py` that was expecting `rho=0`, so I forced the lossless value there.

### DIFF
--- a/skrf/calibration/tests/test_calibration.py
+++ b/skrf/calibration/tests/test_calibration.py
@@ -49,7 +49,7 @@ NPTS = 1
 # WR10/WG27/R900 75 to 110 GHz, 0.1x0.05 inch (2.54x1.27 mm)
 # z0 from 610 to 446 ohm
 WG_lossless = RectangularWaveguide(rf.Frequency(75, 100, NPTS, unit='GHz'),
-                                      a=100*mil, z0_override=50)
+                                      a=100*mil, rho=0, z0_override=50)
 WG = RectangularWaveguide(rf.Frequency(75, 100, NPTS, unit='GHz'), a=100*mil,
                                       rho='gold', z0_override=50)
 
--- a/skrf/media/rectangularWaveguide.py
+++ b/skrf/media/rectangularWaveguide.py
@@ -26,16 +26,15 @@ Characteristic Impedance              :math:`z_0`    :attr:`z0`
 """
 from __future__ import annotations
 
+import warnings
 from numbers import Number
 from typing import TYPE_CHECKING
 
 import numpy as np
-from numpy import exp, pi, sqrt, where
 
 from .. import constants as _const
 from ..data import materials
-from ..tlineFunctions import skin_depth
-from .freespace import Freespace
+from ..tlineFunctions import surface_impedance
 from .media import Media
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


jhillairet (approved):
Great addition! Thanks @ZiadHatab

Issue #1423, opened 2026-08-14
surface_impedance() returns negative real part (non-passive) when rough boundary transitions overlap


evanmarshall:
**Describe the bug**
`skrf.tlineFunctions.surface_impedance()` can return a surface impedance with a negative real part when the rough transitions of two boundaries in a coated-conductor stack overlap, even though every material in the stack is passive. A negative real surface impedance means the model says the metal adds energy to the wave instead of absorbing it, which is physically impossible. Through the `Coaxial` media this propagates into a negative distributed resistance `R`, so a cable built from ordinary passive materials comes out non-passive. No warning is raised; the result looks like a normal finite number.

The graded material profile is built by summing each boundary's CDF contribution independently:

https://github.com/scikit-rf/scikit-rf/blob/34652721a23bb87260916f2cbf063b92f38ab8eb/skrf/tlineFunctions.py#L382-L385

```python
cdf = np.array([_roughness_cdf(x_gauss, r, loc, dist)
                for r, loc, dist in zip(rms_roughness, boundary_loc, distribution)])
mu_gauss = _const.mu_0*(mu_r[0][:, None] + np.einsum('kf,kn->fn', np.diff(mu_r, axis=0), cdf))
ep_gauss = _const.epsilon_0*(ep_r[0][:, None] + np.einsum('kf,kn->fn', np.diff(ep_r, axis=0), cdf))
```

[...]

PR #1427, opened 2026-08-15
Recondition CDFs in surface_impedance to be strictly physical and make C array in coax media

ZiadHatab:
Summary is here: https://github.com/scikit-rf/scikit-rf/issues/1423#issuecomment-5303249398

This PR addresses the case when a user provides unphysical rms roughness that causes the CDFs of one boundary to advance beyond another, resulting in negative mixing weights. There is no "fix" for these inputs as they are inherently unphysical. My approach is to avoid catastrophic values by reconditioning the CDFs and provide the user with a warning to review their inputs. My recoditioning approach is based on [isotonic fit](https://en.wikipedia.org/wiki/Isotonic_regression), that always default to correct answer when user provide physical inputs. I'm sure there are other ways to do this, but in all cases, you need to edit the CDFs for these "bad" inputs.

The second commit makes C in coax media an array matching the size of the frequency array.

### DIFF
--- a/skrf/media/coaxial.py
+++ b/skrf/media/coaxial.py
@@ -520,7 +520,7 @@ class Coaxial(DistributedCircuit, Media):
             distributed capacitance, in F/m
 
         """
-        return 2.*np.pi*self.epsilon_prime/np.log(self.b/self.a)
+        return 2.*np.pi*self.epsilon_prime/np.log(self.b/self.a)*np.ones_like(self.frequency.f)
 
     @property
     def G(self) -> NumberLike:
--- a/skrf/tlineFunctions.py
+++ b/skrf/tlineFunctions.py
@@ -279,6 +279,11 @@ def surface_impedance(f: NumberLike, material_properties: list[dict] | dict,
     Richardson extrapolation, until ``Zs`` changes by less than ``rtol``; a
     ``RuntimeWarning`` is issued if the refinement does not converge.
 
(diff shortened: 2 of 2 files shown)


skrf/media/coaxial.py
```diff
@@ -520,7 +520,7 @@ def C(self) -> NumberLike:
             distributed capacitance, in F/m
 
         """
-        return 2.*np.pi*self.epsilon_prime/np.log(self.b/self.a)
+        return 2.*np.pi*self.epsilon_prime/np.log(self.b/self.a)*np.ones_like(self.frequency.f)
```
jhillairet, line 523: which implementation do you prefer vs #1431 ?

jhillairet (approved):
LGTM. To be discussed the broadcasting implementation (vs the one proposed in #1431 )

skrf/media/coaxial.py
ZiadHatab: I don't think either option would make a significant difference here. Multiplying by `np.ones_like(self.frequency.f)` has negligible overhead compared with the other computations in the coax class. Therefore, I would keep the implementation simple and multiply by `np.ones_like(self.frequency.f)`.

The other RLG parameters are also computed on a per-frequency-point basis, so I don't see a reason to change the logic for C. I don't see an advantage to doing so; it would only make the code harder to understand because you would need to look up `np.broadcast_to()` to understand what's going on.

PR #1435, opened 2026-09-04
support passivity assessment for one-port networks

### LINKED ISSUE #1432: Passivity assessment of 1-port networks
In #1426 the ai made this comment:
> Network.is_passive() returns False for every one-port, whatever |S11| is.
A matched load, an ideal short and an ideal open all come back non-passive, and
so does an active one-port, so the predicate carries no information there. The
path goes through passivity() raising ValueError('Doesn't exist for one ports'), caught at network.py:2169-2170 and turned into False. Those two
lines are also uncovered.

We should consider fixing it.

niketh-keshavan:
fixes https://github.com/scikit-rf/scikit-rf/issues/1432

currently, `Network.is_passive()` always returns `False` for any one-port network because `passivity()` raises `ValueError` for one-port networks which `is_passive()` then catches and converts to `False`.  we can address this by updating `passivity()` to support one-port networks (passivity metric $S^\mathrm{H}S$ reduces to $\lvert S_{11} \rvert$ and simplifying `Network.is_passive()` to evaluate one and multi-port networks through the same test.

changes:

`skrf/network.py`
- update `passivity()` to accept one-port S-parameter arrays which yields metric `(nfreqs, 1, 1)` with entry $\lvert S_{11} \rvert$
- remove `try` / `except` `ValueError` fallback from `Network.is_passive()` so one-ports go through same path as multi-port
- update docstrings accordingly

`skrf/tests/test_network.py`:
- add `test_passivity()` to verify that `Network.passivity` and `rf.network.passivity` return $\lvert S_{11} \rvert$ with shape `(nfreqs, 1, 1)` and that multi-port behavior is unchanged
- extend `test_is_passive` with one-port test cases

### DIFF
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -2170,10 +2170,7 @@ class Network:
         bool : boolean
 
         """
-        try:
-            M = np.square(self.passivity)
-        except ValueError:
-            return False
+        M = np.square(self.passivity)
 
         I = np.identity(M.shape[-1])
         for f_idx in range(len(M)):
--- a/skrf/network.py
+++ b/skrf/network.py
@@ -1814,16 +1814,22 @@ class Network:
     @property
     def passivity(self) -> ndarray:
         r"""
-        Passivity metric for a multi-port network.
+        Passivity metric for a one-port or multi-port network.
 
-        This returns a matrix who's diagonals are equal to the total
+        This returns a matrix whose diagonals are equal to the total
         power received at all ports, normalized to the power at a single
         excitement port.
 
         Mathematically, this is a test for unitary-ness of the
         s-parameter matrix [#]_.
 
-        For two port this is
+        For a one-port network this is
+
+        .. math::
+
+                |S_{11}|
+
+        For a two-port network this is
 
         .. math::
 
@@ -1833,7 +1839,7 @@ class Network:
 
         .. math::
 
-                S^H \cdot S
+                \sqrt( S^H \cdot S)
 
         where :math:`H` is conjugate transpose of S, and :math:`\cdot`
         is dot product.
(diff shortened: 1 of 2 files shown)


skrf/network.py
```diff
@@ -1814,16 +1814,22 @@ def port_tuples(self) -> list[tuple[int, int]]:
     @property
     def passivity(self) -> ndarray:
         r"""
-        Passivity metric for a multi-port network.
+        Passivity metric for a one-port or multi-port network.
```
Vinc0110: Why not just "Passivity metric for a network"?

skrf/network.py
```diff
@@ -8454,9 +8467,6 @@ def passivity(s: np.ndarray) -> np.ndarray:
     ------------
     .. [#] http://en.wikipedia.org/wiki/Scattering_parameters#Lossless_networks
     """
-    if s.shape[-1] == 1:
```
Vinc0110, line 8457: I assume there must have been a reason to exclude this one-port case. Can we really just remove the test and apply the same equation for any network?

skrf/network.py
```diff
@@ -8411,18 +8414,24 @@ def s2g(s: np.ndarray, z0: NumberLike = 50) -> np.ndarray:
 ## these methods are used in the secondary properties
 def passivity(s: np.ndarray) -> np.ndarray:
     r"""
-    Passivity metric for a multi-port network.
+    Passivity metric for a one-port or multi-port network.
```
Vinc0110: Why not just "Passivity metric for a network"?

Vinc0110 (commented):
Thank you for the PR! It looks good already and I like the extra tests.
Please note my comments and questions above (open for discussion).

skrf/network.py
niketh-keshavan: minimal diff =D but I'll change this and below, definitely sounds better

skrf/network.py
niketh-keshavan: pretty sure the math works out the same way
given passivity means $I - S^H S \ge 0$, for 1-port case just reduce to $1 - |S_{11}|^2 \ge 0$, basically $|S_{11}| \le 1$
looking at git history the check was probably just copied from `reciprocity()` where the 1-port case is just 
($S - S^T = 0$), for passivity the 1-port behavior is meaningful.
numPy handles `(1,1)` matrices perfectly so removing the check should just let 1-ports run through the same dissipation test without any special cases

Vinc0110: That makes sense. Thanks for having another look.

Vinc0110 (approved):
I think this PR is ready now. Let's keep it open for a while in case someone else wants to have a look.

jhillairet (approved):

PR #1437, opened 2026-09-07
Allow configuring the default frequency unit

### LINKED ISSUE #1428: default frequency unit  should be configurable
we could have  this set in skrf.io.constants. or could make a config file. 

also i think defaulting to 'hz' is less useful for most users than 'ghz' .

September11111111111:
Closes #1428.

This PR adds `skrf.constants.FREQ_UNIT_DEFAULT` to configure the unit used when `unit` is omitted from `Frequency` or `Frequency.from_f`.

The built-in default remains Hz. Explicit units take precedence, and frequencies are still stored internally in Hz.

### Usage

```python
import skrf as rf

rf.constants.FREQ_UNIT_DEFAULT = "GHz"
freq = rf.Frequency.from_f([1, 2])

# freq.f: [1e9, 2e9] Hz
# freq.f_scaled: [1, 2]
# freq.unit: "GHz"
```

The setting applies to newly created frequency objects in the current Python process. Existing objects are unchanged; no configuration file is required.

### Changes

- Add the configurable default and update API documentation.
- Remove the omitted-unit deprecation warning.
- Explicitly use Hz in NanoVNA calls and an HP8510C example that already supply Hz values.
- Add tests for unit selection, scaling, configuration changes, Touchstone compatibility, and NanoVNA communication.

### Validation

- Focused tests: 47 passed.
- Full Python suite on Windows / Python 3.12.3: 1870 passed, 44 skipped, 4 xfailed, 1 failed.
- The MDIF failure also occurs before this change due to temporary-file handling on Windows. The same test passes under WSL2 Ubuntu 24.04.
- Ruff and `git diff --check` passed.

No physical instrument tests or full notebook execution were performed.


skrf/frequency.py
```diff
@@ -147,11 +149,7 @@ def __init__(self, start: float = 0, stop: float = 0, npoints: int = 0,
 
         """
         if unit is None:
-            warnings.warn('''
```
arsenovic, line 150: i would remove this warning

skrf/frequency.py
arsenovic: i would rename these to start_hz , stop_hz to be explicity.

skrf/vi/vna/nanovna/tests/test_nanovna.py
arsenovic, line 1: this should be removed from this PR

arsenovic (commented):
thanks for the adds

skrf/vi/vna/nanovna/tests/test_nanovna.py
September11111111111: Thanks for the review! I've addressed the comments in d6b39ec6.

skrf/vi/vna/hp/hp8510c.py
```diff
@@ -88,7 +88,7 @@ class HP8510C(VNA):
         freq_block_1 = np.linspace(1e9,2e9,801)
         freq_block_2 = [10e9,11e9,12e9]
         freqs = np.concatenate((freq_block_1, freq_block_2))
-        vna.frequency = skrf.Frequency.from_f(freqs)
+        vna.frequency = skrf.Frequency.from_f(freqs, unit="Hz")
```
jhillairet: why is it needed to inforce the unit to Hz if the default is Hz? Same question for nanovna.py

September11111111111:
Hi @arsenovic, could you please rerun the two timed-out Code testing jobs (Python 3.10 and 3.11) when you have a chance?

Both jobs reached the six-hour limit after the "Running tests" message, without a pytest session header or a reported test failure. I tried reproducing this under WSL Ubuntu 24.04 with Python 3.10.21 and 3.11.16 and the dependency versions recorded in the CI logs. Both full Python test runs passed: 1857 passed, 44 skipped, and 4 xfailed. The local Python builds and runner environment differ from GitHub Actions, so this does not rule out a problem in the PR; I haven't identified the cause yet.

A rerun would help us see whether the timeout recurs. I've also prepared a temporary CI diagnostic patch locally. If it happens again, I can push it for review and approval to run; it captures the live pytest/worker stacks and process information before a bounded timeout and uploads the diagnostic logs. Thanks for your help!

September11111111111:
Hi @arsenovic, I've pushed temporary CI diagnostics in 7500c720 to help investigate the Python 3.10/3.11 timeouts. Could you please review the workflow change and approve the Actions run if approval is required?

I also tested the CI merge commit (`1dbd1e0`) under WSL using fresh environments with the Ubuntu 24.04 Python 3.10.21/3.11.16 builds published by `actions/python-versions`, matching the dependency versions recorded in the failed jobs. Both full Python test runs passed: 1857 passed, 44 skipped, and 4 xfailed (119s and 105s). The underlying WSL environment still differs from the hosted runner, so the CI timeout remains unexplained.

The diagnostic commit preserves the pytest arguments and `-n auto`. For 3.10/3.11 it captures parent/worker stacks and process information roughly every two minutes, stops the test process group after a ten-minute diagnostic limit, and uploads the logs. Only after a diagnostic timeout does it run bounded import/collection probes. I verified normal completion, failure status preservation, and stack capture/worker cleanup on a simulated hang locally.

One local observation: during collection, `skrf/tests/test_vectorfitting.py` runs a unittest suite at import time; multiple workers were simultaneously executing `test_dc_enforcement` inside NumPy QR decomposition. That behavior predates this PR, and I have not established that it caused the CI timeouts. The diagnostics should help distinguish this from other causes on the hosted runner. I plan to remove the temporary instrumentation once we have enough evidence. Thanks!

September11111111111:
Hi @jhillairet, all Python 3.10?3.14 jobs in the main test workflow now pass with Haswell selected.

The remaining failure is in `test-notebooks (3.12)`: the ADRF5720 example raises `BadZipFile` when opening a downloaded archive. The notebook and URL are unchanged from the previous passing run. My local requests currently receive a 404 HTML page, although the official product page still links to the same URL.

Could you please rerun the failed notebook job? If it persists, we can investigate the external download separately. Thank you!

arsenovic:
hey sorry about the delay on this. 

let me clarify the original issue. there are two use-cases with frequency units. 
 
1)  you need to specify the units of the frequencies you are using to construct a Frequency object 
2)  you want to have a default unit that all Frequency objects use regardless of what unit was specified to construct it. 

the 2) is really the ability of skrf to  coerce  frequency units by default,  so when you get  a s2p file with stupid units, you dont have to convert them yourself. typically a user wants to work in a specfied unit, regardless of what a file says. i think my original issue should have clarified this. sorry about that.  

in my mind , this is the expected behavior 
```
Frequency(1,2,unit='ghz')  #  should be 1-9ghz, units of 'ghz'
ntwk = rf.Network('myTouchstone.s2p')  # freq units in this file could be anything
ntwk.frequency.unit # should be set to  rf.constants.DEFFAULT_FREQ_UNIT 
```

maybe  `rf.constants.DEFFAULT_FREQ_UNIT=None` could provide a 'passive' unti, which allows the `rf.Network`  to have frequency untis of  whatever it is init'ed with.

September11111111111:
Hi @arsenovic, thanks again for clarifying the intended behavior. I've updated the implementation in `5595da8e`:

- Explicit input units determine the actual frequencies, while the configured global unit controls the resulting representation, including networks read from Touchstone files.
- `FREQ_UNIT_DEFAULT = None` is the passive default: explicit/file units are preserved, and omitted Frequency input units use Hz.
- Documentation and regression tests cover these cases.

All CI checks for this commit passed, including Python 3.10-3.14, notebooks, linting, and the documentation build.

Could you take another look when you have time and let me know whether this matches your expectations? I'm happy to make further adjustments. Thank you!

